PR #1, opened 2025-12-02
feat(YouTube - LithoFilterPatch): Add support for accessibility filtering

inotia00:
Adds support to filter litho components from their accessibility strings.

Custom filter can now filter accessibility strings with the optional `#` declaration. `#` Is declared after the path, but before the optional buffer declaration of `$`.

&nbsp;

### Custom filter examples
`video_action_bar`
Hides the entire player action bar.

`video_action_bar#id.video.share.button`
Hides the player action bar share button.

`video_action_bar#share.button`
Functionally identical to example above, since partial pattern matches are allowed.

`video_action_bar#id.video.share.button$Mr. Beast`
Hides the player action bar share button, but only if the buffer (which contains the videos title/descriptions/channel) contains "Mr. Beast"

`video_action_bar$Mr. Beast`
Hides the entire player action bar, but only if the buffer (which contains the video title/description/channel) contains "Mr. Beast"

### DIFF
--- a/patches/src/main/kotlin/app/morphe/util/FreeRegisterProvider.kt
+++ b/patches/src/main/kotlin/app/morphe/util/FreeRegisterProvider.kt
@@ -6,6 +6,7 @@ import app.morphe.util.FreeRegisterProvider.Companion.returnOpcodes
 import app.morphe.util.FreeRegisterProvider.Companion.writeOpcodes
 import app.revanced.patcher.extensions.InstructionExtensions.getInstruction
 import app.revanced.patcher.extensions.InstructionExtensions.instructions
+import com.android.tools.smali.dexlib2.Format
 import com.android.tools.smali.dexlib2.Opcode
 import com.android.tools.smali.dexlib2.Opcode.*
 import com.android.tools.smali.dexlib2.iface.Method
@@ -220,34 +221,31 @@ private fun Method.findFreeRegisters(
 
 private fun Method.buildInstructionOffsetArray(): IntArray {
     val instructionCount = instructions.count()
-    val offsetArray = IntArray(instructionCount) { -1 } // Use -1 to indicate invalid/not mapped
+    val offsetArray = IntArray(instructionCount) { -1 }
     var currentOffset = 0
 
     for (i in 0 until instructionCount) {
         val instruction = getInstruction(i)
         val format = instruction.opcode.format
 
-        // Only map non-payload instructions.
-        // Payloads are special and aren't part of normal code flow.
         if (!format.isPayloadFormat) {
             offsetArray[i] = currentOffset
 
+            // Get size in bytes from format.
             val sizeInBytes = format.size
-            currentOffset += if (sizeInBytes > 0) {
-                sizeInBytes / 2  // Convert bytes to code units.
-            } else {
-                // Shouldn't happen for non-payload, but handle it.
-                1
+            val sizeInCodeUnits = when {
+                sizeInBytes > 0 -> sizeInBytes / 2  // Normal instruction
+                format == Format.UnresolvedOdexInstruction -> 1  // Default size
+                else -> 1  // Fallback for any other edge case
             }
+
+            currentOffset += sizeInCodeUnits
         }
-        // Skip payloads - they don't get entries in the array.
-        // Their offsets don't matter for branch calculations,
-        // and they remain -1 in the array.
+        // Skip payloads
     }
 
     return offsetArray
 }
-
 /**
  * Gets all branch target indices for a branch instruction.
  * Returns empty list if not a branch or targets cannot be determined.
@@ -414,18 +412,16 @@ private fun Method.findFreeRegistersInternal(
  *
  * @param targetOffset Target code offset in 16-bit units
  * @param offsetArray Array mapping instruction index to code offset (-1 for payloads)
- * @return Instruction index at or before the target offset, or null if not found
+ * @return Instruction index at the target offset, or null if not found
  */
 private fun Method.findInstructionIndexByOffset(
     targetOffset: Int,
     offsetArray: IntArray
 ): Int? {
-    // Simple linear search (could optimize with binary search if needed).
-    for (i in offsetArray.indices) {
-        val offset = offsetArray[i]
-        if (offset != -1 && offset == targetOffset) {
-            return i
-        }
+    // Simple linear search using indexOfFirst
+    val index = offsetArray.indexOfFirst { it == targetOffset }
+    if (index >= 0) {
+        return index
     }
 
     // Should never happen.
@@ -435,9 +431,9 @@ private fun Method.findInstructionIndexByOffset(
         "Could not find exact instruction offset for method: $this at offset: $targetOffset. " +
                 "Please file a bug report in the Morphe patches repo"
     )
-
     return null
 }
+
 /**
  * Starting from and including the instruction at index [startIndex],
  * finds the next register that is written to and not read from. If a return instruction


extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/LithoFilterPatch.java
```diff
@@ -42,6 +50,12 @@ public String toString() {
             builder.append(" Path: ");
             builder.append(path);
             if (Settings.DEBUG_PROTOBUFFER.get()) {
+                // If both accessibilityId and accessibilityLabel are empty, don't add them to the debug log.
+                if (accessibility.length() > 1) {
+                    // AccessibilityId and AccessibilityLabel are pieces of BufferStrings.
```
inotia00: Would it be better to integrate accessibility into the buffer?

patches/src/main/kotlin/app/morphe/patches/youtube/misc/litho/filter/LithoFilterPatch.kt
```diff
@@ -128,13 +139,44 @@ val lithoFilterPatch = bytecodePatch(
...
+        }
+
+        val accessibilityLabelMethod = with (accessibilityLabelFingerprint) {
+            val index = instructionMatches.first().index
+            method.getInstruction<ReferenceInstruction>(index).reference as MethodReference
+        }
+
+        // TODO: Completely remove compatibility with YouTube 19.16 to make code easier to maintain.
+        //       YouTube 19.16 is sufficiently out of date, and without the 'Disable update screen' patch,
+        //       the 'Update your app' dialog will be shown and users will not be able to use the app.
```
inotia00: The code is complex to make the patch work on YouTube 19.16.

patches/src/main/kotlin/app/morphe/patches/youtube/misc/litho/filter/LithoFilterPatch.kt
```diff
@@ -143,9 +185,55 @@ val lithoFilterPatch = bytecodePatch(
...
+                    indexOfFirstInstructionReversedOrThrow(buttonViewModelIndex, Opcode.IF_EQZ)
+                val nullCheckRegister = getInstruction<OneRegisterInstruction>(nullCheckIndex).registerA
+
+                accessibilityIdRegister = findFreeRegister(nullCheckIndex, false,
+                    freeRegister, buttonViewModelRegister, nullCheckRegister)
+                accessibilityLabelRegister = findFreeRegister(nullCheckIndex, false,
+                        freeRegister, buttonViewModelRegister, accessibilityIdRegister, nullCheckRegister)
+
+                // TODO: Delete this debug line once you have checked the latest YouTube.
+                Logger.getLogger(this::class.java.name).info(
```
inotia00: Remove this debug line as soon as you finish testing on the latest YouTube.

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/LithoFilterPatch.java
LisoUseInAIKyrios: I think accessibility can be separate. The buffer is now very large because of the workaround required for 20.22+

Custom filter could add new syntax to check accessibility strings.

Maybe `#string` to match accessibility
`action_bar#Share` Action bar path with "Share" accessibility.
`action_bar#Share$bold` Action bar path with "Share" accessibility and 'bold' in the buffer

~~Can add support to custom filter in a separate PR.~~ edit: Changes added to this PR.

patches/src/main/kotlin/app/morphe/patches/youtube/misc/litho/filter/LithoFilterPatch.kt
LisoUseInAIKyrios: Can ignore 19.16, because YT dropped support earlier this year, and as you explained there isn't a patch to hide the update required screen.

Right now the only old version I've been checking that still works is 19.43, and we may drop that too if it becomes too much of a burden.

patches/src/main/kotlin/app/morphe/patches/youtube/misc/litho/filter/LithoFilterPatch.kt
```diff
@@ -143,9 +186,55 @@ val lithoFilterPatch = bytecodePatch(
...
+                    getInstruction<OneRegisterInstruction>(buttonViewModelIndex).registerA
+                accessibilityIdIndex = buttonViewModelIndex + 2
+
+                nullCheckIndex =
+                    indexOfFirstInstructionReversedOrThrow(buttonViewModelIndex, Opcode.IF_EQZ)
+                val nullCheckRegister = getInstruction<OneRegisterInstruction>(nullCheckIndex).registerA
+
+                accessibilityIdRegister = findFreeRegister(nullCheckIndex, false,
+                    freeRegister, buttonViewModelRegister, nullCheckRegister)
+                accessibilityLabelRegister = findFreeRegister(nullCheckIndex, false,
```
LisoUseInAIKyrios: `findFreeRegister()` could be refactored into an object, where you create it the same as the first free register:
`val freeRegisters = FindFreeRegisters(insertIndex, ignoredRegisters)`

and it searches and finds all free registers available and stores them internally.

Then free registers are popped off from the object
```
val nullCheckRegister = freeRegisters.allocateFreeRegister()
val accessibilityIdRegister = freeRegisters.allocateFreeRegister()
val accessibilityLabelRegister = freeRegisters.allocateFreeRegister()
```

I considered doing this when I first wrote the function, but at the time we only needed 1 or 2 free registers so it wasn't that useful. But now we're using a lot more free registers and this change would clean up the current requirement of passing all previously allocated free registers.

Additionally, internally this change is needed to correctly search branches.

patches/src/main/kotlin/app/morphe/util/BytecodeUtils.kt
```diff
@@ -49,12 +52,17 @@ import java.util.EnumSet
  * swap register contents if a 4-bit register is required.
  *
  * @param startIndex Inclusive starting index.
+ * @param checkBranch Whether to check branch opcodes. This can only be used on verified indices.
```
LisoUseInAIKyrios: This works surprisingly well a lot of the time, because the same free write register is regularly used for both code paths.

Eventually someone should change the function to search both branches for all free registers then inner union the free registers found in each path. I think searching just 1 level deep would be good enough (search both code paths but don't search branches in either of those paths), and that would simply the implementation because it wouldn't need to keep track of the instruction indices it has visited (it only needs a boolean parameter of "traverseBranches").

inotia00: It looks like the `findFreeRegister` function needs some refactoring, but I haven't looked into it right now because it seems like it would take too much time.

LisoUseInAIKyrios: Yes I agree. It's always been easier to work around the branching issue other ways than to properly search branches. I originally tried adding branch searching and it seemed to work, but I wasn't sure because of the weirdness of JVM branch offsets that work on raw byte offsets.

https://github.com/MorpheApp/morphe-patches/pull/1#discussion_r2579637634

LisoUseInAIKyrios:
Some player action buttons don't have accessibility names, which is strange.

Shorts action buttons cannot be hidden using accessibility, but maybe another way can be figured out.

If YT eventually rolls out the Shorts 'nextgen rendering' feature, which does not appear to use Litho and goes back to custom Android views, then it should be possible to hide the buttons without too much fuss.

LisoUseInAIKyrios:
I'm not seeing any accessibility text on anything in the player. Everything is an empty string or null. Is this expected? I tried both 20.14 and 20.47.

```
LithoFilterPatch: accessibilityText: 
LithoFilterPatch: accessibilityText: 
LithoFilterPatch: accessibilityText: 
LithoFilterPatch: accessibilityText: null
LithoFilterPatch: accessibilityText: 
LithoFilterPatch: accessibilityText: null
LithoFilterPatch: accessibilityText: 
LithoFilterPatch: accessibilityText: null
LithoFilterPatch: accessibilityText: null
LithoFilterPatch: accessibilityText: 
LithoFilterPatch: accessibilityText: 
LithoFilterPatch: accessibilityText: 
LithoFilterPatch: accessibilityText: 
LithoFilterPatch: accessibilityText: null
LithoFilterPatch: accessibilityText: null
LithoFilterPatch: accessibilityText: null
LithoFilterPatch: accessibilityText: 
LithoFilterPatch: accessibilityText: null
LithoFilterPatch: accessibilityText: 
LithoFilterPatch: accessibilityText: null
LithoFilterPatch: accessibilityText: 
LithoFilterPatch: accessibilityText: null
LithoFilterPatch: accessibilityText: null
LithoFilterPatch: accessibilityText: null
LithoFilterPatch: accessibilityText: 
LithoFilterPatch: accessibilityText: null
LithoFilterPatch: accessibilityText: null
LithoFilterPatch: accessibilityText: 
LithoFilterPatch: accessibilityText: null
LithoFilterPatch: accessibilityText: null
LithoFilterPatch: accessibilityText: 
```

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/LithoFilterPatch.java
LisoUseInAIKyrios: I added custom filter support for this PR.

patches/src/main/kotlin/app/morphe/patches/youtube/misc/litho/filter/LithoFilterPatch.kt
LisoUseInAIKyrios: Added this today.

If only 1 path depth is followed and switch statements are handled as a dead end, then the code is simple.

inotia00:
Yes, that's normal. It seems that not all buttons or elements have accessibility ids or accessibility text.

The Shorts action button also has accessibility text, but because the YouTube API response has a slightly unusual structure, it can't be accessed through this method.

inotia00:
This PR also solves this issue:

channel_page

inotia00:
> Some player action buttons don't have accessibility names, which is strange.
> 
> Shorts action buttons cannot be hidden using accessibility, but maybe another way can be figured out.
> 
> If YT eventually rolls out the Shorts 'nextgen rendering' feature, which does not appear to use Litho and goes back to custom Android views, then it should be possible to hide the buttons without too much fuss.

We can hide the Shorts action button using the index: https://github.com/MorpheApp/morphe-patches/pull/4

LisoUseInAIKyrios:
This seems ready to merge?

inotia00:
Yeah, it seems to work fine.

patches/src/main/kotlin/app/morphe/util/FreeRegisterProvider.kt
```diff
@@ -0,0 +1,515 @@
...
+ *                                  encountered before any free registers are found.
+ */
+fun Method.getFreeRegisterProvider(startIndex: Int, vararg registersToExclude: Int) =
+    FreeRegisterProvider(this, startIndex, *registersToExclude)
+
+/**
+ * Simple wrapper around [findFreeRegister] that allows finding then allocating multiple registers.
+ * If you only need a one free register, instead use [findFreeRegister].
+ */
+class FreeRegisterProvider internal constructor(
```
LisoUseInAIKyrios, line 56: For some code such as this, we could selectively release under GPLv3 but with a different GPL 7b requirement than all other code in this repo.

To use any code in this repo requires showing a user facing attribution message (including if any code added to patched app).

But code like this specific class, we could selectively change the attribution requirement to be source code only. Meaning someone else is free to use this exact class in their project (as-is, or modified), as long as the an attribution comment block is preserved in the source code file that states where the code came from. All other code would still require showing a user facing message in the app and patched app.

But, we can do that later (if ever).

(and just so it's clear to anyone reading this, to use _any_ code in this repo requires showing a user facing message that code from Morphe is being used. And that includes this specific source file). See the project license file for exact details.


**update**: FindFreeRegister and a few other classes were changed to a more permissive section 7. See file comment blocks for details.

PR #2, opened 2025-12-02
fix(YouTube - Return YouTube Dislike): Sometimes incorrect dislike counts shown in Shorts

inotia00:
The cause of the issue is that the buffer of the dislike button was changed on the server side.

As a result, 'ReturnYouTubeDislikeFilter' has not updated the videoId since a few months ago.

Tested on YouTube 20.05.46, not sure if it works with the new type of action button (A/B testing hasn't arrived on my account yet).

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/ReturnYouTubeDislikePatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/ReturnYouTubeDislikePatch.java
@@ -100,6 +100,19 @@ public class ReturnYouTubeDislikePatch {
     /**
      * Injection point.
      *
+     * Logs if new litho text layout is used.
+     */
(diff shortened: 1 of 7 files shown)


LisoUseInAIKyrios:
This is working with 20.47 and the new bold icons.

With the bold icons that changed everywhere, it seems only the icon resource names changed and everything else stayed the same.

The new non litho Shorts player hasn't been figured out and is forced off with a patch. Forcing those two flags on doesn't give the experimental Shorts player, possibly because it also requires new server side data, or maybe there's additional other unidentified flags that must also be forced on. That experiment is almost a year old and still not being rolled out to everyone. Maybe YT is taking a while to make changes, or maybe they completely abandoned it.

https://github.com/MorpheApp/morphe-patches/blob/748d9c4c0bd1aef3a55e92048d96fc4f87b1eb2e/patches/src/main/kotlin/app/morphe/patches/youtube/layout/hide/shorts/HideShortsComponentsPatch.kt#L272-L285

PR #7, opened 2025-12-10
ci: Add bundles generator

inotia00:
When a patch is released, [bundle.json](https://github.com/Jman-Github/ReVanced-Patch-Bundles/blob/bundles/patch-bundles/inotia00-patch-bundles/inotia00-latest-patches-bundle.json) and [list.json](https://github.com/Jman-Github/ReVanced-Patch-Bundles/blob/bundles/patch-bundles/inotia00-patch-bundles/inotia00-latest-patches-list.json) are automatically generated.

To test this, a new release must be published via GitHub CI (semantic-release).

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -28,8 +28,8 @@
       {
         "assets": [
           "CHANGELOG.md",
-          "gradle.properties",
           "bundle.json",
+          "gradle.properties",
           "list.json"
         ],
         "message": "chore: Release v${nextRelease.version} [skip ci]\n\n${nextRelease.notes}"


inotia00:
This was tested on a patch repository based on Patcher 19, which was forked a year ago, but I'm not sure if it will work here as well.

(Testing requires the `Release` action)

inotia00:
Yeah it works https://github.com/MorpheApp/morphe-patches/commit/532a0a84cd870e44d5c4c425b96894da9c7a4dcd

PR #8, opened 2025-12-11
fix: `backtick` is replaced with an empty string in bash

inotia00:
Strings surrounded by backticks (`) are [converted to empty strings in bash](https://github.com/MorpheApp/morphe-patches/commit/c21246e19304cbb87ea7936f617a4d0cfbb62a06).

#### CHANGELOG.md

changelog

#### patches-bundle.json

bundle

This PR handles exceptions for backticks.

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -20,7 +20,7 @@
     [
       "@semantic-release/exec",
       {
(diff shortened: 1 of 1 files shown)


inotia00:
This PR still doesn't fix the backtick issue.

Since the error occurs when calling JavaScript, we need to fix the issue in a different way.

I'll open a new PR.

PR #10, opened 2025-12-16
fix(YouTube): Changes the default values for some settings

inotia00:
For the `Hide video action buttons` patch, [the `Hide Clip` setting is not added in YouTube 20.22+](https://github.com/MorpheApp/morphe-patches/blob/2646531c39d9d4fdcff3c6e4d507d050b1495c9d/patches/src/main/kotlin/app/morphe/patches/youtube/layout/buttons/action/HideVideoActionButtonsPatch.kt#L67).

For some users, [the Clip button is still hidden using the old method (hiding via a litho path) in YouTube 20.22+](https://github.com/MorpheApp/morphe-patches/blob/2646531c39d9d4fdcff3c6e4d507d050b1495c9d/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/VideoActionButtonsFilter.java#L56).

The default value for the `Hide Clip` setting is `TRUE`, which hides the button.

There's no way to disable this, as it's not added to the settings.

For handling missing settings in some YouTube versions, changing the default values of those settings to `FALSE` is the easiest workaround.


LisoUseInAIKyrios:
The clip button is now showing for me with 20.22+ and it has a unique path (not a buffer item), so it could be filtered.

Before buffer filtering existed, hide clip button was an option but it had a dialog confirmation that hiding is not always possible for all users. I think that same functionality could return.

PR #12, opened 2025-12-16
fix(AddResources): Change resource system to per app, remove per patch resource system that is problematic with Crowdin

LisoUseInAIKyrios:
Removes the concept of adding patch strings per patch, and instead adds all strings for a given app.

Strings are tiny in size, and it's unnecessary to exclude any strings if a patch is excluded.

This standardizes the string files to work with Crowdin without any pre-processing or post-processing that was previously required.

Also fixes a CLI patcher warning if a default (English) string is deleted before the delete can propagate to/from Crowdin.


LisoUseInAIKyrios:
After spending too much time ripping out old overly complex code, this is ready.

PR #14, opened 2025-12-16
fix(YouTube - Loop video): `Enable loop video` setting not working in playlist

inotia00:
Until now, the `Enable loop video` setting didn't work when playing videos in a playlist.

Instead of calling the loop method, loop video functionality is implemented by simply seeking to 00:00.

### DIFF
--- a/patches/src/main/kotlin/app/morphe/patches/youtube/layout/player/fullscreen/ExitFullscreenPatch.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/youtube/layout/player/fullscreen/ExitFullscreenPatch.kt
@@ -12,6 +12,8 @@ import app.morphe.patches.youtube.misc.settings.PreferenceScreen
 import app.morphe.patches.youtube.misc.settings.settingsPatch
 import app.morphe.patches.youtube.video.information.playerStatusMethod
 import app.morphe.patches.youtube.video.information.videoInformationPatch
+import app.morphe.util.indexOfFirstInstructionOrThrow
+import com.android.tools.smali.dexlib2.Opcode
 
 @Suppress("unused")
 internal val exitFullscreenPatch = bytecodePatch(
@@ -51,9 +53,14 @@ internal val exitFullscreenPatch = bytecodePatch(
             ListPreference("morphe_exit_fullscreen")
         )
 
-        playerStatusMethod.addInstruction(
-            0,
-            "invoke-static { p1 }, $EXTENSION_CLASS_DESCRIPTOR->endOfVideoReached(Ljava/lang/Enum;)V"
-        )
+        playerStatusMethod.apply {
+            val insertIndex =
+                indexOfFirstInstructionOrThrow(Opcode.SGET_OBJECT) + 1
+
+            addInstruction(
+                insertIndex,
+                "invoke-static { p1 }, $EXTENSION_CLASS_DESCRIPTOR->endOfVideoReached(Ljava/lang/Enum;)V",
+            )
+        }
     }
 }
--- a/patches/src/main/kotlin/app/morphe/patches/youtube/video/information/Fingerprints.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/youtube/video/information/Fingerprints.kt
@@ -45,18 +45,17 @@ internal object PlayerInitFingerprint : Fingerprint(
     )
 )
 
-/**
- * Matches using class found in [PlayerInitFingerprint].
- */
-internal object PlayerStatusFingerprint : Fingerprint(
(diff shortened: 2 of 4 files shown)


patches/src/main/kotlin/app/morphe/patches/youtube/layout/player/fullscreen/ExitFullscreenPatch.kt
```diff
@@ -53,13 +51,9 @@ internal val exitFullscreenPatch = bytecodePatch(
...
-            val insertIndex = indexOfFirstInstructionReversedOrThrow(Opcode.RETURN_VOID)
-
-            addInstructionsAtControlFlowLabel(
-                insertIndex,
-                "invoke-static {}, $EXTENSION_CLASS_DESCRIPTOR->endOfVideoReached()V",
-            )
-        }
+        playerStatusMethod.addInstruction(
+            0,
+            "invoke-static { p1 }, $EXTENSION_CLASS_DESCRIPTOR->endOfVideoReached(Ljava/lang/Enum;)V"
```
inotia00: Since patches are processed in lexicographic order, the `Exit fullscreen mode` patch is processed before the `Loop video` patch.

Both patches insert code at index 0, so the patched app has the following smali instructions:

patched

The loop video status is checked first.

If loop video is turned off, the `endOfVideoReached` method is called.

patches/src/main/kotlin/app/morphe/patches/youtube/video/information/Fingerprints.kt
```diff
@@ -45,6 +45,21 @@ internal object PlayerInitFingerprint : Fingerprint(
     )
 )
 
+/**
+ * Matches using class found in [PlayerInitFingerprint].
+ */
+internal object PlayerStatusFingerprint : Fingerprint(
+    accessFlags = listOf(AccessFlags.PUBLIC, AccessFlags.FINAL),
+    returnType = "V",
+    parameters = listOf("L"),
```
inotia00: The first parameter type of this method is Enum:

```
    enum PlayerStatus {
        NEW,
        PLAYBACK_PENDING,
        PLAYBACK_LOADED,
        PLAYBACK_INTERRUPTED,
        INTERSTITIAL_REQUESTED,
        INTERSTITIAL_PLAYING,
        READY,
        VIDEO_REQUESTED,
        VIDEO_PLAYING,
        ENDED
    }
```

patches/src/main/kotlin/app/morphe/patches/youtube/layout/player/fullscreen/ExitFullscreenPatch.kt
LisoUseInAIKyrios: This has been a thing in the past, where different patches are changing the exact same lines of code and by coincidence it works because of the patch name sorting.

Can add a comment in both patches mentioning the other patch is also touching the same lines of code, and that Loop video changes must be first instructions in the final patched app.

patches/src/main/kotlin/app/morphe/patches/youtube/video/information/Fingerprints.kt
LisoUseInAIKyrios: Is this enum relevant for any other patches or possible future patches?

If the PlayerStatusEnum is fingerprinted or hooked, then this fingerprint (`PlayerStatusFingerprint`) could be declared in the patch execute:

```
execute {
   ...
    val playerStatusFingerprint = Fingerprint(
        accessFlags = listOf(AccessFlags.PUBLIC, AccessFlags.FINAL),
        returnType = "V",
        parameters = listOf(PlayerStatusEnumFingerprint.originalClassDef.type),
        filters = listOf(
            methodCall(
                definingClass = "Lj${'$'}/time/Instant;",
                name = "plus"
            )
        )
    )
   ...
}
```

inotia00: Yes, it's related.

Both `endOfVideoReached(Enum)` and `shouldLoopVideo(Enum)` check whether the name of this enum (status.name()) is `ENDED`.

For accuracy, it is better to check Enum class name.

patches/src/main/kotlin/app/morphe/patches/youtube/layout/player/fullscreen/ExitFullscreenPatch.kt
inotia00: Instead of always hooking at index 0, the patch now finds the index of the first opcode (Opcode.SGET_OBJECT).

Even if the patch name changes, it will still be patched to the correct location:

patched_2

LisoUseInAIKyrios:
Can merge all of these whenever you think they're ready

PR #17, opened 2025-12-23
fix(YouTube - Hide ads): Support `Hide fullscreen ads` on Android 13+ devices

inotia00:
In Android 13+, key event injection via Instrumentation is no longer available without additional permissions.

It appears to be a stealth patch not mentioned in the official Android documentation, but you can check it out on the [Google Issue Tracker](https://issuetracker.google.com/issues/244970399).

Instead of implementing complex key event injection or accessibility permissions, a simple patch has been implemented to close fullscreen ads (custom dialogs).


inotia00:
I integrated `hideGetPremiumPatch` into the `Hide ads` patch as I thought it would be better in terms of code maintainability.

inotia00:
Although the `statement_banner` filter was removed in commit https://github.com/MorpheApp/morphe-patches/commit/1ba623899ad6f89e4d4c44114dcea090118887d3, it is still a necessary:

https://github.com/inotia00/ReVanced_Extended/issues/3298#issuecomment-3633262097

`statement_banner` is closer to `YouTube Premium promotion` than `general ads`, so it has been integrated into the `Hide YouTube Premium promotion` setting.

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -357,11 +353,11 @@ This feature is only available for older devices"</string>
...
     <string name="morphe_hide_web_search_results_summary_on">Web search results are hidden</string>
     <string name="morphe_hide_web_search_results_summary_off">Web search results are shown</string>
-
-    <!-- ad.getpremium.hideGetPremiumPatch -->
     <string name="morphe_hide_get_premium_title">Hide YouTube Premium promotions</string>
-    <string name="morphe_hide_get_premium_summary_on">YouTube Premium promotions under video player are hidden</string>
-    <string name="morphe_hide_get_premium_summary_off">YouTube Premium promotions under video player are shown</string>
+    <string name="morphe_hide_get_premium_summary_on">"YouTube Premium promotions are hidden
+
+Limitations: YouTube Doodles page is not available"</string>
```
inotia00, line 361: Is this string appropriate? I may need to change the setting key.

LisoUseInAIKyrios: Yeah that looks ok. Setting keys can be changed if you think so.

patches/src/main/kotlin/app/morphe/patches/youtube/ad/general/Fingerprints.kt
```diff
@@ -23,3 +25,33 @@ internal fun indexOfAddListInstruction(method: Method) =
...
+    }
+)
+
+internal object LithoDialogBuilderFingerprint : Fingerprint(
+    accessFlags = listOf(AccessFlags.PUBLIC, AccessFlags.FINAL),
+    returnType = "V",
+    parameters = listOf("[B", "L"),
+    custom = { method, _ ->
+        method.containsLiteralInstruction(slidingDialogAnimation)
+                && indexOfShowDialogInstruction(method) >= 0
```
LisoUseInAIKyrios: I assume this is older code because this could be simplified/modernized using instruction filters that doesn't need custom block shenanigans.

But this is ok because it'll take too much time to update all the prior more complicated non instruction filter code and additional time to verify the refactoring didn't break anything.

patches/src/main/kotlin/app/morphe/patches/youtube/ad/general/Fingerprints.kt
```diff
@@ -23,3 +27,31 @@ internal fun indexOfAddListInstruction(method: Method) =
...
+internal object LithoDialogBuilderFingerprint : Fingerprint(
+    accessFlags = listOf(AccessFlags.PUBLIC, AccessFlags.FINAL),
+    returnType = "V",
+    parameters = listOf("[B", "L"),
+    filters = listOf(
+        methodCall(
+            opcode = Opcode.INVOKE_VIRTUAL,
+            name = "show"
+        ),
+        literal(slidingDialogAnimation),
```
LisoUseInAIKyrios: this can be simplified to:
```suggestion
        resourceLiteral(ResourceType.STYLE, "SlidingDialogAnimation")
```

And remove the shared field `slidingDialogAnimation = getResourceId(ResourceType.STYLE, "SlidingDialogAnimation")`

patches/src/main/kotlin/app/morphe/patches/youtube/ad/general/HideAdsPatch.kt
```diff
@@ -94,6 +104,60 @@ val hideAdsPatch = bytecodePatch(
             )
         }
 
+        // Hide fullscreen ad
+
+        LithoDialogBuilderFingerprint.let {
+            it.method.apply {
+                // Find the class name of the custom dialog
+                val dialogIndex = it.instructionMatches.first().index
+                val dialogClass = getInstruction(dialogIndex).getReference<MethodReference>()!!.definingClass
```
LisoUseInAIKyrios: If only the instruction is needed (and not the index too), then can use the instruction field on the match object:
```suggestion
                val dialogClass = it.instructionMatches.first().instruction.getReference<MethodReference>()!!.definingClass
```

PR #19, opened 2025-12-23
fix(YouTube - Remove viewer discretion dialog): Not working on YouTube 20.14.43+

inotia00:
This fixes a bug that doesn't work when A/B testing is applied.

To ensure correct behavior, the patch passes the `playabilityStatus.status` value to the extension in the `/player` endpoint response.


LisoUseInAIKyrios:
Merge these whenever you think they're ready 👍

PR #21, opened 2025-12-24
feat: Add new About dialog style

MarcaDian:
This adds a new style to the About dialog.

Day

Night

### DIFF
--- a/extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/MorpheAboutPreference.java
+++ b/extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/MorpheAboutPreference.java
@@ -34,7 +34,10 @@ import org.json.JSONObject;
 import java.net.HttpURLConnection;
 import java.net.SocketTimeoutException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
+import java.util.Map;
 
 import app.morphe.extension.shared.Logger;
 import app.morphe.extension.shared.Utils;
@@ -416,8 +419,20 @@ class WebViewDialog extends Dialog {
 }
 
 class WebLink {
+
+    /**
+     * Localized name replacements for links.
+     */
+    private static final Map<String, String> webLinkNameReplacements = new HashMap<>() {
+        {
+            put("website", "morphe_settings_about_links_website");
+            put("donate", "morphe_settings_about_links_donate");
+            put("translations", "morphe_settings_about_links_translations");
+        }
+    };
+
     final boolean preferred;
-    String name;
+    final String name;
     final String url;
 
     WebLink(JSONObject json) throws JSONException {
@@ -429,8 +444,9 @@ class WebLink {
 
     WebLink(boolean preferred, String name, String url) {
         this.preferred = preferred;
-        this.name = name;
         this.url = url;
+        String localizedNameKey = webLinkNameReplacements.get(name.toLowerCase(Locale.US));
+        this.name = (localizedNameKey != null) ? str(localizedNameKey) : name;
     }
 
     @NonNull
@@ -551,7 +567,6 @@ class AboutRoutes {
             for (int i = 0, length = donations.length(); i < length; i++) {
                 WebLink link = new WebLink(donations.getJSONObject(i));
                 if (link.preferred) {
-                    link.name = str("morphe_settings_about_links_donate");
                     links.add(link);
                 }
             }
@@ -559,9 +574,6 @@ class AboutRoutes {
             JSONArray socials = json.getJSONArray("socials");
             for (int i = 0, length = socials.length(); i < length; i++) {
                 WebLink link = new WebLink(socials.getJSONObject(i));
-                if (link.name.equalsIgnoreCase("Website")) {
-                    link.name = str("morphe_settings_about_links_website");
-                }
                 links.add(link);
             }
 
--- a/patches/src/main/resources/addresources/values/shared/strings.xml
+++ b/patches/src/main/resources/addresources/values/shared/strings.xml
@@ -68,8 +68,9 @@ To translate new languages or improve the existing translations, visit translate
     <string name="morphe_settings_about_links_dev_header">Note</string>
     <string name="morphe_settings_about_links_dev_body">This version is a pre-release and you may experience unexpected issues</string>
     <string name="morphe_settings_about_links_header">Official links</string>
-    <string name="morphe_settings_about_links_donate">Donate</string>
     <string name="morphe_settings_about_links_website">Website</string>
+    <string name="morphe_settings_about_links_donate">Donate</string>
+    <string name="morphe_settings_about_links_translations">Translations</string>
 
     <!-- misc.debugging.enableDebuggingPatch -->
     <string name="morphe_debug_screen_title">Debugging</string>


MarcaDian:
I couldn't check if the logo is displayed correctly, I tried replacing the link in the code but it doesn't load it, is the logo currently loaded from the API?

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/MorpheAboutPreference.java
```diff
@@ -68,55 +68,215 @@ private String createDialogHtml(WebLink[] aboutLinks) {
...
+        // Apply Morphe-style CSS.
+        builder.append("<style>");
+        builder.append("* { margin: 0; padding: 0; box-sizing: border-box; }");
+        builder.append("body { ");
+        builder.append("  background: ").append(backgroundColorHex).append(";");
+        builder.append("  color: ").append(foregroundColorHex).append(";");
+        builder.append("  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, sans-serif;");
+        builder.append("  padding: 24px;");
+        builder.append("  text-align: center;");
+        builder.append("}");
```
LisoUseInAIKyrios: Could use multi line syntax to clean up some of this. I think the string builder could be removed and simpler += is used (the immeasurable perf difference is less important than the readability and simplicity of the code).

```
// Apply Morphe-style CSS.
html += """
    .logo-container {
      margin: 0 auto 24px;
      width: 120px;
      height: 120px;
      border-radius: 28px;
      background: linear-gradient(135deg, """ + morpheBlue + """ 0%, """ + morpheTeal + """ 100%);
      padding: 3px;
      display: inline-block;
      box-shadow: 0 8px 24px rgba(30, 90, 168, 0.2);
    }
    """;
```

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -28,6 +27,7 @@ Second \"item\" text"</string>
     <string name="morphe_settings_screen_09_return_youtube_dislike_title" translatable="false">Return YouTube Dislike</string>
     <string name="morphe_settings_screen_11_misc_title">Miscellaneous</string>
     <string name="morphe_settings_screen_12_video_title">Video</string>
+    <string name="morphe_settings_screen_99_about_title">About Morphe</string>
```
LisoUseInAIKyrios: I think this could remain as just "About" because it already says "Morphe" on the title line above this.

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/MorpheAboutPreference.java
```diff
@@ -300,7 +471,7 @@ class AboutLinksRoutes {
     /**
      * Backup icon url if the API call fails.
      */
-    public static volatile String aboutLogoUrl = "https://morphe.software/favicon.ico";
+    public static volatile String aboutLogoUrl = "https://raw.githubusercontent.com/MorpheApp/morphe-branding/main/assets/morphe-logo/morphe_logo.svg";
```
LisoUseInAIKyrios: This can remain as the website icon, because this will only fallback if GitHub is down (so using GitHub again here will also fail).

LisoUseInAIKyrios:
> I couldn't check if the logo is displayed correctly, I tried replacing the link in the code but it doesn't load it, is the logo currently loaded from the API?

The repo is private so the images won't load unless you temporary change to a raw link from a logged in account (can test locally but don't check-in any auth tokens because that's private):

`https://raw.githubusercontent.com/MorpheApp/morphe-branding/refs/heads/main/assets/morphe-logo/morphe_logo.svg?token=privatetoken"`


But I noticed the existing json link I added isn't right and needs to be fixed (it's missing the refs path)

patches/src/main/resources/addresources/values/youtube/strings.xml
MarcaDian: The word "About" confuses translators and translations, as it cannot be translated unambiguously for some Slavic languages, and it has bothered me all these years. Now I will return it as it was.

LisoUseInAIKyrios: Add a comment to the strings.xml and explain that if it's difficult to translate as just "About" then translate from "About Morphe"

MarcaDian:
> The repo is private

I tried replacing it with a third-party file sharing service and that didn't work either. So I'm hoping that when the repo goes public it won't cause any problems.

LisoUseInAIKyrios:
The about link in the json text block needs to be updated to: `https://raw.githubusercontent.com/MorpheApp/morphe-branding/refs/heads/main/assets/morphe-logo/morphe_logo.svg` because the link I pasted before this PR isn't quite right.

LisoUseInAIKyrios:
The about reddit link needs an update. I think we're using https://reddit.com/r/MorpheApp"

MarcaDian:
<details>
 <summary>Screenshots</summary>




</details>

patches/src/main/kotlin/app/morphe/patches/youtube/misc/settings/SettingsPatch.kt
```diff
@@ -119,6 +117,8 @@ private val settingsResourcePatch = resourcePatch {
                 "morphe_settings_screen_11_misc_bold.xml",
                 "morphe_settings_screen_12_video.xml",
                 "morphe_settings_screen_12_video_bold.xml",
+                "morphe_settings_screen_99_about.xml",
+                "morphe_settings_screen_99_about_bold.xml",
```
LisoUseInAIKyrios: Do we want the about screen first or last in the settings? I think first is less in the way because it's more convenient to have video as last.

LisoUseInAIKyrios:
I'm moving the json to a web address so the logo works now.

Edit: Done

LisoUseInAIKyrios:
About icon now works. I think the pre-release message doesn't need to be red. If anything we want to encourage users to patch with pre-release if they're comfortable with regularly repatching.

It would be nice if the about screen showed if the patches was outdated, but I think adding that can wait.

MarcaDian:
There some errors
image

inotia00:
This may be off topic, but is there a way to show both the current patch version and the latest patch version in this dialog?

LisoUseInAIKyrios:
Yes it could be added. Since the patches json is a static GitHub file it could easily check.

Maybe we should add this now, with a brief message stating it's up to date or outdated.

LisoUseInAIKyrios:
I added a really simple app version check. The latest version is wrong because the api needs to change to organization repo (will change later), but the logic works.

It also doesn't check if the latest version changes anything for that app. Can improve this later, or change this to a dedicated patch that auto checks and notifies the user if they're using really old or known broken patches (basically turn the junk announcement patch into what it was suppose to be in the first place)

The UI presentation of my commit is just a stub and can be improved.

MarcaDian:
How to correctly fix the absence (or presence) of the letter "v" in a version?

LisoUseInAIKyrios:
It's getting the version from two different places that are handled differently. The current version comes from the patches mpp file manifest (which doesn't have a `v`), and the update version is from the api patches bundle (which has a `v`).

I think the 'v' can be stripped from the patch update version in this dialog. I think it's better for users to reference the version as a pure number without 'v' cutting it up.

Edit: Manager also shows patches versions without a 'v'. So this about dialog should not show any v as well.

LisoUseInAIKyrios:
@MarcaDian Do anything you want with this. Otherwise I think it's ok to merge.

LisoUseInAIKyrios:
I added Crowdin to the about links. 

In the future we may also add a link to the documentations repo that has questions and troubleshooting.

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/MorpheAboutPreference.java
```diff
@@ -59,77 +62,224 @@ private static String useNonBreakingHyphens(String text) {
...
+                             }
+                             .link-button:focus {
+                                 outline: none;
+                             }
+                         </style>
+                        """, backgroundColorHex, foregroundColorHex,
+                morpheBlue, morpheTeal,
+                backgroundColorHex,
+                morpheBlue, morpheTeal,
+                foregroundColorHex
```
inotia00, line 218: The IDE shows the following warning:

```
Too many arguments for format string (found: 8, expected: 6)
```

[According to the first commit of this PR](https://github.com/MorpheApp/morphe-patches/pull/21/changes/1f2c578f0891f980904aef3a5190c1c5c13f156a), it seems like `morpheBlue, morpheTeal,` are duplicated.

MarcaDian: That's right, it's probably left over from the previous review, the button text should be `foregroundColorHex`, and it mistakenly receives the first available argument `morpheBlue` (line 171). And so now the button text is blue and no one even noticed that something was wrong.

LisoUseInAIKyrios: Please make a quick chore commit to fix.

PR #22, opened 2025-12-25
fix(YouTube - Loop video): Fix looping button state

MarcaDian:
If disable looping in the settings during playback, the overlay button did not respond to this

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/videoplayer/LoopVideoButton.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/videoplayer/LoopVideoButton.java
@@ -104,10 +104,8 @@ public class LoopVideoButton {
         PlayerControlButton localInstance = instance;
         if (localInstance == null) return;
 
-        Utils.runOnMainThread(() -> {
-            final boolean currentState = Settings.LOOP_VIDEO.get();
-            localInstance.setIcon(currentState ? LOOP_VIDEO_ON : LOOP_VIDEO_OFF);
-        });
+        final boolean currentState = Settings.LOOP_VIDEO.get();
+        localInstance.setIcon(currentState ? LOOP_VIDEO_ON : LOOP_VIDEO_OFF);
     }
 
     /**


extensions/youtube/src/main/java/app/morphe/extension/youtube/videoplayer/LoopVideoButton.java
```diff
@@ -90,37 +82,64 @@ public static void setVisibilityNegatedImmediate() {
...
+    private static void updateIconFromSettings() {
+        if (instance == null) return;
+
+        try {
+            Utils.runOnMainThread(() -> {
+                final boolean currentState = Settings.LOOP_VIDEO.get();
+                instance.setIcon(currentState ? LOOP_VIDEO_ON : LOOP_VIDEO_OFF);
+            });
+        } catch (Exception ex) {
+            Logger.printException(() -> "updateIconFromSettings failure", ex);
```
LisoUseInAIKyrios: Utils.runOnMainThread already catches exception. A try/catch isn't needed here.

extensions/youtube/src/main/java/app/morphe/extension/youtube/videoplayer/LoopVideoButton.java
```diff
@@ -90,37 +82,64 @@ public static void setVisibilityNegatedImmediate() {
...
+    /**
+     * Update icon based on current setting value.
+     */
+    private static void updateIconFromSettings() {
+        if (instance == null) return;
+
+        try {
+            Utils.runOnMainThread(() -> {
+                final boolean currentState = Settings.LOOP_VIDEO.get();
+                instance.setIcon(currentState ? LOOP_VIDEO_ON : LOOP_VIDEO_OFF);
```
LisoUseInAIKyrios: Store the value of the instance in a local variable (just before the if null check above), just in case the instance field changes

extensions/youtube/src/main/java/app/morphe/extension/youtube/videoplayer/LoopVideoButton.java
```diff
@@ -90,37 +82,61 @@ public static void setVisibilityNegatedImmediate() {
...
+
+    /**
+     * Update icon based on current setting value.
+     */
+    private static void updateIconFromSettings() {
+        PlayerControlButton localInstance = instance;
+        if (localInstance == null) return;
+
+        Utils.runOnMainThread(() -> {
+            final boolean currentState = Settings.LOOP_VIDEO.get();
```
LisoUseInAIKyrios: I just noticed, but is this method already called from the main thread? 

I think the calling methods (set visibility) are both called on the main thread.

If it is on the main thread, then calling runOnMainThread effectively delays running run code to the back of the main thread (and other UI code runs first).

MarcaDian: As far as I know, all methods related to the interface are called in the main thread (and in this case the main thread is the UI thread?)

MarcaDian: I don't like that we call this method every time to check the status, but I don't know how else to do it yet. I really doubt that many users use this button, and notice such trifles when switching from settings

LisoUseInAIKyrios: Try removing the runOnMainThread call (it will then need a try/catch for the code).

If it works as expected then there is no need to call run on main thread as that can cause delays or flickering of the UI since it effectively delays the action.

LisoUseInAIKyrios: If we want to avoid this issue, we could change the setting to require restarting the app.

MarcaDian: The `updateIconFromSettings()` method is called from `setVisibility()` and `setVisibilityImmediate()`, which already have their own exception handling in `PlayerControlButton`.

LisoUseInAIKyrios: Then an additional try/catch isn't needed.

LisoUseInAIKyrios:
Anything to do? Ready to merge?

MarcaDian:
I checked again, it seems that visually and functionally everything works as it should.

PR #23, opened 2025-12-26
feat(Spoof video streams): Add an option to sign-in to Android VR

inotia00:
Adds VR auth sign-in.

This fixes playback issues where the following may not work:
- YouTube Livestreams
- YT Music: Any music if using the app in certain regions
- YT Music: Privately uploaded music

### DIFF
--- a/patches/src/main/resources/addresources/values/shared-youtube/strings.xml
+++ b/patches/src/main/resources/addresources/values/shared-youtube/strings.xml
@@ -65,8 +65,8 @@ To sign in, press continue below"</string>
     <!-- "Open website" must use the same translation as morphe_spoof_video_streams_sign_in_android_vr_activation_code_dialog_open_website -->
     <string name="morphe_spoof_video_streams_sign_in_android_vr_activation_code_dialog_message">"Activation code: %s
 
-1. Tap 'Open website' below to open the 'Connect a device website'
-2. Paste from clipboard (or type out by hand) the activation code shown above
+1. Tap 'Open website' below to open the 'Connect a device' website
+2. Paste from the clipboard (or type out) the activation code shown above
 3. Sign in to Google if prompted. If you have multiple accounts then choose this YouTube account
 4. Tap 'Allow YouTube VR' access
 5. Return to this app after reaching the 'Device connected' message"</string>
@@ -78,15 +78,15 @@ To sign in, press continue below"</string>
 
 • Your Morphe settings will not import/export Android VR authentication
 • If this app is unused for 6 months then Google will remove this authentication
-• You will need to sign in again if you later change your Google password or deregister this device from your Google settings"</string>
+• You will need to sign in again if you change your Google password or deregister this device from your Google settings"</string>
     <string name="morphe_oauth2_connection_failure_timeout">OAuth2 is temporarily not available (API timed out)</string>
     <string name="morphe_oauth2_connection_failure_status_400">Android VR sign in expired, try signing into VR again</string>
     <string name="morphe_oauth2_connection_failure_status">OAuth2 is temporarily not available (status %d)</string>
     <string name="morphe_oauth2_connection_failure_generic">OAuth2 is temporarily not available</string>
-    <string name="morphe_oauth2_connection_failure_auth_not_approved">Cannot sign into VR, authentication not approved</string>
+    <string name="morphe_oauth2_connection_failure_auth_not_approved">Cannot sign into VR, website authentication was not approved</string>
     <string name="morphe_oauth2_connection_failure_auth_error">Cannot sign into VR: %s</string>
     <string name="morphe_oauth2_toast_reset">Logged out of Android VR</string>
-    <string name="morphe_oauth2_toast_invalid">Invalid token, reset the cached token</string>
+    <string name="morphe_oauth2_toast_invalid">Invalid VR sign-in token, try signing into VR again</string>
 
     <!-- misc.audio.forceOriginalAudioPatch -->
     <string name="morphe_force_original_audio_title">Force original audio language</string>


inotia00:
It works, but it needs some cleanup.

Feel free to leave a review.

patches/src/main/resources/addresources/values/shared-youtube/strings.xml
```diff
@@ -45,6 +45,46 @@ If you are a YouTube Premium user, this setting may not be required"</string>
...
+2. Click 'Open website' to go to the verification page.
+3. Enter the activation code.
+4. Sign in when prompted. If you have multiple Google Accounts, select the account you use with YouTube. If you're already signed in, skip to the next step.
+5. Once signed in, return to Morphe and click 'Sign in to Android VR' again.
+6. Click 'Get authorization token' to get an authorization token.
+
+※ Click 'Reset' to remove an authentication token or activation code."</string>
+    <string name="morphe_spoof_video_streams_sign_in_android_vr_dialog_get_activation_code_text">Get activation code</string>
+    <string name="morphe_spoof_video_streams_sign_in_android_vr_dialog_get_authorization_token_text">Get authorization token</string>
+    <string name="morphe_spoof_video_streams_sign_in_android_vr_dialog_reset_text">Reset</string>
```
LisoUseInAIKyrios: If stream fetching fails (livestream/music and the user is in a region where only Android Studio works), then instead of showing a "Could not fetch any streams" toast we can show a dialog with a message that Android VR sign-in is needed and show navigation steps to reach the setting.

extensions/shared/library/src/main/java/app/morphe/extension/shared/oauth2/requests/OAuth2Requester.java
```diff
@@ -0,0 +1,347 @@
...
+                // 3. The refresh token has not been used for 6 months
+                //
+                // In this case, a response code of 400 is returned
+                // Since the refresh token is no longer valid, all locally stored tokens are removed
+                Logger.printDebug(() -> "Invalid token, clear all");
+                clearAll(false);
+            } else {
+                handleConnectionError(str("morphe_oauth2_connection_failure_status", responseCode), null);
+                // TODO: Clear auth token?
+            }
```
LisoUseInAIKyrios, line 332: @inotia00 Where this TODO is, if the response is unexpected (not success or failed), should `authorization` be cleared?

I assume connection problems below the auth should be kept because the connection failing has nothing to do with auth or responses.

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/OAuth2Preference.java
```diff
@@ -0,0 +1,197 @@
...
+                    Utils.showToastShort(str("morphe_spoof_video_streams_sign_in_android_vr_toast_get_authorization_code_failed"));
+                } else {
+                    String refreshToken = accessTokenData.refreshToken;
+                    if (refreshToken != null && !refreshToken.isEmpty()) {
+                        BaseSettings.OAUTH2_REFRESH_TOKEN.save(refreshToken);
+                        OAuth2Requester.setAuthorization(accessTokenData);
+
+                        Context context = getContext();
+                        String dialogTitle =
+                                str("morphe_spoof_video_streams_sign_in_android_vr_success_dialog_title");
```
LisoUseInAIKyrios: After the user gets the auth approved in the external browser and returns to the app, it's not intuitive they need to tap signin preference again.

Ideally, we can show a dialog that shows something like "Press continue after completing the webpage steps", and it fetches the approved auth.

extensions/shared/library/src/main/java/app/morphe/extension/shared/oauth2/requests/OAuth2Requester.java
inotia00: A status code of 400 indicates an invalid refresh token, so it's okay if the refresh token is automatically removed.

For other status codes, it might be better to show a toast message like `The refresh token may be invalid. Try manually resetting the token and reissuing it.`

(It's best to show the toast message only once)

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/OAuth2Preference.java
inotia00: The ideal way to minimize user confusion is to implement WebViewActivity:

Issue an activation code → Copy to clipboard → Launch WebViewActivity → Upon detecting a successful login, finish WebViewActivity → Issue an access token

LisoUseInAIKyrios: Will a web activity require the user to sign into Google again, because it's not the external browser that's already logged in?

extensions/shared/library/src/main/java/app/morphe/extension/shared/oauth2/requests/OAuth2Requester.java
```diff
@@ -0,0 +1,354 @@
...
+            connection.setFixedLengthStreamingMode(body.length);
+            connection.getOutputStream().write(body);
+
+            final int responseCode = connection.getResponseCode();
+
+            if (responseCode == HTTP_STATUS_CODE_SUCCESS) {
+                synchronized (OAuth2Requester.class) {
+                    AccessTokenData fetchedAccessTokenData = new AccessTokenData(refreshToken,
+                            Requester.parseJSONObjectAndDisconnect(connection));
+                    Logger.printDebug(() -> "new accessToken: " + fetchedAccessTokenData);
```
LisoUseInAIKyrios: Just to confirm, this is safe to log and the user shares publicly?

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/OAuth2Preference.java
inotia00: Yes, the user will need to sign in again in the WebView.

LisoUseInAIKyrios: I think it would be less of a hassle if they don't have to sign-in again.

It also presents slightly less concern of asking them to use their google login inside the app.

patches/src/main/resources/addresources/values/shared-youtube/strings.xml
LisoUseInAIKyrios: For now I'll try adding just a long toast message. It will be a lot simpler to add then showing a dialog.

extensions/shared/library/src/main/java/app/morphe/extension/shared/oauth2/requests/OAuth2Requester.java
inotia00: - Access token is used to call the YouTube API. In essence, an access token is a PAT that is valid for one hour.
- Access token expires after one hour.
- Expired access token cannot be used anywhere. (Requests made with an expired access token will return a response code of 400 or 403)
- Refresh token allows access token re-issuance, but refresh token do not expire. (So refresh token is excluded from debug logs)

inotia00: Since it's only valid for one hour, I included it in the debug log.

For strict security, it's better to disallow even that.

(Or you can mask out only the last few characters so that they print in the debug log)

LisoUseInAIKyrios: Ok then I'll remove it from the logs.

extensions/shared/library/src/main/java/app/morphe/extension/shared/oauth2/requests/OAuth2Requester.java
LisoUseInAIKyrios: I added a "Try signing in to VR again" toast that is shown only once (when the token is cleared)

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/OAuth2Preference.java
inotia00: We can implement something as similar as possible using CustomTabsClient instead of WebView.

If the `Open links externally` patch is excluded, links opened on YouTube will open in the in-app browser, which is CustomTabsClient.

### Pros

- Since the web browser is used as is, the user is likely already signed in to their Google account.
- The app can detect when the in-app browser is closed and proceed to the next step (issuing an access token).

#### Cons

- The user must manually close the in-app browser.
- There's no way to verify that the user successfully linked to YouTube VR.

LisoUseInAIKyrios: This sounds better because it'll remove the confusing last step of taping the sign-in preference again 

> There's no way to verify that the user successfully linked to YouTube VR.

Is this detectable by blindly trying the next step (get the auth token)?  I think we can ignore this situation if needed.

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/OAuth2Preference.java
```diff
@@ -0,0 +1,197 @@
...
+import app.morphe.extension.shared.oauth2.requests.OAuth2Requester;
+import app.morphe.extension.shared.settings.BaseSettings;
+import app.morphe.extension.shared.ui.CustomDialog;
+
+@SuppressWarnings({"FieldCanBeLocal", "deprecation", "unused"})
+public class OAuth2Preference extends Preference implements Preference.OnPreferenceClickListener {
+
+    private void init() {
+        setOnPreferenceClickListener(this);
+    }
```
LisoUseInAIKyrios, line 147: The summary text could be different if a sign-in token is saved.

After going thru the setup steps, the preference still says "Click here to see how to sign in to Android VR"

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/OAuth2Preference.java
LisoUseInAIKyrios: Another idea is to add a hook for the app resuming, and do the second part of the auth step automatically when the user returns to the app.

This would be the simplest of changes because the patched app handles pressing the button for you. If the auth doesn't work then the user must not have done something right (didn't login if needed, didn't finish the steps), and we could show a different dialog.

LisoUseInAIKyrios: For now, I changed it so after going thru the website and the app is resumed, it automatically completes the auth process.

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/OAuth2Preference.java
```diff
@@ -0,0 +1,232 @@
...
+
+    @Override
+    public boolean onPreferenceClick(Preference preference) {
+        Context context = getContext();
+        String dialogTitle = str("morphe_spoof_video_streams_sign_in_android_vr_about_title");
+        String dialogMessage = str("morphe_spoof_video_streams_sign_in_android_vr_dialog_message");
+        String resetButtonText = str("morphe_spoof_video_streams_sign_in_android_vr_dialog_reset_text");
+        String okButtonStringKey;
+        Runnable okButtonRunnable;
+        if (isActivationCodeDataAvailable()) {
```
LisoUseInAIKyrios: If the user is currently signed-in (and everything appears to be valid), this dialog could show a much shorter different text. Something like:

"Android VR sign-in successful. To remove sign-in press "reset" to continue".
and the only buttons are "Reset" and "Ok" (closes the dialog).

LisoUseInAIKyrios: Dialogs adjusted to show different content based on the current signed in or not signed in state.

LisoUseInAIKyrios:
The UI now shows the following flow:

<details>
<summary>Sign-in flow</summary>
sign in 1
sign in 2
sign in 3

### After user approves Google add device (Shown automatically)
sign in 4
sign in 5

### If user taps "Sign-in to Android VR" while signed in
sign in 6

</details>

inotia00:
Looks good, and it's better than the way RVX is implemented:

1. It safely handles sign-out using the `/revoke` endpoint.
2. It's thread-safe, so access tokens are safely refreshed at the intended interval.
3. If the refresh token is invalid, the app shows a toast message `Android VR sign in expired, try signing into VR again` and clears the saved token.

3-1. If the Google account password is changed, the `GmsCore support` patch simply shows this message:

<details>
<summary>refresh_token failed (status code 400)</summary>
account
</details>

3-2. This PR shows a toast message `Android VR sign in expired, try signing into VR again` and handles sign-out.

PR #37, opened 2026-01-01
feat(YouTube - Auto-captions): Add an option to disable auto-captions when the device volume is muted

0xrxL:


0xrxL:
@LisoUseInAIKyrios Has been more easier for me to open a new one, keeping the changes from the old one and adding the mode selector suggestion

extensions/youtube/src/main/java/app/morphe/extension/youtube/settings/Settings.java
```diff
@@ -143,6 +144,7 @@ public class Settings extends BaseSettings {
     public static final BooleanSetting COPY_VIDEO_URL = new BooleanSetting("morphe_copy_video_url", FALSE);
     public static final BooleanSetting COPY_VIDEO_URL_TIMESTAMP = new BooleanSetting("morphe_copy_video_url_timestamp", TRUE);
     public static final BooleanSetting DISABLE_AUTO_CAPTIONS = new BooleanSetting("morphe_disable_auto_captions", FALSE, true);
+    public static final EnumSetting<DisableAutoCaptionsPatch.DisableAutoCaptionsStyle> DISABLE_AUTO_CAPTIONS_STYLE = new EnumSetting<>("morphe_disable_auto_captions_style", DisableAutoCaptionsPatch.DisableAutoCaptionsStyle.KEEP_BOTH, false);
```
LisoUseInAIKyrios: Can this enum and `BooleanSetting("morphe_disable_auto_captions", FALSE, true)` be merged together? So the enum setting controls everything? It would be one less setting.

patches/src/main/kotlin/app/morphe/patches/youtube/layout/autocaptions/AutoCaptionsPatch.kt
```diff
@@ -17,14 +19,15 @@ val autoCaptionsPatch = bytecodePatch(
     dependsOn(
         sharedExtensionPatch,
         settingsPatch,
+        versionCheckPatch
```
LisoUseInAIKyrios, line 24: I will check tomorrow, but the client flags may be present in earlier app targets.

extensions/youtube/src/main/java/app/morphe/extension/youtube/settings/Settings.java
0xrxL: > Can this enum and `BooleanSetting("morphe_disable_auto_captions", FALSE, true)` be merged together? So the enum setting controls everything? It would be one less setting.

Yes, I'll merge it asap.

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/DisableAutoCaptionsPatch.java
```diff
@@ -1,17 +1,58 @@
 package app.morphe.extension.youtube.patches;
 
+import androidx.annotation.Nullable;
 import app.morphe.extension.youtube.settings.Settings;
 
 @SuppressWarnings("unused")
 public class DisableAutoCaptionsPatch {
+    public enum DisableAutoCaptionsStyle {
```
LisoUseInAIKyrios: Since this can now enable auto captions when the volume is muted, maybe rename this patch to `AutoCaptionsPatch`

patches/src/main/kotlin/app/morphe/patches/youtube/layout/autocaptions/AutoCaptionsPatch.kt
LisoUseInAIKyrios: It's present in 20.26+

LisoUseInAIKyrios:
@0xrxL I added support for 20.14 and 20.21 (has only the single "hide auto captions", which I presume also hides when muted).

Can you check if this all works? I don't ever get auto captions so I can't verify what's here.

Also, you need to reset to your repo branch because I had to force push.

0xrxL:
> @0xrxL I added support for 20.14 and 20.21 (has only the single "hide auto captions", which I presume also hides when muted).
> 
> Can you check if this all works? I don't ever get auto captions so I can't verify what's here.
> 
> Also, you need to reset to your repo branch because I had to force push.

I'll check tomorrow.

0xrxL:
@LisoUseInAIKyrios There was some problems here:

1) A fingerprint was wrong, than I restored it.
2) Enum and Arrays.xml wasn't formatted in the same way (so the selected options wasn't effectively the same described)..
3) captionsButtonStatus field was used wrongly where needed, and added where doesn't.

Now everything works as expected.

LisoUseInAIKyrios:
Just say when it's ready to merge.

0xrxL:
> Just say when it's ready to merge.

It's already. You can merge it.

PR #41, opened 2026-01-02
fix: Remove restrictions on AGP due to the use of hardcoded registers in LithoFilterPatch

inotia00:


settings.gradle.kts
```diff
@@ -20,7 +20,7 @@ pluginManagement {
 }
 
 plugins {
-    id("app.morphe.patches") version "1.0.0"
+    id("app.morphe.patches") version "1.0.1"
```
inotia00, line 23: This PR depends on https://github.com/MorpheApp/morphe-patches-gradle-plugin/pull/4 and may be merged after gradle plugin 1.0.1 is released.

patches/src/main/kotlin/app/morphe/patches/youtube/misc/litho/filter/LithoFilterPatch.kt
```diff
@@ -87,18 +93,46 @@ val lithoFilterPatch = bytecodePatch(
...
+                    getInstruction<OneRegisterInstruction>(insertIndex).registerA
 
-            addLithoFilter = { classDescriptor ->
                 addInstructions(
-                    2,
-                    """
-                        new-instance v1, $classDescriptor
-                        invoke-direct { v1 }, $classDescriptor-><init>()V
-                        const/16 v2, ${filterCount++}
-                        aput-object v1, v0, v2
```
inotia00, line 100: FYI: The use of hardcoded registers forces the use of AGP 8.2.2.
(These lines were written about two years ago)

Since AGP 8.3.0, the bytecode generation method has changed slightly, so using hardcoded registers causes a Java verifier error.

The easiest workaround is to use a helper method.

This workaround works on both AGP 8.2.2 and 8.3.0+.

LisoUseInAIKyrios (approved):

patches/src/main/kotlin/app/morphe/patches/youtube/misc/litho/filter/LithoFilterPatch.kt
wchill: Do you have more context on this issue? If the bytecode generation breaks on hardcoded registers, won't other patches be affected as well?

LisoUseInAIKyrios: This is patching extension code and not the target app.

The hard coded registers were because the old patch was assuming the compiler generated code would always look a certain way and always use certain registers.

This is a unique situation and this kind of behavior isn't found in any other patches. This issue only affects patching extension code (and not code in a target APK).

All other patches should be ok.

settings.gradle.kts
LisoUseInAIKyrios: This fix avoids the issues with the old AGP library and works with both [old and new AGP](https://github.com/MorpheApp/morphe-patches/pull/41#discussion_r2656966627).  So this PR can be merged now and merge plugin AGP dependency change later.

PR #145, opened 2026-01-09
fix(YouTube - Auto captions): Cannot disable auto captions with 20.21 app target

0xrxL:


patches/src/main/resources/addresources/values/youtube/arrays.xml
```diff
@@ -135,8 +135,8 @@
         <item>@string/morphe_auto_captions_style_legacy_entry_2</item>
     </string-array>
     <string-array name="morphe_auto_captions_style_legacy_entry_values">
-        <item>DISABLE_BOTH</item>
-        <item>KEEP_BOTH</item>
+        <item>ENABLED</item>
+        <item>DISABLED</item>
```
LisoUseInAIKyrios: This will try to set the `morphe_auto_captions_style` enum setting to this style which doesn't yet exist.

Instead of adding a legacy enum, add legacy enum declarations to the existing style enum and then this works.

0xrxL: Oh...in settings.java then. Ok

extensions/youtube/src/main/java/app/morphe/extension/youtube/settings/Settings.java
```diff
@@ -147,6 +147,7 @@ public class Settings extends BaseSettings {
     public static final BooleanSetting COPY_VIDEO_URL = new BooleanSetting("morphe_copy_video_url", FALSE);
     public static final BooleanSetting COPY_VIDEO_URL_TIMESTAMP = new BooleanSetting("morphe_copy_video_url_timestamp", TRUE);
     public static final EnumSetting<AutoCaptionsStyle> AUTO_CAPTIONS_STYLE = new EnumSetting<>("morphe_auto_captions_style", AutoCaptionsStyle.BOTH_ENABLED, false);
+    public static final EnumSetting<AutoCaptionsStyleLegacy> AUTO_CAPTIONS_STYLE_LEGACY = new EnumSetting<>("morphe_auto_captions_style", AutoCaptionsStyleLegacy.ENABLED, false);
```
LisoUseInAIKyrios: Legacy should have a different key: `morphe_auto_captions_style_legacy`

0xrxL:
@LisoUseInAIKyrios Just to be clear about the several (useless) commits here: I literally created this PR with my phone (because, as I said somewhere else, I'm busy these days). 😁

LisoUseInAIKyrios:
Lol, that's ok. I will try 20.21 later and verify as much as I can.

0xrxL:
> Lol, that's ok. I will try 20.21 later and verify as much as I can.

Should work. The logic is practically the same for both versions, but just with a different Enum.

patches/src/main/resources/addresources/values/youtube/arrays.xml
```diff
@@ -135,8 +135,8 @@
         <item>@string/morphe_auto_captions_style_legacy_entry_2</item>
     </string-array>
     <string-array name="morphe_auto_captions_style_legacy_entry_values">
-        <item>DISABLE_BOTH</item>
-        <item>KEEP_BOTH</item>
+        <item>BOTH_ENABLED</item>
+        <item>BOTH_DISABLED</item>
```
LisoUseInAIKyrios, line 139: I think this is the only fix actually needed. It was using the wrong enum names.

Issue #148, opened 2026-01-09
bug(YouTube): "Hide creator store shelf" can hide 'You' tab components


ILoveOpenSourceApplications:
### Bug description

https://github.com/user-attachments/assets/28612958-754b-4491-bd58-40bc1d5917ba

From what I've seen it looks like the thread pool size issue from back in the day. But I'm unsure about this. Going back and forth between Subscriptions and Home seems to reload the You tab back to normal.

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

LisoUseInAIKyrios:
Whatever the bug is, it's something with YouTube itself because this can happen on unpatched too.

It's probably a thread race in YTs own code and the patch hide checks are adding just enough latency for the bug shows up more often. Changing the thread pool size is basically a bandaid to avoid the conditions instead of YT actually fixing the bug.

In 21.02 I saw there are new places where thread executors are being passed around, and the executor was not in 20.51. So maybe these need fixes too.

LisoUseInAIKyrios:
@ILoveOpenSourceApplications try turning off all the debug flags and see if that fixes it.

ILoveOpenSourceApplications:
No, it didn't.

LisoUseInAIKyrios:
Actually, I now remember the thread pool size was set by some different offline client flags that isn't hooked by the debug patch. Will look later if there seems to be anything new.

LisoUseInAIKyrios:
And just to confirm, this issue first showed up with 21.02?

ILoveOpenSourceApplications:
No, it was there in `v20.37.48` as well.

LisoUseInAIKyrios:
If it's convenient, try older versions for a little while until you find which 20.x release the issue first appears in. Only need to check 20.36.46, 20.35.34, 20.34.37, etc.

ILoveOpenSourceApplications:
The issue is caused by "Hide creator store shelf". Somehow `horizontal_shelf` is getting triggered and causing all this. When turning it off the issue doesn't happen. With channel page open and playing a video with store shelf, it even hides the "For You" shelf, even if for a limited time.

ILoveOpenSourceApplications:
This was why I couldn't replicate the issue consistently because it only occurred on videos with the store shelf below the player. Once the culprit was identified, it's top easy to replicate. 

[...]

PR #171, opened 2026-01-10
refactor(YouTube - Hide layout components): Consolidate hide horizontal shelves code into a single location

LisoUseInAIKyrios:
Consolidates all horizontal shelf hiding into a single code location, which is required for an attempted fix of the player shopping shelf that sometimes shows when resuming the app.

This change should also fix the player description components sometimes incorrectly hiding if "Hide horizontal shelves" is enabled and the player video description is opened then immediately minimized.

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/LayoutComponentsFilter.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/LayoutComponentsFilter.java
@@ -24,6 +24,8 @@ import app.morphe.extension.shared.ByteTrieSearch;
 import app.morphe.extension.shared.Logger;
 import app.morphe.extension.shared.StringTrieSearch;
 import app.morphe.extension.shared.Utils;
+import app.morphe.extension.shared.settings.BooleanSetting;
+import app.morphe.extension.shared.settings.Setting;
 import app.morphe.extension.youtube.patches.ChangeHeaderPatch;
 import app.morphe.extension.youtube.settings.Settings;
 import app.morphe.extension.youtube.shared.NavigationBar;
@@ -349,15 +351,17 @@ public final class LayoutComponentsFilter extends Filter {
                 new Pair<>(Settings.HIDE_QUIZZES_SECTION, "post_base_wrapper_slim"),
                 // May no longer work on v20.31+, even though the component is still there.
                 new Pair<>(Settings.HIDE_ATTRIBUTES_SECTION, "cell_video_attribute")
-        ).forEach(pair ->
-                descriptionSearch.addPattern(pair.second.getBytes(StandardCharsets.UTF_8),
-                        (textSearched, matchedStartIndex, matchedLength, callbackParameter) -> {
-                            //noinspection unchecked
-                            AtomicReference<Boolean> hide = (AtomicReference<Boolean>) callbackParameter;
-                            hide.set(pair.first.get());
-                            return true;
-                        }
-                )
+        ).forEach(pair -> {
+                    BooleanSetting setting = pair.first;
+                    descriptionSearch.addPattern(pair.second.getBytes(StandardCharsets.UTF_8),
+                            (textSearched, matchedStartIndex, matchedLength, callbackParameter) -> {
+                                //noinspection unchecked
+                                AtomicReference<Boolean> hide = (AtomicReference<Boolean>) callbackParameter;
+                                hide.set(setting.get());
+                                return true;
+                            }
+                    );
+                }
         );
 
         addPathCallbacks(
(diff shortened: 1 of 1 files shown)


LisoUseInAIKyrios:
@ILoveOpenSourceApplications Can you try this PR for a few days and see if everything seems working?

I can't find any gaming/music/places shelves in any videos and this could use some verification.

ILoveOpenSourceApplications:
I'll test it, sure.

ILoveOpenSourceApplications:
This could possibly resolve my issue #148.

ILoveOpenSourceApplications:
Issues found so far:

1) If channel page is opened in the background and a video with store shelf is played, playing it in fullscreen and then as a miniplayer brings out the store shelf in channel page home. Refreshing the page hides it. This doesn't happen in the current iteration. 
2) Issues from #148 are still there.
3) Ticket shelf content(2) can only be viewed when both "Hide creator store shelf" and "Hide ticket shelf" are off. If the later is only off, the content remains hidden and the header(1) remains. 
Screenshot_20260111_021525_YouTube

LisoUseInAIKyrios:
I overlooked that ticket shelf appears inside the player.

That makes fixing the player shopping shelf using the player type not possible. A different fix is needed.

ILoveOpenSourceApplications:
Basically, everything happens when the player goes into fullscreen and coming out of it.

LisoUseInAIKyrios:
If you have time try again with the change I just made.

This may still fix a potential issue of the player description components incorrectly hiding if hide horizontal shelves is on (especially if the description is opened and the player immediately minimized at the same time), but that's only speculative and it may not happen in practice.

LisoUseInAIKyrios:
Actually, this approach to fix shopping shelf might still work because only the maximized (not fullscreen) needs to be checked.

ILoveOpenSourceApplications:
@inotia00 has a file to check and see if the description is open. That could help in the case of description. But the case of player, I'm unsure if there's any other workaround.

ILoveOpenSourceApplications:
Why is this issue happening in the first place? Is it because the buffer components `shopping_item_card_list` doesn't return when going into fullscreen and coming out?

LisoUseInAIKyrios:
It's because another horizontal shelf is being loaded somewhere else, and due to the 20.31+ workaround it's using the buffer of a different horizontal shelf. So the buffer has the wrong contents in it.

The underlying issue is when a component is dragged off screen then on (or set invisible then visible), litho doesn't parse the java buffer again and it seems to be using C++ code to manage the buffer. So, the hook for the buffer is never called because the buffer data is already in C++ land where no patches currently work.

LisoUseInAIKyrios:
I spent a few full days trying to figure out a different solution, including trying to associated the refreshed component creation with the original buffer used for the initial creation of the component. But it seems reused components are re-created every time (at least, everything within arms reach from the existing code hook is recreated).

It's a mess and the current solution only exists because the buffer almost always has the identifier very close to the start of it, and the identifier is parsed out of it which is then used as a key to lookup the buffer when the refreshed component is created.

ILoveOpenSourceApplications:
Then what about adding exceptions per every horizontal setting? Is that possible? What I meant is, if it's "Hide ticket shelf" then, all the other buffers of `horizontal_shelf.e` are exempted except `ticket_item` and so on for others?

LisoUseInAIKyrios:
Basically that's already going on now. a filter checks for each shelf item (playables for example), and if the filter matches then it returns the current show/hide for that setting value. This logic acts as both an exception check and the filter check.

ILoveOpenSourceApplications:
What about using browser ID?

LisoUseInAIKyrios:
Great suggestion. Will have to research more.

ILoveOpenSourceApplications:
Can check out rvx repo for a head start.

inotia00:
### Shopping shelf sometimes doesn't hide when exiting fullscreen mode.

When exiting fullscreen mode with a swipe gesture, the PlayerType changes in the following order:

> [!Note]
> WATCH_WHILE_FULLSCREEN → WATCH_WHILE_SLIDING_MAXIMIZED_FULLSCREEN → WATCH_WHILE_MAXIMIZED

https://github.com/MorpheApp/morphe-patches/blob/7b631dfb71054c7ea0c8d455a42bfcbf42a1b9e9/extensions/youtube/src/main/java/app/morphe/extension/youtube/shared/PlayerType.kt#L138-L140

Currently `isMaximizedOrFullscreen()` only checks `WATCH_WHILE_MAXIMIZED` and `WATCH_WHILE_FULLSCREEN`, so the shopping shelf may not hide when current PlayerType is `WATCH_WHILE_SLIDING_MAXIMIZED_FULLSCREEN`.

### Shopping shelf sometimes doesn't hide when resuming the app.

This is already explained in https://github.com/MorpheApp/morphe-patches/issues/432#issuecomment-3852733155.

LisoUseInAIKyrios:
The shopping shelf currently only looks in the buffer, because checking the player type can fail if a video is opened then immediately minimized before the shopping shelf component can load.

LisoUseInAIKyrios:
I can no longer reproduce the shopping shelf showing when using this PR.

If this PR doesn't fix it, it still consolidates all shopping shelf logic into a single place which fixes one hide setting can incorrectly hide an unrelated shelf.

Issue #237, opened 2026-01-18
bug(YouTube): Playback buffering / does not work


tanvir8282:
### Bug description

21.03.34 has playback issues...its just loading never plays..tried different spoof clients...vpns too

![Image](https://github.com/user-attachments/assets/84f28098-0cf8-4dca-8748-13e6f572b96d)
![Image](https://github.com/user-attachments/assets/8aa70a3b-8bc1-443f-84fd-5bfb7ca81c78)

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

LisoUseInAIKyrios:
Try VR sign-in.

LisoUseInAIKyrios:
Also try logging into a different YT account.

And try uninstalling patched YouTube and try patching `20.37.48` and `20.14.43`. 

All versions above `20.37.48` are experimental because they're too new to be well tested.

tanvir8282:
> Also try logging into a different YT account.
> 
> And try uninstalling patched YouTube and try patching `20.37.48` and `20.14.43`.
> 
> All versions above `20.37.48` are experimental because they're too new to be well tested.

Ohh...Im using 21.02.35 with 1.6.0 stable patches..
I was just trying this ans its youtubes Beta release also..



And One feedback from me...I was having Video auto vanishing in homepage after clicking it on previous patches ...i also attached videos,...
This issue Seems to have fixed in 1.6.0 patches..



Thanks for the update bro 🫰

tanvir8282:
> Try VR sign-in.
<details>

![Image](https://github.com/user-attachments/assets/0a858d32-9f85-496c-8210-69bfc3da76bc)
![Image](https://github.com/user-attachments/assets/d7b4b5d9-4849-4e99-a5a3-4ab963e0d701)

</details>

Same issues

LisoUseInAIKyrios:
Try turning off all debug flags:

1. Settings > Morphe > Miscellaneous > Debugging > Enable debugging
2. Force close the app
3. Open the app, open a video that fails
4. Go to: Settings > Morphe > Miscellaneous > Debugging > Debug flag manager
5. Select all flags (upper left button)
6. Add all flags (lower left >> button)
7. Save, restart
8. Check same video again


If this changes nothing, then try other steps above (patch old app version, use a different YT account).

tanvir8282:
> Try turning off all debug flags:
> 
[...]

PR #249, opened 2026-01-21

inotia00:
### Android No SDK clients no longer support sign-in.

Android No SDK clients support sign-in, but access tokens issued by some MicroG appear to be incompatible.

Implementing the `Android No SDK sign-in` setting can resolve the issue, but the code becomes significantly more complex.

As a temporary workaround, Android No SDK is treated as client that do not allow sign-in.

Assume the user has set `Android No SDK` as their default client.

Most videos will play in `Android No SDK`.

For age-restricted or purchased videos, if the user obtained an access token through `Android VR sign-in`, the video will play in `Android VR`; otherwise, it will play in `Android Studio`.

### Fixed an issue where the patch was not working on YouTube 21.03+.

Added a missing commit when adding YouTube 21.03+ support (Closes https://github.com/MorpheApp/morphe-patches/issues/237).

### DIFF
--- a/patches/src/main/kotlin/app/morphe/patches/shared/misc/spoof/Fingerprints.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/shared/misc/spoof/Fingerprints.kt
@@ -116,6 +116,7 @@ internal object CreateStreamingDataFingerprint : Fingerprint(
 
 internal object BuildMediaDataSourceFingerprint : Fingerprint(
     accessFlags = listOf(AccessFlags.PUBLIC, AccessFlags.CONSTRUCTOR),
+    returnType = "V",
     parameters = listOf(
         "Landroid/net/Uri;",
         "J",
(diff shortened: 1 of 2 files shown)


0xrxL:
Inotia00 = Captain America with Mjölnir. 😁

error-reporting:
is access tokens issued by microg-re by WSTxda considered incompatible with Android No SDK client?

LisoUseInAIKyrios:
Not sure if it's the same thing mentioned here, but very old microG installs can have different access tokens that behave differently and may or may not work right with spoof streams today.

Meaning that if an old access token is on a microG installation, clearing the microG app data will refresh the token and the token behavior may then be different.

error-reporting:
@inotia00 @LisoUseInAIKyrios access tokens created by wstxda microg-re works fine for android no sdk client logged in

LisoUseInAIKyrios:
Other clients work where authentication is needed, and if no SDK isn't available the spoof stream patch automatically falls over to one of those.

error-reporting:
i mean that android no sdk client still works with auth for me and the no auth change should not be done

LisoUseInAIKyrios:
Some users are starting to report No SDK is failing for them.

In the past other spoof clients first started to fail with auth for certain users, then later started to fail for all auth users. The same series of events likely is playing out again for No SDK.

oyugio:
<details>

Screenshot_20260122_165241_YouTube
Screenshot_20260122_165131_Settings_resized
</details>
👍

error-reporting:
what's the difference between revanced/morphe Spoof video streams and rvx spoof streaming data patch?

Issue #129, opened 2026-01-07
bug(YouTube): Ads shown between Shorts while swiping


retta1:
### Bug description

I'm getting ads while swiping down shorts like this also they have no buttons like other shorts videos

<img width="1080" height="2136" alt="Image" src="https://github.com/user-attachments/assets/ee518d26-86c7-48bd-8024-bb31a55b54bf" />
<img width="1080" height="2136" alt="Image" src="https://github.com/user-attachments/assets/da27c55e-f4bf-4427-adfc-4398210e7346" />

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

LisoUseInAIKyrios:
You can enable `Switch Create with Notifications` and it _maybe_ will turn off these Shorts ads.

Issue has been present before Morphe existed and only happens to some users in certain regions of the world.

retta1:
> You can enable `Switch Create with Notifications` and it _maybe_ will turn off these Shorts ads.
> 
> Issue has been present before Morphe existed and only happens to some users in certain regions of the world.

i did i think it fixed but create button is there and I cannot hide it even tho I enabled hide create button in settings 

<img width="1080" height="153" alt="Image" src="https://github.com/user-attachments/assets/6dd41f5f-f99b-43f1-8e01-eaeb96f7f37d" />

BlackGold8282:
To hide that button, 'Hide Create button in toolbar' is required, not 'Hide Create button in navigation bar'.
Morphe doesn't have that feature yet.

Figim:
> You can enable `Switch Create with Notifications` and it _maybe_ will turn off these Shorts ads.
> 
> Issue has been present before Morphe existed and only happens to some users in certain regions of the world.

It still shows ads sometimes

LisoUseInAIKyrios:
That's why it's: _maybe_ will turn off.

Issue #117, opened 2026-01-06
bug(Youtube): Layout form factor "Tablet" display errors at start up most of the time


NekoZ410:
### Bug description

YouTube: 20.37.48
Patch: 1.0.0
Patch Method: Manager
Selected patches: Default, customized app name, customized app dark/light theme color

When the app starts, it only displays one video at a time and requires a manual reload to show other videos normally.

<img width="1080" height="2280" alt="Image" src="https://github.com/user-attachments/assets/57015fb6-e95e-4f81-af3a-581e3776a181" />

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

LisoUseInAIKyrios:
I don't have any issues with tablet form factor, but I also don't notice any difference.

Try automotive form factor.

andrix02:
layout have some empty spaces when we use tablet device. I tried automotive form factor and tablet, but it was same

l205306:
https://www.reddit.com/r/MorpheApp/comments/1q65yyv/youtube_not_optimized_for_tablets/

Same issue?

NekoZ410:
> I don't have any issues with tablet form factor, but I also don't notice any difference.
> 
> Try automotive form factor.

Automotive layout make me feels like browsing across shelves of goods so I don't want to use it.
I just want to keep "infinite scrolling" experience (mobile layout), but still want a padding for each item to avoid accidental touch when using phone with curved screen edge (tablet layout, experienced on ReVanced).

Btw, tablet layout on Morphe don't have padding anymore :(((((

Morphe:

<img width="1080" height="2280" alt="Image" src="https://github.com/user-attachments/assets/eb818517-1330-47a7-be79-5a5c216d3e06" />

ReVanced:

<img width="1080" height="2280" alt="Image" src="https://github.com/user-attachments/assets/f297aafa-c938-45de-aabf-057eb7444516" />

LisoUseInAIKyrios:
Try patching `20.14.43` or `20.21.37`

It appears the tablet override logic has changed in newer versions.

If anyone finds the last version of YT where the tablet layout setting works, please post your findings here.

NekoZ410:
> Try patching `20.14.43` or `20.21.37`
> 
> It appears the tablet override logic has changed in newer versions.
> 
> If anyone finds the last version of YT where the tablet layout setting works, please post your findings here.

[...]

PR #260, opened 2026-01-23
fix(YouTube - Hide ads): `Hide video ads` does not hide Shorts ads

inotia00:
### Before PR

To hide Shorts ads, users had to enable the `Switch Create with Notifications` setting.

### After PR

Shorts ads will only be hidden when the `Hide video ads` setting is enabled (Closes https://github.com/MorpheApp/morphe-patches/issues/129).

### DIFF
--- a/patches/src/main/resources/addresources/values/youtube/strings.xml
+++ b/patches/src/main/resources/addresources/values/youtube/strings.xml
@@ -601,7 +601,7 @@ Adjust volume by swiping vertically on the right side of the screen"</string>
     <string name="morphe_hide_notifications_button_summary_off">Notifications button is shown</string>
     <!-- 'Notifications' should be translated using the same localized wording YouTube displays for the tab. -->
     <string name="morphe_switch_create_with_notifications_button_title">Switch Create with Notifications</string>
-    <string name="morphe_switch_create_with_notifications_button_summary_on">"Create button is switched with Notifications button"</string>
+    <string name="morphe_switch_create_with_notifications_button_summary_on">Create button is switched with Notifications button</string>
     <string name="morphe_switch_create_with_notifications_button_summary_off">Create button is not switched with Notifications button</string>
     <string name="morphe_switch_create_with_notifications_button_user_dialog_message">If changing this setting does not take effect, try switching to Incognito mode.</string>
     <string name="morphe_hide_navigation_button_labels_title">Hide navigation button labels</string>


patches/src/main/kotlin/app/morphe/patches/youtube/misc/contexthook/ClientContextHookPatch.kt
```diff
@@ -0,0 +1,177 @@
...
+private lateinit var clientVersionField: FieldReference
+private lateinit var messageLiteBuilderField: FieldReference
+private lateinit var messageLiteBuilderMethod: MethodReference
+private lateinit var osNameField: FieldReference
+
+enum class Endpoint(
+    vararg val parentFingerprints: Fingerprint,
+    var smaliInstructions: String = "",
+) {
+    BROWSE(BrowseEndpointParentFingerprint),
```
inotia00, line 30: There are unused enums or functions.

These will be used in an upcoming PR.

patches/src/main/kotlin/app/morphe/patches/youtube/misc/contexthook/ClientContextHookPatch.kt
```diff
@@ -0,0 +1,177 @@
...
+        val helperMethodName = "patch_setClientContext"
+
+        Endpoint.entries.filter {
+            it.smaliInstructions.isNotEmpty()
+        }.forEach { endpoint ->
+            endpoint.parentFingerprints.forEach { parentFingerprint ->
+                // For some reason, declaring this in Fingerprints.kt always returns the same class.
+                // This seems to be due to the method cache in Patcher.
+                // As a workaround, declare the fingerprint in the patch.
+                val EndpointRequestBodyFingerprint = Fingerprint(
```
LisoUseInAIKyrios: If I understand right, it's because the fingprinting is storing the match result. There is `Fingerprint.clearMatch()` that could be used here since this fingerprint is used as a filter for multiple other fingerprints, but declaring it locally here also works.

patches/src/main/kotlin/app/morphe/patches/youtube/misc/contexthook/ClientContextHookPatch.kt
```diff
@@ -0,0 +1,177 @@
...
+                            )
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+
+fun addClientVersionHook(endPoint: Endpoint, descriptor: String) {
```
LisoUseInAIKyrios, line 150: Method is currently unused. I assume this will be used for another upcoming change.

inotia00: This function will be used in an upcoming PR.

In the latest YouTube, the Squircle thumbnail is no longer used in the Home and Subscriptions feeds when the device is a tablet and in portrait mode: https://github.com/MorpheApp/morphe-patches/issues/117.

This is related to the app version, and can be resolved using `Spoof app version`.

Using `Spoof app version` changes the app version across all endpoints, which can have unintended side effects, such as disabling the bold icon.

The `addClientVersionHook` function allows you to change the app version only for the feeds you want.

I'm currently developing a feature that uses this to spoof the app version only when loading the Home and Subscriptions feeds.

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -601,13 +601,9 @@ Adjust volume by swiping vertically on the right side of the screen"</string>
...
     <string name="morphe_switch_create_with_notifications_button_title">Switch Create with Notifications</string>
-    <string name="morphe_switch_create_with_notifications_button_summary_on">"Create button is switched with Notifications button
-
-Note: Enabling this also forcibly hides video ads"</string>
+    <string name="morphe_switch_create_with_notifications_button_summary_on">"Create button is switched with Notifications button"</string>
     <string name="morphe_switch_create_with_notifications_button_summary_off">Create button is not switched with Notifications button</string>
-    <string name="morphe_switch_create_with_notifications_button_user_dialog_message">"Disabling this setting will also disable Shorts ad blocking.
-
-If changing this setting does not take effect, try switching to Incognito mode."</string>
+    <string name="morphe_switch_create_with_notifications_button_user_dialog_message">"If changing this setting does not take effect, try switching to Incognito mode."</string>
```
```suggestion
    <string name="morphe_switch_create_with_notifications_button_user_dialog_message">If changing this setting does not take effect, try switching to Incognito mode.</string>
```

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -601,13 +601,9 @@ Adjust volume by swiping vertically on the right side of the screen"</string>
     <string name="morphe_hide_notifications_button_summary_off">Notifications button is shown</string>
     <!-- 'Notifications' should be translated using the same localized wording YouTube displays for the tab. -->
     <string name="morphe_switch_create_with_notifications_button_title">Switch Create with Notifications</string>
-    <string name="morphe_switch_create_with_notifications_button_summary_on">"Create button is switched with Notifications button
-
-Note: Enabling this also forcibly hides video ads"</string>
+    <string name="morphe_switch_create_with_notifications_button_summary_on">"Create button is switched with Notifications button"</string>
```
```suggestion
    <string name="morphe_switch_create_with_notifications_button_summary_on">Create button is switched with Notifications button</string>
```

Issue #113, opened 2026-01-06
bug(YouTube): Disable bold icons sometimes not working on Action buttons,


m3trixix:
### Bug description

![Image](https://github.com/user-attachments/assets/2136bbf7-1a84-4cef-ab07-c82efac5f71f)
### Bug description
Action buttons will sometimes not have bold icons (except for the Like button).

### Temporary action

- [x] I tried force stopping and restarting the application.
- [x] I tried turning the switch on and off again.

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

ILoveOpenSourceApplications:
I think the issue is that even though "Disable bold icons" is off, the UI is showing the old thin icons. The same is happening with me too. I was hoping it would be fixed when updating to v21.02.32, but no, it's not. Maybe one of the flags which is being forced off is also forcing off the icons from appearing on the UI. For me the icon that's bold are the Navigation buttons, feed flyout menu & Shorts player, and the settings icons. The rest are still using the old thin icons.

Edit: Nevermind, it's almost like the app was waiting for me to post about this, to enable the icons. 😑

LisoUseInAIKyrios:
The player overlay bold icons are intentionally forced off until the overlay buttons can be updated to work correctly (all the underlying code is now different and some patches need changes to use it).

The player action buttons are a server side layout, and I don't think any client side flag has any influence over whatever layout YT sends for it. YT may be experimenting with a mix of old/new layouts for some users.

I've noticed if I downgrade the app some old UI designs will return for a while even if I later install the newer app and clear the app data. I think YT sees the user downgraded and applies some old layout flags or "old grouchy don't touch my app user" logic because they interpret the user wasn't happy with the current app.

ILoveOpenSourceApplications:
Could be. Well the image in this issue, I believe is a bug from YouTube themselves, where the like button for some reason, was the new bold icon even on older versions such as 20.14.

inotia00:
Starting in October 2025, Google began testing [Themed Like Animations](https://9to5google.com/2025/10/28/every-like-button-animation-youtube/).

For example, if you click the Like button on `NBA Highlights`, a basketball animation will play.

This is implemented using lottie animations, and all animations are bold icons.

There is an issue where these `Themed Like Animations` are applied even in versions where the bold icon is not used in the action button (YouTube 20.38.37 or earlier).

[...]

PR #261, opened 2026-01-23
fix(YouTube): Replace themed like animation applied from wrong version due to server-side fault

inotia00:
Closes https://github.com/MorpheApp/morphe-patches/issues/113.

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/FIxLikeButtonPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/FIxLikeButtonPatch.java
@@ -10,9 +10,8 @@ public class FIxLikeButtonPatch {
 
     private static final String THEMED_LIKE_ANIMATIONS_PREFIX =
             "https://www.gstatic.com/youtube/img/lottie/custom_animated_like_icon/";
-    // FIXME: Before merging to the dev branch, change this value to 'https://cdn.jsdelivr.net/gh/MorpheApp/morphe-patches@dev/patches/src/main/resources/actionbar/assets/'
     private static final String THEMED_LIKE_ANIMATIONS_LEGACY_ICON_RESIZED_URL_PREFIX =
-            "https://cdn.jsdelivr.net/gh/MorpheApp/morphe-patches@refs/heads/fix/like-button/patches/src/main/resources/actionbar/assets/";
+            "https://cdn.jsdelivr.net/gh/MorpheApp/morphe-patches@dev/patches/src/main/resources/actionbar/assets/";
     private static final String THEMED_LIKE_ANIMATIONS_LEGACY_DARK_ICON_RESIZED_URL =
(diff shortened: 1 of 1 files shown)


patches/src/main/resources/actionbar/assets/animated_like_icon_dark_v4.json
inotia00, line 1: ## Additional context

This PR temporarily redirects to the old-style lottie animation until the server fixes the issue.

The URLs for the original lottie animations (old-style lottie animations) are:

```
http://www.gstatic.com/youtube/img/lottie/animated_like_icon/animated_like_icon_dark_v4.json
http://www.gstatic.com/youtube/img/lottie/animated_like_icon/animated_like_icon_light_v4.json
```

These animations are used when the like button in a comment is clicked, so they're too small for the action bar.

So, I changed the lottie animations to fit the size of the action bar.

This file is a resized lottie animation.

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/FIxLikeButtonPatch.java
```diff
@@ -0,0 +1,52 @@
...
+import app.morphe.extension.shared.Utils;
+import app.morphe.extension.youtube.patches.spoof.SpoofAppVersionPatch;
+import app.morphe.extension.youtube.shared.PlayerType;
+
+@SuppressWarnings("unused")
+public class FIxLikeButtonPatch {
+
+    private static final String THEMED_LIKE_ANIMATIONS_PREFIX =
+            "https://www.gstatic.com/youtube/img/lottie/custom_animated_like_icon/";
+    // FIXME: Before merging to the dev branch, change this value to 'https://cdn.jsdelivr.net/gh/MorpheApp/morphe-patches@dev/patches/src/main/resources/actionbar/assets/'
```
inotia00: TODO: Remove this comment before merging this PR.

inotia00:
This is a bit off-topic, but YouTube 20.39+ uses bold icons in the action bar.

Even if the `Disable bold icons` setting is enabled, bold icons in the action bar are not disabled (it depends on the server).

For more flexible code management, it's better to remove the `Disable bold icons` setting and move to the `Spoof app version`.

LisoUseInAIKyrios:
> This is a bit off-topic, but YouTube 20.39+ uses bold icons in the action bar.
> 
> Even if the `Disable bold icons` setting is enabled, bold icons in the action bar are not disabled (it depends on the server).
> 
> For more flexible code management, it's better to remove the `Disable bold icons` setting and move to the `Spoof app version`.

In earlier versions (such as 20.31.42), the bold icons are rarely enabled by YT so an override was needed to force on.

But, forcing bold on before YT decides to enable it can cause minor UI glitches in the settings menu (maybe elsewhere too) where some icons are bold and some are missing. This mostly happens on a clean install if the settings menu is immediately opened, and I assume it's because the bold icons have not yet been downloaded.

If YT is now showing bold icons for the recommended 20.37.48 or 20.40.45, then the bold override could be removed in favor of a spoof app version as that'll fix all these issues.

PR #276, opened 2026-01-24
feat(Reddit): Add support for Reddit `2026.04.0`

wchill:
Added Reddit patches from https://github.com/wchill/rvx-morphed with appropriate changes to get them to build in the official Morphe patch bundle.

### DIFF
--- a/extensions/all/spoof-wifi/build.gradle.kts
+++ b/extensions/all/spoof-wifi/build.gradle.kts
@@ -1,7 +1,7 @@
 import java.lang.Boolean.TRUE
 
 extension {
-    name = "extensions/all/connectivity/wifi/spoof/spoof-wifi.rve"
+    name = "extensions/all/connectivity/wifi/spoof/spoof-wifi.mpe"
 }
 
 android {


patches/src/main/kotlin/app/morphe/patches/all/misc/signature/BaseSpoofSignaturePatch.kt
```diff
@@ -0,0 +1,48 @@
...
+            buildMap {
+                if (!packageName.isNullOrEmpty()) put(PACKAGE_NAME, "\"$packageName\"")
+                if (!certificateData.isNullOrEmpty()) put(CERTIFICATE_BASE64, "\"$certificateData\"")
+            }
+        }
+    }
+
+    val logger = Logger.getLogger(this::class.java.name)
+
+    extendWith("extensions/spoof-signature.mpe")
```
LisoUseInAIKyrios, line 26: Is there a `spoof-signature` extension?

I'm facing a different issue where YouTube patched with this branch fails to launch with an odd Kotlin class not found:

```
 java.lang.NoSuchMethodError: No static method checkNotNull(Ljava/lang/Object;)V in class Lkotlin/jvm/internal/Intrinsics; or its super classes (declaration of 'kotlin.jvm.internal.Intrinsics' ap
 	at app.morphe.extension.youtube.swipecontrols.views.SwipeControlsOverlayLayout.getDrawable(SwipeControlsOverlayLayout.kt:61)
```

I thought it may be the additional libraries Reddit is pulling in is breaking the library resolution, but if I comment out those 2 extra libraries the error still persists.

LisoUseInAIKyrios: The YT launch error is fixed by adding a proguard keep config.

Why proguard is stripping it is strange.

wchill: Yeah, forgot to add this. Will move it into shared.

wchill: RVX needed some custom proguard rules which might be relevant.

wchill: Ah, nvm, you already took care of these things

wchill:
FYI, I've had issues logging into the official reddit app on emulator, so if you're going to do testing you'll need to use a real device. (This will also impact the automated testing plans I had; will need to see what checks they are doing)

LisoUseInAIKyrios:
Everything seems to be working.

wchill:
We can add support for 2026.04.0 as well in the compatible versions list, seems there are minimal changes there and users are reporting successful patches with the RVX Morphed patch bundle.

cyberboh:
> We can add support for 2026.04.0 as well in the compatible versions list, seems there are minimal changes there and users are reporting successful patches with the RVX Morphed patch bundle.

Why not delete the Reddit version, it will support latest app target. Users always patch latest version, if any issues occurred, they can report bug to this morphe patches repo.

wchill:
We also need to set a lower bound on the app version so people don't get confused when their old app versions are no longer supported. It's a way to reduce support burden.

wchill:
fyi @LisoUseInAIKyrios the GH Actions artifact is not publicly accessible. Only organization members can download it.

cyberboh:
> We also need to set a lower bound on the app version so people don't get confused when their old app versions are no longer supported. It's a way to reduce support burden.

If that the issue, better add only lower and higher version, not all version. That will be simplicity the version, imho.
But, up to you 🙏

wchill:
This might be a good time to enable the navigation bar patch if we can get it working.

If I use the APKM as is or merge it with APKEditor, the navigation bar patch crashes the app at startup because of an invalid resource ID.

However, if I decompile the APK with APKEditor, rebuild it with APKEditor again with no other changes, then patch the rebuilt APK, it starts working again.

Not sure what's going on, but the log line showing that the context was set in Morphe utils seems to be emitted after the app has already crashed, so it might be an initialization thing. The logic that inotia used to get the resource identifiers is also not quite the same, which might be important: https://github.com/inotia00/revanced-integrations/blob/revanced-extended/app/src/main/java/app/revanced/integrations/reddit/patches/NavigationButtonsPatch.java

wchill:
> There are too many unnecessary classes, methods, and changes for compatibility with older code.
> 
> Unused classes and methods should be removed.
> 
> The patch syntax is based on ReVanced Patcher 21. It's better to rewrite all patches from scratch using the latest Morphe Patcher syntax.
> 
>     1. `app.morphe.extension.shared.ResourceUtils` is not needed, use `app.morphe.extension.shared.Utils` instead.

I think separating resource related utility functions out is fine. Having everything in a Utils file instead reduces maintainability and readability. Is there a reason why you want them all to be in one file?

>     2. `app.morphe.extension.shared.StringRef` also doesn't need to be modified.

Don't know the context for these changes.

>     3. All resource-related patches should be removed (the latest Reddit will generate a compile error when attempting to compile resources).

The resource related patches being kept around (in commented form) only modify AndroidManifest.xml, so this is not a problem given we have plans to support manifest-only patches.

>     4. Avoid using Constants where possible. YouTube and YouTube Music don't use Constants.

I can do this

>     5. There's no reason the `Spoof signature` patch should be implemented as a shared patch.

On the contrary, there are a few other apps where `Spoof signature` would be useful. (One that comes to mind is RedReader, which uses the certificate data to generate an AES256 decryption key to decrypt the client ID used for logging into reddit.) As long as the Java code and native libs are not in the regular shared extension project, there should not be any real cost to keeping the patch itself in shared.

>     6. `PatchList` should also be removed.

I can do this

>     7. The settings patch contains a lot of unnecessary code for compatibility with older Reddit versions. It needs to be cleaned up.

Support for everything before 2025.40.0 has been dropped already and the relevant code removed, what else is there to clean up?

>     8. `BytecodeUtils` also doesn't need to be modified.

Maybe `indexOfFirstStringInstruction` can be replaced with filters, but I don't agree with not including `methodSignature()` (or something similar) or modifying `findMutableMethodOf()`.

* `methodSignature` is useful in Patcheddit patches as well (currently I have to manually construct the signatures in a few cases), so it might as well be kept around.
* `findMutableMethodOf` is ok taking in `MethodReference` instead of `Method`: `Method` extends `MethodReference`, there is nothing specific to `Method` that is being used, and it would be semantically incorrect to use a `Method` instead of `MethodReference` in the one patch where it's relevant.

LisoUseInAIKyrios:
Resource code was cleaned up by merging the ResourcesUtils and the existing resource related Utils together. I think having a separate resource utility class is better just for organization, and there is already a resource utils on the kotlin patches side of the code so the organization is consistent.

StringRef got an overdue cleanup because it doesn't need to manage looking up resources when other code now does this (the class is very old and hasn't been touched in years).

Re-writing fingerprints with Morphe filter syntax will be quick and easy, but we could do some/most of that now then incrementally finish the last fingerprints as patches are changed/updated.

LisoUseInAIKyrios:
For item `4. Avoid using Constants where possible. YouTube and YouTube Music don't use Constants.`

The existing Reddit shard compatibility tag can stay. Patcher will soon have some changes to the compatibility tag and YouTube and YT Music would benefit from also having a shared compatibility tag like Reddit.

LisoUseInAIKyrios:
Only item left is 7?

> 7. The settings patch contains a lot of unnecessary code for compatibility with older Reddit versions. It needs to be cleaned up.

For item 8, the two methods added to BytecodeUtils are needed by some old legacy fingerprint usage that I didn't want to touch because it's fairly complex and changing the code to filters can wait until the next time the patch is modified.

Item 3 (remove resource patches), can be done later because we may figure out a resource fix and it's easier to keep the  patches still compiling but disabled. Can delete them later if we give up on solving resource patching.

extensions/shared/library/src/main/java/app/morphe/extension/shared/Utils.java
```diff
@@ -1197,4 +1170,12 @@ protected boolean removeEldestEntry(Entry eldest) {
...
         };
     }
+
+    /**
+     * @return whether the device's API level is higher than a specific SDK version.
+     */
+    @ChecksSdkIntAtLeast(parameter = 0)
+    public static boolean isSDKAbove(int sdk) {
+        return Build.VERSION.SDK_INT >= sdk;
+    }
```
kitadai31, line 1180: This `isSDKAbove` method is unnecessary.
It can be replaced with normal `Build.VERSION.SDK_INT` comparison.
 
This was used in RVX in order to support Android 5-6 for YT Music while keeping minSdk 24 (Android 7).
Because if the minSdk is set to 24, the compiler removes the normal `if (Build.VERSION.SDK_INT >= 24)` statements.

extensions/reddit/build.gradle.kts
```diff
@@ -0,0 +1,7 @@
+dependencies {
+    compileOnly(project(":extensions:shared:library"))
+    compileOnly(project(":extensions:reddit:stub"))
+    compileOnly(libs.annotation)
+    implementation(libs.hiddenapi)
+    coreLibraryDesugaring(libs.desugar.jdk.libs)
```
kitadai31: coreLibraryDesugaring is not needed
Because the recent versions of Reddit is Android 9+

By the way, I know it is from RVX, but I think it was actually not working from the beginning.
Because, to use core library desugaring, `compileOptions.isCoreLibraryDesugaringEnabled = true` is needed in build.gradle, but RVX didn't have it.
And to use it in ReVanced Extensions in particular, it is needed to bundle another dex file from build intermediate directory into rvp file [like this](https://github.com/kitadai31/revanced-patches-android6-7/commit/3f32f6bc09bc81b67cf82cf1ecc0999e88788178#diff-ce28239b6580555cc53f4c64582939f753c84d0bb42e44dd1a601a6fa6829a0dR22-R26).

In addition, we can set minSdk to 28 in this file

wchill: I swear I've used the desugar libs in my old wchill/revanced-patches repo with no issues.

extensions/shared/library/src/main/java/app/morphe/extension/shared/Utils.java
LisoUseInAIKyrios: Can leave this for now, because there's an open request to add support for YT Music for Android 5-6/7.

Can remove later if needed.

wchill:
Fixed the navigation bar patch by just passing the resource IDs directly from the smali instructions and then maintaining our own map of resource ID -> label name.

Also it supports disabling the Answer button now.

Also removed a bunch of obsolete code from the navigation bar patches that was not being used anymore.

LisoUseInAIKyrios:
Seems like everything is done?

The last of the old fingerprints that don't use instruction filters can be refactored the next time the patch is touched. It's not worth the time or energy right now when the patches works and they currently don't need any functional patch changes.

wchill:
I think we can merge this... All the patches look like they're working after the port, and any other desired changes can be in a follow up

extensions/shared/library/src/main/java/app/morphe/extension/shared/Utils.java
kitadai31: Even if we support Android 5 in the future, we should not use `isSDKAbove` regardless.

Supporting Android 5-6 while keeping minSdk at 24 was not a ideal solution.
In fact, due to this, RVX overlooked the fact that it was using APIs that were not supported by Android 5.0 for more than half a year.

However, since RVX used a single shared extension for all apps, setting a low minSdk would cause inconvenience.
c.f. https://github.com/inotia00/revanced-integrations/pull/57#discussion_r1756252743

Because of this, RVX had to rely on the `isSdkAbove()` hack.

On the other hand, Morphe has independent extensions for each app, so you can set minSdk to 21 only for YT Music codes.

Therefore, the `isSdkAbove()` hack won't be necessary for Morphe.

extensions/reddit/build.gradle.kts
kitadai31: That's obvious
Because It isn't enabled (`isCoreLibraryDesugaringEnabled = true` is missing)
So the `coreLibraryDesugaring(libs.desugar.jdk.libs)` had had no effects

Let's just remove it so that it don't cause confusions in the future

(Of course, this can be done in follow-up commits if you prioritize merging this PR quickly)

extensions/shared/library/src/main/java/app/morphe/extension/shared/Utils.java
LisoUseInAIKyrios: The shared extension library uses Android 8.0 minimum, and YT Music pulls in shared extension. So this hack may still be needed for there.

extensions/reddit/build.gradle.kts
LisoUseInAIKyrios: library was removed, and min sdk was changed to Android 9.0 (min version of oldest supported Reddit)

kitadai31: @LisoUseInAIKyrios I saw 588c148
But that build.gradle.kts is for stub module, and [extensions/reddit/build.gradle.kts](https://github.com/MorpheApp/morphe-patches/pull/276/files#diff-0547ed7ef361723c67e0db4cc7c6f619b072d8b372b191c28a5a59952817416f) still doesn't have minSdk
I recommend to add minSdk to it as well

p.s. Thank you for very quick response!

extensions/reddit/src/main/java/app/morphe/extension/reddit/patches/SpoofSignaturePatchReddit.java
kitadai31, line 1: This seems to be unused code.
Instead, `extensions/spoof-signature/src/main/java/app/morphe/extension/all/misc/signature/SpoofSignaturePatch.java` is referenced by patches instead.

Also, I noticed that the credit comment (`// https://github.com/L-JINBIN/ApkSignatureKillerEx/...`) is missing in the correct `SpoofSignaturePatch.java`.
I think it is better to add it
(This was already the case in the original RVX code.)

extensions/all/spoof-wifi/build.gradle.kts
```diff
@@ -0,0 +1,29 @@
+import java.lang.Boolean.TRUE
+
+extension {
+    name = "extensions/all/connectivity/wifi/spoof/spoof-wifi.rve"
```
wchill: Wrong file extension

wchill:
Going to squash merge this now since all the big stuff and low hanging fruit has been addressed, feel free to assign to me any issues that come up

Issue #284, opened 2026-01-25
feat(YouTube): Add `Restore old search filters` patch


inotia00:
### Feature description

[YouTube recently removed several search filters, including `Sort by upload date` and `Sort by rating`](https://www.reddit.com/r/youtube/comments/1qjgsln/youtube_removed_the_ability_to_sort_by_upload/).

Add a patch to revert these search filters.

### Additional context

I'm listing several methods I've tried over the past three days to restore these search filter options:

> [!Note]
> #### ​1. Spoofing clients
> 
> : [When YouTube removed the `Sort by oldest` button from channels](https://www.reddit.com/r/youtube/comments/yeu51a/why_on_earth_would_youtube_remove_the_sort_by/), spoofing the app version to 17.06.35 worked for several months.
> 
> So I tried several things like changing the app version, changing the form factor, changing the client id, etc. but no success.
> 
> Google has removed search filters from all platforms, including YouTube Web, YouTube TV, and YouTube VR, as well as the YouTube app.
> 
> #### ​2. Implementing search filter dialog
> 
> <img width="611" height="974" alt="Image" src="https://github.com/user-attachments/assets/be2cdcf7-2c3f-4703-83de-127ced592d1e" />
> 
> : Fortunately, the search filter dialog is an alert dialog, not a `litho view`, so theoretically, it could be replaced with a custom dialog (Implements all behavior of the dialog, including the dialog layout and onClickListener).
> 
> I'm not a UI designer, so I couldn't choose this approach, even though it was technically the simplest workaround.
> 
> #### ​3. Parsing and modifying the `/search` endpoint response
> 
> Like Chrome Extensions (Manifest V2), parsing and modifying the endpoint response is also a valid approach.
> 
> Since all requests and responses are processed as protos in the Android YouTube app, this is the most difficult implementation.
> 
> Nevertheless, I chose this approach because I absolutely hate UI design.
> 
> Since the `MessageLite` interface is an immutable object, I had to map all proto fields to use the proto builder.
> 
> I wrote the proto file based on the [YouTube.js](https://github.com/LuanRT/YouTube.js) code and referenced the [SmartTube](https://github.com/yuliskov/SmartTube) code to find the constant values ​​for the search filters `Sort by upload date` and `Sort by rating`.
> 
> After many attempts, I successfully implemented the patch without using the private YouTube API.

### Motivation

`Sort by upload date` and `Sort by rating` have been useful filter options for me, so I absolutely needed them.

[...]

PR #285, opened 2026-01-25
feat(YouTube): Add `Restore old search filters` patch

inotia00:
Closes https://github.com/MorpheApp/morphe-patches/issues/284


LisoUseInAIKyrios:
Patching protocol buffers bring some complexity and future maintenance, but I think it's well worth it because restoring search filters is a highly desired feature.

adrianis86:
Why starting with 1.13.0 the patch is not available anymore?

LandonMoran:
Because it didn't work anymore, after YouTube server-side changes

Issue #43, opened 2026-01-02
feat(YouTube): Adapt patches from RVX


Momud06:
### Feature description

Adapt more patches from RVX into YouTube Morphe.
Things that are missing that I found:
Ability to hide youtube settings.
Ability to hide titles of videos in fullscreen.
Ability to hide buttons in the You section.
Ability to hide the channel bar in the subscriptions section.
Ability to hide the category selection on the main and subsciptions sections.
Ability to hide the microphone button when searching.
Ability to hide channel sections.
Ability to hide Thanks and Report action buttons.
Ability to hide the collapse button in the player.
Ability to exclude channels from sponsorblock.
Ability to remove the start up animation.
Ability to hide buttons that appear in 3 dot menus next to videos, posts, playlists.

### Motivation

Because YouTube RVX has always been the best, but it's no longer continued in that form.

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] This is a feature request for YouTube or YT Music

lucai9000:
Yeah, would be great. Also i really would like to see yt 19.16.39 patch support just like in rvx until it's possible. This version is the last one that enables you scrubbing in timeline while audio still plays.

johnnyw94:
Please add toolbar options in your list! 

<details>

![Screenshot_2026-01-02-19-52-52-93_0635f5993b1f7a268b0ea09a0649cdab.jpg](https://github.com/user-attachments/assets/173fb288-7ec2-43cb-9946-cd6dbe2ffa11)

</details>

micromurad:
also the custom branding icons

Tonwalter888:
Also remove these two buttons in toolbar.

<details>

![Screenshot_2026-01-03-15-16-54-60_b50abca68101f9ac06900ab663f0038c.jpg](https://github.com/user-attachments/assets/d36644e1-a765-403b-9eab-a25db8b042bf)

![Screenshot_2026-01-03-15-16-12-36_b50abca68101f9ac06900ab663f0038c.jpg](https://github.com/user-attachments/assets/bb7854d9-52d7-44ab-a9a5-8105e0bbf04d)

</details>

btncua:
In short, all developments in RVX over so many years should be transferred and adapted.

johnnyw94:
- Invert snack bar theme

yazaovon:
Add some experimental from RVX
[...]

PR #298, opened 2026-01-26
feat(YouTube - Navigation bar): Add settings to hide toolbar buttons

LisoUseInAIKyrios:
Code adapted from @inotia00 

https://github.com/inotia00/revanced-patches/blob/dev/patches/src/main/kotlin/app/revanced/patches/youtube/utils/toolbar/ToolBarHookPatch.kt

&nbsp;
This PR also renames the patch `Navigation buttons` to `Navigation bar`


wchill:
Can we also track this in #43 so we know what RVX patches still need to be ported

ANewDawn:
Does this patch hide the navigation bar also?

LisoUseInAIKyrios:
No. It hides toolbar items (upper nav bar)

ILoveOpenSourceApplications:
Why was `Hide voice search button` excluded when adapted? Was it not working?

LisoUseInAIKyrios:
I didn't try adding it.

Code is here:

https://github.com/inotia00/revanced-patches/blob/3c6a371e32f2f15f8de6c9ed66126df4fbab7efb/patches/src/main/kotlin/app/revanced/patches/youtube/general/toolbar/ToolBarComponentsPatch.kt#L376-L423

PR #305, opened 2026-01-27
feat(YouTube - Hide layout components): Add "Hide Latest videos button" setting

ILoveOpenSourceApplications:
Code adapted from @inotia00's [RVX FeedComponentsPatch.kt](https://github.com/inotia00/revanced-patches/blob/3c6a371e32f2f15f8de6c9ed66126df4fbab7efb/patches/src/main/kotlin/app/revanced/patches/youtube/feed/components/FeedComponentsPatch.kt#L91-L109).

Need to know a replicable method for the button to show up for proper testing and see if it works. (Let me know if you know the way.)

hide-latest-videos

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/LayoutComponentsFilter.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/LayoutComponentsFilter.java
@@ -475,7 +475,6 @@ public final class LayoutComponentsFilter extends Filter {
         Utils.hideViewUnderCondition(Settings.HIDE_LATEST_VIDEOS_BUTTON.get(), view);
     }
 
-
     /**
      * Injection point.
      */


LisoUseInAIKyrios:
Try subscribing to a ton of channels, then leave the app open in the subscription tab and it shows up when new videos are added but before the feed is refreshed?  Might need to go into a video then back out.

ILoveOpenSourceApplications:
The problem is when `Latest videos` button appears and I switch to the You tab and return back to Home tab, the feed automatically refreshes and the button goes away. So I'll leave a video on for sometime and check if its popped up or not.

LisoUseInAIKyrios:
Add a temporary toast in the filter code that returns true to hide the refresh component.:

```
    public static void hideLatestVideosButton(View view) {
        Utils.showToastLong("Blocked latest videos");
        Utils.hideViewUnderCondition(Settings.HIDE_LATEST_VIDEOS_BUTTON, view);
    }
```

If  you can get the toast to show and there is no refresh button on screen, then the patch works.

patches/src/main/kotlin/app/morphe/patches/youtube/layout/hide/general/Fingerprints.kt
```diff
@@ -176,3 +176,30 @@ internal object SearchBoxTypingStringFingerprint : Fingerprint(
...
+internal object LatestVideosContentPillFingerprint : Fingerprint(
+    accessFlags = listOf(AccessFlags.PUBLIC, AccessFlags.FINAL),
+    returnType = "V",
+    parameters = listOf("L", "Z"),
+    filters = listOf(
+        resourceLiteral(ResourceType.LAYOUT, "content_pill"),
+        methodCall(
+            smali = "Landroid/view/LayoutInflater;->inflate(ILandroid/view/ViewGroup;Z)Landroid/view/View;"
+        ),
+        opcode(Opcode.MOVE_RESULT_OBJECT, location = MatchAfterImmediately())
```
LisoUseInAIKyrios, line 188: Did you write this? The original code didn't use filters.

Looks great.

ILoveOpenSourceApplications: Nah, I'm not that good. Had the help of AI to port it.

ILoveOpenSourceApplications:
I think it's not working. I had the setting on and still saw the button. I wonder what went wrong.

ILoveOpenSourceApplications:
Tested and works.

LisoUseInAIKyrios:
What changed from when it may not have been working?

ILoveOpenSourceApplications:
Commit https://github.com/MorpheApp/morphe-patches/pull/305/commits/f6335f72bae282d82d38963a4ca96de686c9d5a0 must've done it.

LisoUseInAIKyrios:
That commit doesn't actually change anything. Maybe something else went wrong with previous verification.

ILoveOpenSourceApplications:
Maybe, I left a 2 hour video to run and minimized it and saw no `Latest videos` button with the setting on. So, yay me I guess!.

LisoUseInAIKyrios:
Did you try the toast code change?

But if it seems to work then we can merge now.

ILoveOpenSourceApplications:
I did, but patiently waiting for the button to show up without for sure knowing whether it did or not is quite tiresome and I choose to believe that it's working as intended.

PR #335, opened 2026-01-30
feat(YouTube): Add experimental support for `21.05.264`

LisoUseInAIKyrios:
**Edit**: If experimenting, then use pre-release patches. Two additional fixes were needed for 21.05.

Also adds experimental YT Music `9.04.50`

To patch this, use Morphe home page patching and provide YT [21.05.264 (universal nodpi)](https://www.google.com/search?q=youtube+21.05.264+nodpi), and confirm the Morphe warning that the version is not the recommended version.

If you experience bugs or issues that only happen on the latest YT release, but are confirmed not happen with the recommended version (`20.40.45` at the time of this PR), then please post your findings in https://github.com/MorpheApp/morphe-patches/issues/36


LisoUseInAIKyrios:
`20.40.45` has enough difference in the debugging hook that's is a hassle to support with the latest experimental. 

May drop `20.40.45` and change to `20.41.41`

Edit: Can ignore the double/long hooks for 20.40 because they're not that useful.

Issue #266, opened 2026-01-23
feat(YouTube / YT Music): Spoof video streams TV/Web clients


tanvir8282:
### Feature description
In rvx inotia00 added TV related clients it was under 'Javascripts' section which seems to be stable Than Android VR and others...as some of them has Audio track menu missing and other things such as AV1 not works....

Please consider adding these clients too.....

### Motivation

The clients adding feature will stop Buffer related issues as many of users facing issues with Android VR and In SDK client too.

Thanks

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] This is a feature request for YouTube or YT Music

LisoUseInAIKyrios:
Please make a separate issue for Return YouTube username

Edit: a prior issue already existed

tanvir8282:
> Please make a separate issue for Return YouTube username

Okay..

LisoUseInAIKyrios:
For now, if anyone has issues with VR, then try VR sign-in because it fixes all known user/region issues with VR.

No SDK had issues for certain users but now seems to be working again. If No SDK works then it's by far the best client and is better than all web/tv clients.

Update: No SDK is now failing for most users.

tanvir8282:
> For now, if anyone has issues with VR, then try VR sign-in because it fixes all known user/region issues with VR.
> 
VR caused some issues for me ....not all the time but yeah i faced it


> No SDK had issues for certain users but now seems to be working again. If No SDK works then it's by far the best client and is better than all web/tv clients.

I am having buffer issues with Sdk clients with 1.7.0 patch today with Yt 21.02.35.....not with all videos i clicked but some of it....


Now using beta youtube 21.04.222 with 1.8.0-dev 2......with same client no issue till now......no videos buffer issue

BlackGold8282:
It would be nice to have a client (excluding NoSDK) that can use audio track menu.

BlackGold8282:
@LisoUseInAIKyrios

Currently, ‘LiveStream for Kids’ is not playing in YT Morphe.

It plays on the ‘Unknown’ client but experiences playback issues after 1 minute.

https://www.youtube.com/live/jPX7bz6qz6g

LisoUseInAIKyrios:
What if Android VR sign-in is used?

BlackGold8282:
'VR Auth' can't play Kids videos, so even if I log in, it plays on the 'Unknown' client.

Unplayable Video
• VR Auth: [Kids]
[...]

PR #340, opened 2026-01-31

inotia00:
### Remove `Android No SDK` and `Android Music No SDK`

- `Android No SDK` has been used by many open-source YouTube extractors, including yt-dlp.

- PoToken is now required to use `Android No SDK`.

- So both `Android No SDK` and `Android Music No SDK` have been removed.

### Add `TV` and `TV Simply`

- `TV` and `TV Simply` have been added as workarounds for the lack of playability.

- Both `TV` and `TV Simply` need to solve the JavaScript challenge for deobfuscation, so J2V8 is used as the JS runtime engine (Official YouTube TV app also uses J2V8).

- Both `TV` and `TV Simply` have the side effect that the video may start 2-5 seconds late, but they are good alternatives to `Android No SDK`.

- `TV` is the only client that supports all videos, even in incognito mode.

- `TV Simply` requires a PoToken when signed out, so it does not work in incognito mode.

### Credits

- [TeamNewPipe/NewPipeExtractor](https://github.com/TeamNewPipe/NewPipeExtractor) (Referenced for writing deobfuscation classes and methods)

- [yuliskov/MediaServiceCore](https://github.com/yuliskov/MediaServiceCore) (Referenced for writing classes for the yt-dlp-ejs wrapper for Java)

- [yt-dlp/ejs](https://github.com/yt-dlp/ejs) (Source of scripts for solving the latest player JavaScript challenge)

- [LuanRT/YouTube.js](https://github.com/LuanRT/YouTube.js) (Referenced for writing .proto files)

### DIFF
--- a/extensions/shared/library/src/main/java/app/morphe/extension/shared/Utils.java
+++ b/extensions/shared/library/src/main/java/app/morphe/extension/shared/Utils.java
@@ -417,7 +417,7 @@ public class Utils {
         clipboard.setPrimaryClip(clip);
     }
 
-    public static boolean isNotEmpty(final String str) {
+    public static boolean isNotEmpty(@Nullable String str) {
         return str != null && !str.isEmpty();
     }
 
--- a/extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/SharedPrefCategory.java
+++ b/extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/SharedPrefCategory.java
@@ -1,5 +1,6 @@
 package app.morphe.extension.shared.settings.preference;
 
+import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.preference.PreferenceFragment;
@@ -37,21 +38,25 @@ public class SharedPrefCategory {
         removeKey(key);
     }
 
+    @SuppressLint("ApplySharedPref") // Must use commit to ensure default value is not saved to preferences.
     private void saveObjectAsString(@NonNull String key, @Nullable Object value) {
         preferences.edit().putString(key, (value == null ? null : value.toString())).commit();
     }
 
+    @SuppressLint("ApplySharedPref") // Must use commit to ensure default value is not saved to preferences.
     public void clear() {
-        preferences.edit().clear().apply();
+        preferences.edit().clear().commit();
     }
 
     /**
      * Removes any preference data type that has the specified key.
      */
+    @SuppressLint("ApplySharedPref") // Must use commit to ensure default value is not saved to preferences.
     public void removeKey(@NonNull String key) {
         preferences.edit().remove(Objects.requireNonNull(key)).commit();
     }
 
+    @SuppressLint("ApplySharedPref") // Must use commit to ensure default value is not saved to preferences.
     public void saveBoolean(@NonNull String key, boolean value) {
         preferences.edit().putBoolean(key, value).commit();
     }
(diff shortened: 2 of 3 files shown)


settings.gradle.kts
```diff
@@ -20,7 +20,7 @@ pluginManagement {
 }
 
 plugins {
-    id("app.morphe.patches") version "1.1.0"
+    id("app.morphe.patches") version "1.1.1"
```
inotia00, line 23: This PR depends on https://github.com/MorpheApp/morphe-patches-gradle-plugin/pull/9 and may be merged after gradle plugin 1.1.1 is released.

inotia00:
Closes https://github.com/MorpheApp/morphe-patches/issues/266.
(PoToken generator is currently broken, so this PR is sufficient to close the issue)

wchill:
Shouldn't the extension code be in a separate extension package? If J2V8 and the other dependencies are added to shared, won't reddit and other apps also end up having essentially dead code bloating up the APK size

LisoUseInAIKyrios:
It could.

All of YT and YT Music could be extracted into their own shared extension. So YouTube would be 3 extensions:
- YouTube
- YouTube & YT Music Shared
- Shared

That could be a separate refactor. Not sure how it will work with the gradle java plugin.

LandonMoran:
Will this affect like this stuff? Because I used to use Android no SDK because it had like stable volume and stuff. But now I'm using Android VR but it has no stable volume

LisoUseInAIKyrios:
Tv simply has stable volume.

0xrxL:
The only need for Android_No_SDK is for live videos marked for kids, and for these seems to be working still. But I'm afraid that uDrop will need TV client sooner or later.

@inotia00 @LisoUseInAIKyrios can't you simplify the code, to be more easier ported?

inotia00:
`TV` can play live videos marked for kids (even in incognito mode).

The PR has already been written with minimal code changes.

This code cannot be further minimized than it is now.

(Writing all code in Kotlin would reduce the code size further, but in this case, Morphe would throw a runtime exception)

error-reporting:
> Will this affect like this stuff? Because I used to use Android no SDK because it had like stable volume and stuff. But now I'm using Android VR but it has no stable volume

visionos has stable volume too

LisoUseInAIKyrios:
This is fantastic and a big improvement since No SDK is now gone.

NekoZ410:
I've been using Android No SDK as default client because there's no other client that have Audio track menu. It still have an minor issue when play time approaching 0:55 - 1:00 and player will have to pause a bit then continue with locked video quality, but it's still good enough to use.
Does new replacement clients have Audio track menu?

LisoUseInAIKyrios:
TV Simply has an audio track menu

fromhgbwithluv:
@inotia00 The n-sig extractor from Metropolis is much simpler to implement and maintain, in my opinion. You can check this PR https://github.com/MetrolistGroup/Metrolist/pull/2764

Moreover, Android also included v8 engine in-built

inotia00:
> @inotia00 The n-sig extractor from Metropolis is much simpler to implement and maintain, in my opinion. You can check this PR [MetrolistGroup/Metrolist#2764](https://github.com/MetrolistGroup/Metrolist/pull/2764)
> 
> Moreover, Android also included v8 engine in-built

I've already checked.

The latest Metrolist appears to use `WEB_REMIX` client, but when I checked the network packets, it still uses the `ANDROID_REEL` client to play music (Same as NewPipe and SmartTube).

So I can't verify that the PoToken generator they refactored is valid.

Metrolist obtains a PoToken using WebView and the DroidGuard VM, fetches streaming data using the `WEB_REMIX` client, and ultimately uses `ANDROID_REEL`.

There is a 3-second delay at the start of each song due to the WebView and DroidGuard VM startup, but this is unnecessary.

The `ANDROID_REEL` client does not require a PoToken.

(Also the n-sig extractor is not used in `ANDROID_REEL` client)

fromhgbwithluv:
> So I can't verify that the PoToken generator they refactored is valid.

I see. As I checked, their PoToken with WebView is the same with NewPipe ones, and also have appeared in [a PR of mine](https://github.com/inotia00/revanced-patches/pull/141) to rvx. Therefore it should work correctly, just give a try xD

> There is a 3-second delay at the start of each song due to the WebView and DroidGuard VM startup, but this is unnecessary.

The WebView is initialized only once at startup, isn't it ? If so, that is not a really big problem

PR #341, opened 2026-01-31
feat(YouTube - Captions): Add "Set caption cookies" and "Fix transcript" settings

inotia00:
### Added `Set caption cookies` setting

- This setting is derived from the [`Set Transcript Cookies` setting in RVX](https://github.com/inotia00/ReVanced_Extended/issues/3147).

- Since the auto-generate captions issue still hasn't been fixed even on the latest unpatched YouTube, this setting has been added to Morphe as well.

- You can also obtain valid cookies using `Get caption cookies`, but cookies obtained this way take some time to become usable.

- You can also export cookies from Chrome by following [this document](https://gist.github.com/inotia00/2dc2d83b3d3b0dc91ec6ddb963a6099f).

### Added `Fix transcript` setting

- Google has also started requiring PoTokens for caption requests (Timed Text API, Transcript API).

- MicroG cannot resolve the PoToken challenge for captions, resulting in a `Precondition check failure` error.

- Currently, the only workaround is to spoof the app version to `20.05.46`.

- `Fix transcript` setting only spoofs the app version for the `/get_transcript` endpoint, preventing the old UI from being used (YouTube 20.47.62 has a new UI for Transcripts, which is still used even if the `Fix transcript` setting is enabled).

- `20.05.46` is no longer needed in `Spoof app version target`, so it has been removed.

### `Auto captions` patch has been merged into the `Captions` patch

- Since captions and transcripts use the same API, they have been integrated into one patch.


patches/src/main/kotlin/app/morphe/patches/youtube/layout/captions/CaptionsPatch.kt
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Caption settings.  Used to organize all caption related settings together.
+ */
+internal val settingsMenuCaptionGroup = mutableSetOf<BasePreference>()
+
+@Suppress("unused")
+val captionsPatch = bytecodePatch(
+    name = "Captions",
+    description = "Adds an option to disable captions from being automatically enabled or to set caption cookies.",
```
inotia00, line 16: Is the description appropriate?

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -590,7 +581,41 @@ Adjust volume by swiping vertically on the right side of the screen"</string>
     <string name="morphe_hide_save_button_summary_on">Save button is hidden</string>
     <string name="morphe_hide_save_button_summary_off">Save button is shown</string>
 
-    <!-- layout.buttons.navigation.navigationBarPatch -->
+    <!-- layout.captions.captionsPatch -->
+    <string name="morphe_captions_screen_title">Captions</string>
+    <string name="morphe_captions_screen_summary">Disable captions from being automatically enabled or set caption cookies</string>
```
inotia00, line 586: Is the description appropriate?

extensions/youtube/src/main/java/app/morphe/extension/youtube/settings/preference/GetCaptionCookiesPreference.java
```diff
@@ -0,0 +1,146 @@
...
+            if (!newValue.equals(cookieString)) {
+                Logger.printDebug(() -> "new Cookies loaded: " + newValue);
+
+                // For some reason, using 'Settings.CAPTION_COOKIES.save(newValue)' resets the settings.
+                // TODO: Find the cause of the issue and replace it with 'Settings.CAPTION_COOKIES.save(newValue)'
+                AbstractPreferenceFragment.settingImportInProgress = true;
+                Setting.privateSetValueFromString(Settings.CAPTION_COOKIES, newValue);
+                Settings.CAPTION_COOKIES.saveToPreferences();
+                AbstractPreferenceFragment.settingImportInProgress = false;
+                AbstractPreferenceFragment.showRestartDialog(getContext());
```
inotia00, line 73: For some reason, when I try to save settings in Preference, it resets to default.

patches/src/main/kotlin/app/morphe/patches/youtube/layout/captions/CaptionsPatch.kt
LisoUseInAIKyrios: Looks ok to me

patches/src/main/resources/addresources/values/youtube/strings.xml
LisoUseInAIKyrios: Maybe remove "set caption cookies" because it's a little technical.

extensions/youtube/src/main/java/app/morphe/extension/youtube/settings/preference/GetCaptionCookiesPreference.java
LisoUseInAIKyrios: It's probably the preference fragment overwriting the value during a sync.

All of this old UI preference syncing could be greatly improved by making UI preference subclasses (MorpheSwitchPreference, MorpheListPreference, etc), and each manages it's own Setting including syncing to/from. That would fix the old messy approach of having a fragment do the syncing which may be interfering with what's going on here.

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -1229,9 +1254,8 @@ This will change the appearance and features of the app, but unknown side effect
 
 If later turned off, it is recommended to clear the app data to prevent UI bugs."</string>
     <string name="morphe_spoof_app_version_target_title">Spoof app version target</string>
-    <string name="morphe_spoof_app_version_target_entry_1">20.13.41 - Restore non collapsed video action bar</string>
-    <string name="morphe_spoof_app_version_target_entry_2">20.05.46 - Restore transcript functionality</string>
-    <string name="morphe_spoof_app_version_target_entry_3">19.35.36 - Restore old Shorts player icons</string>
+    <string name="morphe_spoof_app_version_target_entry_19_35">19.35.36 - Restore old Shorts player icons</string>
+    <string name="morphe_spoof_app_version_target_entry_20_13">20.13.41 - Restore non collapsed video action bar</string>
```
LisoUseInAIKyrios, line 1258: TODO: Rename these keys on Crowdin before merging

patches/src/main/resources/addresources/values/youtube/strings.xml
LisoUseInAIKyrios: I just noticed the settings has a cookies override. So this is ok.

Issue #368, opened 2026-02-01
bug(Spoof video streams): Spoofing does not work if the default client is not `TV` with pre-release patches


98xe:
### Bug description

S23, Android 16
YouTube `20.40.45`
Patches `v1.10.0-dev.1`

--

After updating to `v1.10.0-dev.1` (or later), the `Android VR` client stops working. I experience the same symptoms as https://github.com/MorpheApp/morphe-patches/issues/115.

- Signing in to `Android VR` does not fix the issue.
- `TV Simply` works as an alternative, but has the downside of the 2+ second buffer delay.
- Rolling back to `v1.9.0` fixes the issue. `Android VR` becomes usable.

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

LisoUseInAIKyrios:
A minor fix is needed.

Either don't patch with pre-release, or use pre-release with a TV client.

98xe:
Yeah, I wrote this as a draft earlier today, but since then others have already reported pretty much the same thing (like https://github.com/MorpheApp/morphe-patches/issues/360), so that's why I closed it.

ILoveOpenSourceApplications:
Could close #235 as a duplicate of this?

LisoUseInAIKyrios:
This is specific to pre-release.

Non pre-release the fix is to not use No SDK.

inotia00:
Fixed in [Morphe Patches 1.10.0-dev.5](https://github.com/MorpheApp/morphe-patches/releases/tag/v1.10.0-dev.5).

If the cache remains, the Morphe API may not immediately load the latest patch. In this case, you can patch 1.10.0-dev.5 [via the CLI](https://github.com/MorpheApp/morphe-cli/blob/main/docs/README.md).

Issue #235, opened 2026-01-18
bug(YouTube): 720p quality


Zari2015:
### Bug description

The highest quality that can reach is 720p despite some videos have 1440p or even 4k.

I do not want to brag about it or to become a weight but in the others forks do not have this problem and some phones have a screen of straight up 1440p resolution and the quality is unmatched.

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

LisoUseInAIKyrios:
Try using VR sign-in.

If that doesn't work then uninstall and try patching 20.14

D4vidGarcia:
I'm getting this same issue after a few seconds of playback (around 20-30). The resolution drops to something lower than or equal to 480p in my case, even when I have it set to 4k, 1440p, or 1080p.

<img width="2340" height="1080" alt="Image" src="https://github.com/user-attachments/assets/21936df3-e757-422b-8d3c-2510f92c3d18" />
<img width="2340" height="1080" alt="Image" src="https://github.com/user-attachments/assets/43d6ddc3-5d28-47b4-9a94-724060f64918" />

As you can see, the resolution is extremely low even though I have it set to 2160p. Also, it's not my internet's fault, as I have 1 Gbps up and down, a Wi-Fi 6 connection, and a Galaxy S24 FE. And it used to work just fine before the 1.7.0 update. Now I'm on 1.8.0, and it has the exact same problem. I patched with the recommended version (v20.40.45)

PR #373, opened 2026-02-02
fix(Spoof video streams): Spoofing does not work if the default client is not `TV`

inotia00:
Closes https://github.com/MorpheApp/morphe-patches/issues/368.

### DIFF
--- a/extensions/shared/library/src/main/java/app/morphe/extension/shared/spoof/requests/StreamingDataRequest.java
+++ b/extensions/shared/library/src/main/java/app/morphe/extension/shared/spoof/requests/StreamingDataRequest.java
@@ -262,7 +262,6 @@ public class StreamingDataRequest {
             }
 
             PlayerResponse.Builder responseBuilder = playerResponse.toBuilder();
-
             if (!playerResponse.hasStreamingData()) {
                 handleDebugToast("Debug: Ignoring empty streaming data (%s)", clientType);
                 return null;


patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -1596,7 +1596,8 @@ Video playback with AV1 may stutter or drop frames."</string>
     <string name="morphe_spoof_video_streams_about_no_av1">• No AV1 video codec</string>
     <string name="morphe_spoof_video_streams_about_no_stable_volume">• Stable volume is not available</string>
     <string name="morphe_spoof_video_streams_about_js">• Video may start late</string>
-    <string name="morphe_spoof_video_streams_about_kids_videos">• Kids videos may not play when logged out or in incognito mode</string>
+    <string name="morphe_spoof_video_streams_about_kids_videos_incognito">• Kids videos may not play when logged out or in incognito mode</string>
+    <string name="morphe_spoof_video_streams_about_kids_videos_live">• Kids live videos may not play</string>
```
inotia00: Is there a better way to show side effects?

inotia00: If `TV` is added to the list of available clients, these side effects are no longer necessary:

https://github.com/MorpheApp/morphe-patches/blob/4de5d1e7caad86e53e5a7d73469991887deed2a9/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/spoof/SpoofVideoStreamsPatch.java#L45-L49

inotia00:
https://github.com/MorpheApp/morphe-patches/issues/235 can also be closed, as `TV`, which works even when logged out or in incognito mode, has been added to the available clients.

LisoUseInAIKyrios:
Merge when ready

Issue #183, opened 2026-01-11
bug(YouTube): Shorts player has no overlay buttons when "Spoof app version" is enabled


LisoUseInAIKyrios:
### Bug description

When patching very new versions of YouTube (tested with `21.02.32`), the Shorts player can show no overlay buttons when version spoofing is enabled. It's unclear if an old client side flag needs to be forced on, or if the server provided UI layout is  missing.

### Solution

If you want version spoofing and you use the Shorts player, then patch an older version such as `20.21.37` or `20.26.46` and then the Shorts player should work with spoof app version

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

LisoUseInAIKyrios:
This same issue happened 2+ years ago when version spoofing to YouTube 16.x and 17.x still worked but the patched app was much more recent.

jpa102:
testing on 21.02.32 with patches 1.6.0-dev.4 spoofed to 19.35.36, i tried searching for flags and it appears `45632000` shows most of the shorts ui

<div align="center">
<details>
<table>
<thead>
<tr>
<th>unspoofed 21.02.32</th>
<th>spoofed 19.35.36 - 45632000 flag not blocked</th>
<th>spoofed 19.35.36 - 45632000 flag blocked</th>
</tr>
</thead>

<tbody>

<tr>
<td>
<img width="444" height="989" alt="Image" src="https://github.com/user-attachments/assets/4908ba85-7f78-4d00-93cb-b8c120b421b4" />
</td>

<td>
<img width="444" height="988" alt="Image" src="https://github.com/user-attachments/assets/8d6e837a-4362-4b6b-99a1-83171c60119c" />
</td>

<td>
<img width="445" height="989" alt="Image" src="https://github.com/user-attachments/assets/d20f917e-4198-4067-acd8-3907d97ffd63" />
</td>
</tr>

</tbody>
</table>
</details>
</div>

jpa102:
oddly enough, the sound source (below share button) is missing and turns out it's `45632045` that's controlling it

<div align="center">
<details>
<table>
<thead>
<tr>
<th>45632045 flag not blocked</th>
<th>45632045 flag blocked</th>
</tr>
</thead>

<tbody>

<tr>
<td>
<img width="444" height="989" alt="Image" src="https://github.com/user-attachments/assets/81ef7300-c83b-4021-b947-2a621c08d224" />
</td>

[...]

PR #375, opened 2026-02-02
feat(YouTube - Spoof app version): Add target version `20.28.41`

inotia00:
### Add target version `20.28.41 - Disable bold icons`

- Starting with YouTube 20.39, bold icons are also applied to the video action bar (Enforced by the endpoint response).

- If the user is using YouTube 20.39+, bold icons will still be applied to the video action bar even if `Disable bold icons` is enabled.

- Since the bold icons are now somewhat dependent on the endpoint response, it would be more ideal to disable the bold icons by `Spoof app version` instead of `Disable bold icons`.

### Fix - Shorts player has no overlay buttons (Closes https://github.com/MorpheApp/morphe-patches/issues/183)

- YouTube 20.28 or earlier returns an endpoint response for the old Shorts overlay.

- Even if the app version is spoofed as 20.28 or earlier, the new Shorts overlay flags (`45632000` and `45632045`) are still used on the client side, preventing the Shorts overlay from being drawn.

- This PR forces the flags for the new Shorts overlay to be disabled when the app version is spoofed to 20.28 or earlier.

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/spoof/SpoofAppVersionPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/spoof/SpoofAppVersionPatch.java
@@ -1,5 +1,6 @@
 package app.morphe.extension.youtube.patches.spoof;
 
+import app.morphe.extension.youtube.patches.VersionCheckPatch;
 import app.morphe.extension.youtube.settings.Settings;
 
 @SuppressWarnings("unused")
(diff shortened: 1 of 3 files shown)


extensions/music/src/main/java/app/morphe/extension/music/settings/MusicActivityHook.java
```diff
@@ -26,8 +25,17 @@ public class MusicActivityHook extends BaseActivityHook {
     @SuppressLint("StaticFieldLeak")
     public static MusicSearchViewController searchViewController;
 
+    /**
+     * How much time has passed since the first launch of the app. Simple check to prevent
+     * forcing bold icons on first launch where the settings menu is partially broken
+     * due to missing icon resources the client has not yet received.
+     */
+    private static final long MINIMUM_TIME_AFTER_FIRST_LAUNCH_BEFORE_ALLOWING_BOLD_ICONS = 30 * 1000; // 30 seconds.
```
LisoUseInAIKyrios, line 33: I haven't recently checked, but if YT is now showing bold icons for all users with 20.31 then this force bold logic can be removed.

Issue #358, opened 2026-02-01
bug(YouTube): Empty space is left when ads are hidden on tablets


Rainman67:
### Bug description

Perhaps someone can help me. Since installing patches 1.80 and 1.90, I always have an empty field in the overview in the upper left corner. Everything was fine up to and including version 1.70. Does anyone have any idea what the cause could be? The patch selection is unchanged.

![Image](https://github.com/user-attachments/assets/8b5e7726-36bb-4ea7-925a-98dccdc9bc2d)

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

LisoUseInAIKyrios:
What version YT did  you patch?

It looks like you have some non default in-app settings (narrow navigation buttons?). Backup your Morphe YT settings, clear the settings, save, and restart.

BlackGold8282:
@LisoUseInAIKyrios 

This is a tablet, not a phone.

Rainman67:
> This is a tablet, not a phone.

And that seems to be the problem. I reinstalled the app as recommended. The first setting I changed was the format, which I changed from "Standard" to "Tablet". That's what's causing the error. As I said, everything works correctly with version 1.70.

BlackGold8282:
I know that if mix playlist is hidden in a very old tablet layout, there will be empty space.

I don't know if that problem still occurs.

Rainman67:
The setting "hide mix playlists" is not activated.

LisoUseInAIKyrios:
YouTube seems to have changed the tablet layout for newer app targets. When using an tablet emulator and 20.40 I could not see any difference in the layout compared to a phone emulator.

This may be a side effect of forcing tablet mode.

Try patching an older version of the app such as 20.31

Rainman67:
Trying 20.31.39 I get the message "no patches available"

Rainman67:
But what I don't understand is that everything was fine with patches 1.70. If I reinstall the APK created with them, then everything is working perfectly again. Therefore, something must have changed in the patches...?

LisoUseInAIKyrios:
`20.31.42` is supported.

Are you using pre-release patches? Try turning off and patch with stable patches.

Rainman67:
[...]

PR #377, opened 2026-02-02
fix(YouTube - Hide ads): Empty space is left when ads are hidden on tablets

inotia00:
Closes https://github.com/MorpheApp/morphe-patches/issues/358.

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/AdsFilter.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/AdsFilter.java
@@ -253,6 +253,15 @@ public final class AdsFilter extends Filter {
     }
 
     /**
+     * Injection point.
+     */
+    public static String hideAds(String osName) {
+        return Settings.HIDE_GENERAL_ADS.get()
+                ? "Android Automotive"
+                : osName;
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


inotia00:
### Additional context

1. Hiding elements from the home feed via a litho filter leaves a empty space on tablets.

2. This is why turning on `Hide mix playlists` on tablets leaves a empty space.

3. https://github.com/MorpheApp/morphe-patches/issues/358 is likely due to ads, not mix playlists.

4. Previously, when `Switch Create with Notifications` was enabled, the OS name was spoofed to `Android Automotive`, and this OS name was applied to all endpoints.

5. Since https://github.com/MorpheApp/morphe-patches/pull/260 (1.8.0-dev.6), the OS name was spoofed only on some endpoints to ensure independent operation of the patch.

6. As of 1.8.0-dev.6, the OS name is no longer spoofed on the `/browse` and `/search` endpoints, so the response from the home feed contains ads (Filtered by litho filter).

7. On tablets, hiding elements from the home feed via litho filter leaves a empty space: https://github.com/MorpheApp/morphe-patches/issues/358

8. The workaround for this issue is to spoof the OS name on the `/browse` and `/search` endpoints (Related setting is `Hide general ads`).

ILoveOpenSourceApplications:
LGTM?

PR #400, opened 2026-02-03
chore(Hex patch): Fix bug in implementation of Boyer-Moore algorithm

hoo-dles:
Well, this was a miserable bug...

The current implementation was not matching the last byte of my pattern, so it was replacing a different byte string somewhere else in the binary. Really annoying when some of your hex patches work (still unique with 1 byte less), and then one doesn't.

Honestly not sure how this bug survived so long. I think this fix is sufficient, but after I tested it I didn't have any more brain capacity to check it against existing psuedocode of the algorithm. So maybe someone double check me.

Anyway, I have less hair now and probably a few more points on my blood pressure.

### DIFF
--- a/patches/src/main/kotlin/app/morphe/patches/shared/misc/hex/HexPatchBuilder.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/shared/misc/hex/HexPatchBuilder.kt
@@ -110,18 +110,15 @@ class Replacement(
      */
     private fun indexOfPatternIn(haystack: ByteArray): Int {
         val needle = bytes
-
-        val haystackLength = haystack.size - 1
-        val needleLength = needle.size - 1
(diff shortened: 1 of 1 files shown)


LisoUseInAIKyrios:
Sorry for the inconvenience in the old code.

We're also cleaning up and fixing code we didn't write.

Will eventually move this class to the `patches-utils` project when it's created. Been very busy with things.

hoo-dles:
No worries, I know this code has been sitting around for a while.

hoo-dles:
> Will eventually move this class to the `patches-utils` project when it's created. Been very busy with things.

By the way, ping me when you guys get around to this. I have an extension of the hex patch for Hermes bytecode that regenerates the footer hash, which may be useful to others who are patching React Native with Morphe.

PR #404, opened 2026-02-03
fix(YouTube - Hide layout components): `Hide subscribed channels bar` leaves empty space

inotia00:
Closes https://github.com/MorpheApp/morphe-patches/issues/399.

Tested on YT 20.14 and 21.05.

PR #421, opened 2026-02-04
fix(YouTube - Hide layout components): Resolve "Hide subscribed channels bar" leaves empty space in landscape mode

inotia00:
This is a follow-up PR to https://github.com/MorpheApp/morphe-patches/pull/404.

From YouTube 20.21 to 20.40, there were two very similar classes.

Since both classes had almost identical methods, they could not be distinguished from each other by the fingerprint in https://github.com/MorpheApp/morphe-patches/pull/404.

This is why it didn't work on YouTube 20.40.45.


ILoveOpenSourceApplications:
By the way, was the restart dialog for the tablet users, from the previous PR?

Issue #100, opened 2026-01-05
bug(YouTube - Hide layout components): "Hide Show more button" does not work


ILoveOpenSourceApplications:
### Bug description

"Hide Show more button" only hides the text and the drop down icon making it look like a big empty space but still leaves behind the button area and action of expanding when tapping it.

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

LisoUseInAIKyrios:
Yes it's a side effect of the way the hide was changed to support newer app targets.

If I remember correctly @0xrxL found how to hide it without showing the empty space. Comments are somewhere in the 11 month long discussion in the unmerged pull request I first opened 1 year ago in ReVanced.

0xrxL:
> Yes it's a side effect of the way the hide was changed to support newer app targets.
> 
> If I remember correctly [@0xrxL](https://github.com/0xrxL) found how to hide it without showing the empty space. Comments are somewhere in the 11 month long discussion in the unmerged pull request I first opened 1 year ago in ReVanced.

Yeah...to hide it properly you need to nullify the method where ```R.layout.expand_button_down``` is placed.

Just that...

ILoveOpenSourceApplications:
Is the above fix achieved in a similar manner to #187?

LisoUseInAIKyrios:
Yes.

PR #422, opened 2026-02-04
fix(YouTube - Hide layout components): `Hide Show more button` leaves empty space

inotia00:
- Fix: `Hide Show more button` leaves empty space.
- Fix: `Hide Show more button` hides the button not only in search results but also in channels.
- Tested on YouTube 20.14, 20.21, 20.40, and 21.05.


ILoveOpenSourceApplications:
Closes #100.

PR #466, opened 2026-02-08
refactor: Clean up code

inotia00:
### Description

In the process of adding Reddit support, unused dummy classes were added, or constants were placed in the wrong places.

- Unused dummy classes have been removed.

- Fixed an issue where all compatibility constants were located in the Reddit directory.

- Fixed an issue where compatibility constants were not used in some YouTube patches.

- Removed the `Restore old search filters` patch, which no longer works.

- Removed support versions that do not have any significant meaning: https://github.com/MorpheApp/morphe-patches/pull/466#discussion_r2778688140.

- Changed the `Spoof signature` patch from a shared patch to a Reddit-targeted patch, as it only works on Reddit.

- Changed the `Override certificate pinning` patch to a available patch, as it's used to analyze network packets.

### Test results


inotia00:
Reddit patches still don't use the latest Morphe Patcher syntax.

I'd like to rewrite Reddit patches using the latest Morphe Patcher syntax, but if https://github.com/MorpheApp/morphe-patches/pull/347 is merged, some fingerprints will have to be rewritten.

So, I'm hesitant to do this in this PR.

@LisoUseInAIKyrios What would be a better approach? Should I rewrite the Reddit patch in this PR, or should I wait until https://github.com/MorpheApp/morphe-patches/pull/347 is merged?

patches/src/main/kotlin/app/morphe/patches/youtube/shared/Constants.kt
```diff
@@ -0,0 +1,18 @@
...
+internal object Constants {
+    val COMPATIBILITY: Pair<PackageName, Set<VersionName>> = Pair(
+        "com.google.android.youtube",
+        setOf(
+            "20.40.45",
+            "20.37.48",
+            "20.31.42",
+            "20.26.46",
+            "20.21.37",
+            "20.14.43",
```
inotia00: **20.21.37** is the last version that supports all litho filters.
**20.26.46** is the last version that supports Android 8.0.
**20.31.42** is the first version to support bold icons and is a good comparison to 20.40.45.

**20.14.43** and **20.37.48** are not significant versions, unlike the versions mentioned above, so I would like to remove them from the supported versions.

LisoUseInAIKyrios: Makes sense and I agree.

Although I wonder if 20.26.46 should be dropped because 20.21.37 is a much better choice for Android 8.0. All filters work with 20.21, and there is a performance advantages to the older non native buffer filtering which may be particularly relevant for older slower devices. 

20.26 doesn't offer bold layout or any new noteworthy changes either.

LisoUseInAIKyrios:
> @LisoUseInAIKyrios What would be a better approach? Should I rewrite the Reddit patch in this PR, or should I wait until #347 is merged?

I would wait until 347 is merged.

patches/src/main/kotlin/app/morphe/patches/youtube/layout/branding/header/ChangeHeaderPatch.kt
```diff
@@ -96,7 +96,7 @@ val changeHeaderPatch = resourcePatch(
         changeHeaderBytecodePatch
     )
 
-    compatibleWith(COMPATIBILITY_YOUTUBE)
+    compatibleWith(COMPATIBILITY)
```
LisoUseInAIKyrios: Some linting tools (SpotBugs) will warn against shadowing class names in the same project.

To keep from accidentally importing the wrong compatibility, I think the shared field can keep the app name `COMPATIBILITY_YOUTUBE`.

patches/src/main/kotlin/app/morphe/patches/shared/misc/metadata/PackageMetaDataPatch.kt
```diff
@@ -0,0 +1,43 @@
...
+        // com.google.android.youtube
+        packageName = packageMetadata.packageName
+
+        // 20.40.45
+        packageVersion = packageMetadata.packageVersion
+
+        // 2040
+        packageVersionSimplified = packageVersion.split(".")
+            .take(2)
+            .joinToString("")
```
inotia00: **[This commit will be reverted]**

If we change the access modifier of the [`packageMetadata` parameter of ResourcePatchContext](https://github.com/MorpheApp/morphe-patcher/blob/e9ce5f6ae31ea1922eeffee4dbaa35387a028fa1/src/main/kotlin/app/morphe/patcher/patch/ResourcePatchContext.kt#L29) to *public* (or remove), we can get the package name or app version without kotlin reflection.

inotia00: @LisoUseInAIKyrios I remember a similar discussion before, but I couldn't find which repo it was created in.

I've [committed example code](https://github.com/MorpheApp/morphe-patches/pull/466/changes/44740e7797dc6f3854da09f150dd89a5a62f3fb0) that finds the package name and app version from a resource patch.

Please refer to this code when preparing the upcoming Morphe Patcher.

LisoUseInAIKyrios: Is https://github.com/MorpheApp/morphe-patcher/issues/45

LisoUseInAIKyrios: Someone can make dev change to patcher and expose the field. Should also briefly mention it in the patcher docs.

inotia00: #### When the `google_play_services_version` is used:

4ms

#### When Kotlin reflection is used:

245ms

#### When the access modifier of the packageMetadata parameter of ResourcePatchContext is set to public:

0ms

LisoUseInAIKyrios (approved):
Anything left to do before merging?

inotia00:
Can be merged now.

Issue #413, opened 2026-02-03
feat(YouTube): add 20.39.41 as a spoof app version target


jpa102:
### Feature description

this particular app version seems to be the last version to feature the checkboxes when inside save to playlist, any version greater than that and it's the bookmark version where the user can only tap once in a playlist

some screenshots:

<div align="center">
<details>
<table>
<thead>
<tr>
<th>unspoofed 21.02.32</th>
<th>spoofed 20.39.41</th>
</tr>
</thead>

<tbody>

<tr>
<td>
<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/abc0faad-ce81-4775-90f5-de1b07267660" />
</td>

<td>
<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/582b4fae-1f4e-4f77-bbd8-33d0f7a7e314" />
</td>
</tr>

</tbody>
</table>
</details>
</div>

the name of the spoof app version target could be called something like:

* 20.39.41: Restore old Save to playlist design

### Motivation

* for easy adding a video to multiple playlists
* easily undo adding a video to a playlist
* restoring a part of an old ui design

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] This is a feature request for YouTube, YT Music, or Reddit

LisoUseInAIKyrios:
Is it confirmed that spoofing to that version gives the feature?

The spoof target can be manually edited in the import/export settings.

jpa102:
from what i can observe when manually editing `spoof_app_version_target`, it does change the look of the save to playlist ui

all i did was switch back and forth between 20.39.41 and 20.40.45 on my 21.02.32 version with patches 1.11.0-dev.3

RichardReallyHard:
the new save to playlist is so annoying
will spoof to older app version break new version features ?
is it possible to at least compact the new list & keep the popup persisting by html or css ?

LisoUseInAIKyrios:
Spoofing to old versions will prevent most/all 20.40+ features from showing.

20.39 is a very recent spoof target so no issues are expected.

Changing YouTube UI elements is very difficult, especially something like this.

PR #493, opened 2026-02-10
feat(YouTube): Add `Disable layout updates` patch

### LINKED ISSUE #490: feat(YouTube): RVX "Disable Layout updates" patch
### Feature description

Port rvx's Disable Layout updates patch for youtube so the profile is at the top. So the profile icon is next to the notification icon up top. It looks cleaner. 

### Motivation

Just to make the look of Morphe more adaptive for myself with my profile icon being up top. And I'm sure a lot of people will appreciate as well. 

### Acknowledgements

LisoUseInAIKyrios:
### Description

Adapts patch from @inotia00  https://github.com/inotia00/revanced-patches

Closes #490

### Additional context

The 'Change accounts' button will not show in the library tab, but changing the account can be done thru:

`Home feed settings menu > Tap YouTube account name > Tap different account`

### Test results

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/DisableLayoutUpdatesPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/DisableLayoutUpdatesPatch.java
@@ -1,0 +1,33 @@
+package app.morphe.extension.youtube.patches;
+
+import java.util.List;
+
+import app.morphe.extension.shared.Logger;
+import app.morphe.extension.youtube.settings.Settings;
+
+@SuppressWarnings("unused")
+public class DisableLayoutUpdatesPatch {
+
+    private static final List<String> REQUEST_HEADER_KEYS = List.of(
+            "X-Youtube-Cold-Config-Data",
+            "X-Youtube-Cold-Hash-Data",
+            "X-Youtube-Hot-Config-Data",
+            "X-Youtube-Hot-Hash-Data"
+    );
+
+    private static final boolean DISABLE_LAYOUT_UPDATES = Settings.DISABLE_LAYOUT_UPDATES.get();
+
+    /**
+     * @param key   Keys to be added to the header of CronetBuilder.
+     * @param value Values to be added to the header of CronetBuilder.
+     * @return Empty value if setting is enabled.
+     */
+    public static String disableLayoutUpdates(String key, String value) {
+        if (DISABLE_LAYOUT_UPDATES && REQUEST_HEADER_KEYS.contains(key)) {
+            Logger.printDebug(() -> "Blocking: " + key);
+            return "";
+        }
+
+        return value;
+    }
+}
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/settings/Settings.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/settings/Settings.java
@@ -281,6 +281,7 @@ public class Settings extends BaseSettings {
     public static final BooleanSetting HIDE_PLAYER_FLYOUT_WATCH_IN_VR = new BooleanSetting("morphe_hide_player_flyout_watch_in_vr", FALSE);
 
     // General layout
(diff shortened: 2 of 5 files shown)


ReplacementFit4095:
I noticed that the save to playlist ui also looks different (referencing issue #413 for this) when the setting is enabled

Can this info be also included in the "Changes include"? I only did some short testing and it appears to be consistent when turning it off and on

LisoUseInAIKyrios:
How does it look different? 

There are some other changes not listed in the setting dialog text. The only important ones to list are limitations.

If the difference can be stated concisely in one line, then can maybe can mention it.

ReplacementFit4095:
Like when the Disable layout updates setting is on, the icons in the platlist ui change from bookmark style to the one where you can tap on the checkboxes, as shown in the referenced issue

> There are some other changes not listed

I see, this is one of the side effects I've seen while I was using the patch, along with these:

* Icons inside the Explore tab (sidebar) reverting back to the thin variants and missing entries (YouTube Premium, YouTube Studio, etc...)
* Missing Languages entry in YouTube settings

LisoUseInAIKyrios:
Some of the new 21.+ menus might also be missing. And maybe more in the future.

"Missing Languages entry in YouTube settings"

Could be summarized into something general like:
"Some YouTube settings menus are missing"

Make a pull request with the proposed change if you'd like.

ghost:
This patch has always been really troublesome, I just had an issus where I couldnt hide Additional Settings from the player quality menu. It was DLU patch that was causing issue even though it was turnedd off. So even if you add this patch and then never touch it doesnt mean that it wont cause layout issues. My tip is, if you are having layout issues, the best thing to do it to turn DLU patch on and off again. This could be mentioned underneath the patch.

inotia00:
This patch forces the cold/hot config and hash (session data) to be replaced with empty values.

Session data is used for all endpoint requests, and can be updated with the response to each endpoint request.

(Session data is not only used to enable feature flags, but also to prove that the user is not a bot, as session data flow should be maintained if a legitimate client is being used)

If `Disable layout update` has been enabled even once, a session data conflict may occur. In this case, the layout may not change even if `Disable layout update` is turned off again.

The most reliable workaround is to clear the app data.

Replacing or modifying session data with empty values ​​can result in negative evaluations by Google's bot detection system, so I personally do not recommend using `Disable layout update`.

LisoUseInAIKyrios:
The setting can me modified to include a user dialog when trying to turn it on, and strongly suggest the user should clear their app data if they later turn off the feature.

Similar wording to the spoof app version setting.

ghost:
> This patch forces the cold/hot config and hash (session data) to be replaced with empty values.
> 
> Session data is used for all endpoint requests, and can be updated with the response to each endpoint request.
> 
> (Session data is not only used to enable feature flags, but also to prove that the user is not a bot, as session data flow should be maintained if a legitimate client is being used)
> 
> If `Disable layout update` has been enabled even once, a session data conflict may occur. In this case, the layout may not change even if `Disable layout update` is turned off again.
> 
> The most reliable workaround is to clear the app data.
> 
> Replacing or modifying session data with empty values ​​can result in negative evaluations by Google's bot detection system, so I personally do not recommend using `Disable layout update`.

Looks like it can cause issues even though its not touched once, like in my case. I had it turned off since the day I patched it, but still it somehow caused layout issues and broke other patches. It should not do that. And in my case not even clearing data/cache didnt help, once you add this patch in it has a chance to cause layout issues. So like I said, turning it on/off, or off/on might fix layout issues you are having. You could also add a warning in the morphe manager about this patch, like its recommended to exclude this patch to avoid layout issues and patch breakage.

LisoUseInAIKyrios:
If the patch is off in the settings then it's not doing anything. Whatever you experienced is something unrelated to layout updates.

ghost:
> If the patch is off in the settings then it's not doing anything. Whatever you experienced is something unrelated to layout updates.

But its weird because it got fixed once i toggled this setting on, restarted, turned it off, restarted.

LisoUseInAIKyrios:
It probably forced your client to get new persistent layout data. The same would have happened if you cleared the app data.

ghost:
> It probably forced your client to get new persistent layout data. The same would have happened if you cleared the app data.

Nope, clearing app data didnt work in my case. I had to touch this setting. I might do some testing in the future where I exclude this patch, just to see if I have the same issue or not.

PR #537, opened 2026-02-14
fix(Spoof video streams): Close streams in JavaScriptManager.readFromFile

saieshshirodkar:
This fixes a resource leak where FileInputStream and BufferedReader were not being closed on Android API < 26.

The bug could cause:  Memory leaks,File descriptor exhaustion ,App crashes after repeated calls

Fixed by using try-with-resources to ensure streams are always closed properly.

### DIFF
--- a/extensions/shared/library/src/main/java/app/morphe/extension/shared/spoof/js/JavaScriptManager.java
+++ b/extensions/shared/library/src/main/java/app/morphe/extension/shared/spoof/js/JavaScriptManager.java
@@ -426,11 +426,12 @@ public final class JavaScriptManager {
                 //noinspection ReadWriteStringCanBeUsed
                 return new String(Files.readAllBytes(file.toPath()), StandardCharsets.UTF_8);
             } else {
-                FileInputStream fis = new FileInputStream(file);
-                BufferedReader reader = new BufferedReader(new InputStreamReader(fis));
-                String line;
-                while ((line = reader.readLine()) != null) {
-                    sb.append(line).append("\n");
+                try (FileInputStream fis = new FileInputStream(file);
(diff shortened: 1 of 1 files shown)


LisoUseInAIKyrios:
Rebase off dev, and then resolve/apply these changes. The files were recently refactored and some moved to different locations.

error-reporting:
> This fixes a resource leak where FileInputStream and BufferedReader were not being closed on Android API < 26.
> 
> The bug could cause: Memory leaks,File descriptor exhaustion ,App crashes after repeated calls
> 
> Fixed by using try-with-resources to ensure streams are always closed properly.

hey can you extend this to Android api 33 please as morphe tv client has occasional playback issue after every few video or app running for a long time without restart

LisoUseInAIKyrios:
This issue would affect all android releases. YouTube requires Android 8.0+, which means it's Android 26+

LisoUseInAIKyrios:
?

error-reporting:
> ?

yeah i don't know why he deleted it either

kitadai31:
If both YouTube and YT Music are Android 8.0+, this `else` block is an unreachable code.

I think Morphe should increase minSdk to 26 for all extensions.

error-reporting:
> If both YouTube and YT Music are Android 8.0+, this `else` block is an unreachable code.
> 
> I think Morphe should increase minSdk to 26 for all extensions.

yt music has available versions usable for android version down to 6.0

LisoUseInAIKyrios:
I didn't notice this was in an else block.

The change could still be made in case support for YT Music for Android 6.0 is later added.

inotia00:
The memory leak issue is not caused by this.

It is caused by the V8 runtime. It will be fixed in an upcoming PR.

Issue #474, opened 2026-02-08
bug(YouTube Music): Mount (root) install always crashes on launch


kastentop2005:
### Bug description

Yes. It just crashes. Tried with 8.40.54 and 9.05.51

Regular YouTube works

<img width="1080" height="2209" alt="Image" src="https://github.com/user-attachments/assets/e841d36a-9c5c-4e65-a95a-b32e9eb0c46a" />

### Error logs

```shell
FATAL EXCEPTION: main
Process: com.google.android.apps.youtube.music, PID: 13479
java.lang.UnsatisfiedLinkError: dlopen failed: library "libelements.so" not found
	at java.lang.Runtime.loadLibrary0(Runtime.java:1090)
	at java.lang.Runtime.loadLibrary0(Runtime.java:1012)
	at java.lang.System.loadLibrary(System.java:1765)
	at isy.<clinit>(PG:15)
	at java.lang.Class.newInstance(Native Method)
	at android.app.AppComponentFactory.instantiateApplication(AppComponentFactory.java:76)
	at androidx.core.app.CoreComponentFactory.instantiateApplication(PG:1)
	at android.app.Instrumentation.newApplication(Instrumentation.java:1350)
	at android.app.LoadedApk.makeApplicationInner(LoadedApk.java:1601)
	at android.app.LoadedApk.makeApplicationInner(LoadedApk.java:1533)
	at android.app.ActivityThread.handleBindApplication(ActivityThread.java:8596)
	at android.app.ActivityThread.-$$Nest$mhandleBindApplication(Unknown Source:0)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2826)
	at android.os.Handler.dispatchMessage(Handler.java:110)
	at android.os.Looper.loopOnce(Looper.java:273)
	at android.os.Looper.loop(Looper.java:363)
	at android.app.ActivityThread.main(ActivityThread.java:9983)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:632)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:975)
```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

LisoUseInAIKyrios:
```
java.lang.UnsatisfiedLinkError: dlopen failed: library "libelements.so" not found
```

[...]

PR #564, opened 2026-02-17
fix(Spoof video streams): Using the TV client can cause a V8 runtime memory leak

inotia00:
### Fixed a V8 runtime memory leak issue

- To prevent excessive memory leaks in the V8 runtime, the runtime is periodically released (closed).

- Videos may start approximately 0.2 seconds later when the V8 runtime is re-created, but this prevents memory leaks.

### `TV` has been removed from the `Player JavaScript variant` setting

- Among the `Player JavaScript variant` settings, `TV` was removed as it did not appear to work for some users.

- The default value has been changed to `Phone` (Mobile Web).

### The User-Agent of the `TV` client has changed

- It's not clear, but it seems like the JS challenge gets a little easier the older the User-Agent is used.

### `Spoof video streams` patch has been integrated into the `GmsCore support` patch

- Closes https://github.com/MorpheApp/morphe-patches/issues/474.

### yt-dlp-ejs has been updated

### DIFF
--- a/extensions/music/src/main/java/app/morphe/extension/music/patches/spoof/SpoofVideoStreamsPatch.java
+++ b/extensions/music/src/main/java/app/morphe/extension/music/patches/spoof/SpoofVideoStreamsPatch.java
@@ -19,9 +19,9 @@ public class SpoofVideoStreamsPatch {
     public static void setClientOrderToUse() {
         // For some users No SDK can fail at 1 minute. Only use it if the user has explicitly set it.
         List<ClientType> availableClients = List.of(
+                TV,
                 ANDROID_VR_1_47_48,
                 VISIONOS,
-                TV,
                 ANDROID_VR_1_54_20
         );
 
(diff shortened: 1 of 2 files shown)


fromhgbwithluv:
The libj2v8.so files seem to be used only to check if the application supports J2V8, so I believe they should be removed.
P/s: The v8 engine mainly come from `com.eclipsesource.v8.V8`

inotia00:
> The libj2v8.so files seem to be used only to check if the application supports J2V8, so I believe they should be removed. P/s: The v8 engine mainly come from `com.eclipsesource.v8.V8`

When the V8 runtime is created, `System.loadLibrary(j2v8)` (libj2v8.so) is called internally: https://github.com/eclipsesource/J2V8/blob/ef51daad093f21f714727b704ec740a5fb1e5f1e/src/main/java/com/eclipsesource/v8/LibraryLoader.java#L84

If the native library does not exist, a fatal exception is thrown.

PR #575, opened 2026-02-18
feat: Add `Disable Play Store updates`, `Custom branding name for Reddit`, improve `Change package name`

wchill:
### Description

- Adds `Disable Play Store updates` (previously called `Change version code`).
- Adds `Custom branding name for Reddit` (currently only changes the app launcher name).
- Changes universal `Change package name` to work with Reddit and other apps that previously did not work with the patch.

### Additional context

Do not merge until all of the below have been merged.
- [x] Patcher: https://github.com/MorpheApp/morphe-patcher/pull/48 
- [x] CLI: https://github.com/MorpheApp/morphe-cli/pull/55
- [x] Manager: https://github.com/MorpheApp/morphe-manager/pull/273

### Test results


cyberboh:
Is it possible to add `Change version code` patch into Reddit? It can be used to hide update appear in play store.

wchill:
It shouldn't be necessary with the change package name patch.

inotia00:
iirc, if the `Change package name` patch is applied to Reddit, you can not log in to your account.

See: https://github.com/inotia00/ReVanced_Extended/issues/2912.

The last version that allowed login after applying the `Change package name` patch was Reddit 2023.39.1, but Reddit recently blocked logins with that version.

cyberboh:
> It shouldn't be necessary with the change package name patch.

It is too risky, tbh. `Changing the version code` is safer. I’ve been using it on an old `Twitter` version `10.49.0` for almost a year with no issues so far. If it’s possible to adapt that patch to `Reddit`, it would be much more useful, otherwise it might not be worth the risk.

wchill:
I found some spots in the login flow where the modified package name is being sent, I'll try and adapt the patch to work.

LisoUseInAIKyrios:
Does changing the package name affect "open with" for other apps that can open/send to reddit?

I'm not sure if any apps do this, but I think changing package name will break that functionality.

wchill:
I don't think there are any apps that do that. I imagine everything just goes through the normal intent handler system.

patches/src/main/kotlin/app/morphe/patches/reddit/layout/branding/name/CustomBrandingNamePatch.kt
```diff
@@ -0,0 +1,67 @@
...
+        document("res/values-v24/strings.xml").use { document ->
+            mapOf(
+                "app_name" to appName
+            ).forEach { (k, v) ->
+                val stringElement = document.createElement("string")
+
+                stringElement.setAttribute("name", k)
+                stringElement.textContent = v
+
+                document.getElementsByTagName("resources").item(0).appendChild(stringElement)
```
kitadai31, line 64:
```suggestion
                document.documentElement.appendChild(stringElement)
```

In general, to get the root element of the document (in this case `<resources>`), use  `documentElement`.
(It is `getDocumentElement()` replaced by kotlin's implicit getters.)

In addition, getting the same element each time in a loop can be slow. (Especially when calling `childNodes`)
However, since it only loops once in this case, it should be fine as is.

wchill:
@inotia00 the login not working is due to a recaptcha API call that fails when the name is changed. Once this is patched, this patch works again.

The package name does leak through in other telemetry, but so does the app name, so YMMV when it comes to reddit detecting or caring about it.

image

LisoUseInAIKyrios:
Maybe change package name should also bytecode patch and change all usage of `context.getPackageName()` to use the original package name. Can extract the logic so the generic change package name and the reddit specific patch can use it (or reddit change package name patch depends on universal change package name).

patches/src/main/kotlin/app/morphe/patches/reddit/layout/branding/packagename/ChangePackageNamePatch.kt
```diff
@@ -0,0 +1,133 @@
...
+            writeText(
+                readText()
+                    .replace(
+                        "package=\"$PACKAGE_NAME_REDDIT",
+                        "package=\"$redditPackageName"
+                    )
+                    .replace(
+                        "$PACKAGE_NAME_REDDIT.DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION",
+                        "$redditPackageName.DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION"
+                    )
```
LisoUseInAIKyrios: Some of this functionality is already in the universal change package name patch:

```
            if (updatePermissions == true) {
                val permissions = manifest.getElementsByTagName("permission").asSequence()
                val usesPermissions = manifest.getElementsByTagName("uses-permission").asSequence()

                val receiverNotExported = "DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION"

                (permissions + usesPermissions)
                    .map { it as Element }
                    .filter { it.getAttribute("android:name") == "$packageName.$receiverNotExported" }
                    .forEach { it.setAttribute("android:name", "$newPackageName.$receiverNotExported") }
            }
```

Can all the reddit specific logic be moved into the universal change package name, and the only reddit specific patch is change app name?  We can add reddit specific logic to the universal patch that only applies if the original package name is Reddit. That would simplify two patches (a universal change package name and a Reddit specific patch), into just one universal patch.

Should change app name be a separate universal patch of it's own?

wchill:
Changing all instances of context.getPackageName() is likely to break things, hence I didn't do this.

LisoUseInAIKyrios:
Changing the bytecode package name could be a patch option in the universal patch (default off).

optimatrix0:
I’ve been following the discussion about the version code patch, and I’d like to raise a few important points.

First, not everyone wants to change the package name, especially if it’s not strictly necessary for a given build. Even though the package name can be altered, that approach is usually reserved for clone versions of an app.

Second, whenever possible, keeping the original package name provides better compatibility overall. For less technical users in particular, having the original Reddit package name would be more practical, as it would reduce the need for additional modifications just to make the app work — and it could also lower your maintenance workload.

Another factor is that this is a community-driven project. Adding support for version code changes is something users are actively requesting. If implemented as a universal patch, it would especially benefit non-root devices, which represent the vast majority of users today. Root usage has dropped significantly due to major changes from Google, so solutions that work without root are increasingly important.

In short, making Morph compatible with version code modification would help avoid package conflicts, reduce the need for repackaging, and respond directly to community demand. I understand this may not align perfectly with your ideal design, but I hope you’ll consider it and share your thoughts.

Personally, I just want to be able to open Google Play, tap “Update all,” and check for updates without issues. Right now, that’s not possible because Reddit remains out of date.

If this approach isn’t acceptable, then setting a different package name as the default could at least prevent conflicts from happening in the first place.

optimatrix0:
Huge congratulations to the team. Reducing the RAM requirements for the patching process is something that will benefit everyone — especially users like me on mid-range devices (hello from a Moto G75 😄).

Another great point is how fast this project is evolving. Compared to similar efforts, the pace of development here has been impressive, and that’s fantastic for us users.

Please don’t interpret my previous comment as negativity. At most, it’s constructive criticism — just an idea I thought might be helpful, even if it may or may not fit your vision.

That said, if you ever decided to switch the Reddit package name to something aligned with Morph apps (for example, app.morph.* similar to what’s done for YouTube or YouTube Music), that would also be perfectly fine — arguably even better. In that scenario, updates could be fully managed by the manager, and users wouldn’t have to deal with Google Play flagging the original app as outdated.

Either way, your project is a solid 10/10. It’s refreshing to see active development and responsiveness to the community. I also recently caught up on the broader drama around similar projects — including the PR that sat in review for 11 months — and yeah, that kind of situation is frustrating for anyone who has followed those spaces for a while.

Keep up the amazing work.

wchill:
Universal version code patch is fine, I just am not going to add it to this PR. Main reason for me to add a package name patch is so that I can test without having to juggle both the unpatched and patched apps on the same device.

wchill:
@LisoUseInAIKyrios I need a way to get the package name from a bytecode patch in order to do the universal patch for context.getPackageName(). So the related patcher change will need to land before I can write the code for it.

LisoUseInAIKyrios:
Does the matching patcher PR have that change? Add the change to that PR or make a separate if needed.

LisoUseInAIKyrios:
Change version code should be applied by default for Reddit.

I don't see any use case of allowing app store updates for any patched Reddit install 

And since we're trying to eliminate the problematic "pick and choose" patches of expert mode, that means the patch should be applied as a dependency of a core reddit patch (or use some other way change version code is applied by default to Reddit).

wchill:
inotia is already working on the change version code patch using the raw manifest, right? Though I suppose modifying the raw manifest might run into issues if AndroidManifest.xml is modified in other ways, so it would make sense to just implement it here.

LisoUseInAIKyrios:
With arsclib the universal version code patch will work, right? A raw patch strategy may not be needed if the universal patch can be used.

patches/src/main/kotlin/app/morphe/patches/all/misc/versioncode/ChangeVersionCodePatch.kt
```diff
@@ -0,0 +1,29 @@
+package app.morphe.patches.all.misc.versioncode
+
+import app.morphe.patcher.patch.intOption
+import app.morphe.patcher.patch.resourcePatch
+import app.morphe.util.getNode
+import org.w3c.dom.Element
+
+internal val changeVersionCodePatch = resourcePatch(
+    name = "Change version code",
```
LisoUseInAIKyrios: Just to be clear, this code is from RVX and was present before we recently refactored and deleted stuff.

patches/src/main/kotlin/app/morphe/patches/all/misc/versioncode/ChangeVersionCodePatch.kt
```diff
@@ -0,0 +1,29 @@
...
+    description = "Changes the version code of the app. This can be useful to prevent Play Store updates, but can also lead to unexpected issues.",
+    use = false
+) {
+    val versionCodeOption by intOption(
+        key = "versionCode",
+        default = 2100000000,
+        title = "Version code",
```
LisoUseInAIKyrios: Can we delete this option?

It has never made sense to use the lowest version code. That will permanently turn on app store nags.

It should always use the highest version code.

wchill:
Added universal version code patch and added it as a dependency to the settings patch.

For the package name, I modified BytecodePatchContext to just make the packageMetadata available

patches/src/main/kotlin/app/morphe/patches/all/misc/versioncode/ChangeVersionCodePatch.kt
```diff
@@ -0,0 +1,29 @@
+package app.morphe.patches.all.misc.versioncode
+
+import app.morphe.patcher.patch.intOption
+import app.morphe.patcher.patch.resourcePatch
+import app.morphe.util.getNode
+import org.w3c.dom.Element
+
+internal val changeVersionCodePatch = resourcePatch(
+    name = "Change version code",
+    description = "Changes the version code of the app. This can be useful to prevent Play Store updates, but can also lead to unexpected issues.",
```
LisoUseInAIKyrios: Should we rename this to `Disable app store updates` or something similar.

Users should not be forced to translate technical jargon onto something they can more easily understand.

 They should know what a patch does for them. `Hide ads`, `Disable swipe controls`, etc.

patches/src/main/kotlin/app/morphe/patches/all/misc/versioncode/ChangeVersionCodePatch.kt
wchill: I actually wrote this one from scratch, it's not copied from there. So it's original code

LisoUseInAIKyrios: Oh even better.

patches/src/main/kotlin/app/morphe/patches/reddit/layout/branding/name/CustomBrandingNamePatch.kt
```diff
@@ -0,0 +1,67 @@
...
+import java.io.FileWriter
+import java.nio.file.Files
+
+private const val APP_NAME = "Reddit Morphe"
+
+@Suppress("unused")
+val customBrandingNamePatch = resourcePatch(
+    name = "Custom branding name for Reddit",
+    description = "Changes the Reddit app name to the name specified in patch options.",
+) {
```
LisoUseInAIKyrios, line 22: Should this be default off?

patches/src/main/kotlin/app/morphe/patches/reddit/layout/branding/packagename/ChangePackageNamePatch.kt
```diff
@@ -0,0 +1,85 @@
...
+val changePackageNamePatch = baseChangePackageNamePatch(
+    name = "Change Reddit package name",
+    description = "Changes the package name for Reddit to the name specified in patch options.",
+    shouldUpdatePermissions = true,
+    shouldUpdateProviders = true,
+    shouldPatchGetPackageName = false,
+    block = {
+        compatibleWith(COMPATIBILITY_REDDIT)
+        dependsOn(spoofSignaturePatch, recaptchaPatch)
+    },
```
LisoUseInAIKyrios: Should we incorporate the Reddit specific changes into the universal change package name patch? Then it's just one patch instead of the possibility of picking two (universal change package name and Reddit specific).

Since disable play store updates patch (change version code) is now universal and a dependency for Reddit, that means all patched Reddit apps have app store updates off.  So changing package name could be an opt-in expert mode universal patch, just like it is for all other apps.

wchill: We can punt that decision for now, just remove the name so it disables the patch but leave it in

patches/src/main/kotlin/app/morphe/patches/reddit/layout/branding/name/CustomBrandingNamePatch.kt
wchill: Up to you, is it off by default for YouTube?

patches/src/main/kotlin/app/morphe/patches/reddit/layout/branding/packagename/ChangePackageNamePatch.kt
wchill: As for the actual decision here, I'd rather be in favor of coming up with a way to hide the universal patch in favor of the reddit-specific one when reddit is selected. Consolidating all package-specific changes into universal patches smells like an anti-pattern when the potential changes to fix breakage for each app are likely to be wildly different.

patches/src/main/kotlin/app/morphe/patches/reddit/layout/branding/name/CustomBrandingNamePatch.kt
wchill: Set it off for now

patches/src/main/kotlin/app/morphe/patches/reddit/layout/branding/packagename/ChangePackageNamePatch.kt
LisoUseInAIKyrios: Change package name is a little unique because it works for any app, which means it may need some app specific tweaks. If we can keep the app specific code isolated then I think it may be worth it to logically have just one patch where the end user doesn't need to know or care about how it works.

In the future maybe sub-patches can exist based on package name and Manager/CLI treats them as a single patch (base universal change package name + reddit change package).  Would need patcher support.

LisoUseInAIKyrios: We can keep this how it is for now. I hate refactoring dsl patches because the simple is now unnecessarily complicated by recreating abstract classes passing around blocks for methods. We should ditch DSL for patches soon.

PR #637, opened 2026-02-25
fix(YouTube - Hide end screen suggested video): Add user dialogs to minimize user confusion

### LINKED ISSUE #635: bug: end screen cards
### Bug description

Even though I have disabled end-screen cards and end-screen suggested videos, it still shows a suggested video. And it also auto-plays it. How do I stop auto-playing as well? On the "end-screen suggested video," there's this tip to go to "YouTube settings playback autoplay next video," but this option isn't available anywhere.

### Error logs

```shell

```

### Acknowledgements

utafrali:
Fixes #635

Simplifies the end screen suggested video hiding implementation by removing the auto-nav related fingerprints and unnecessary instruction lookups. The current approach was overcomplicated and relied on fingerprints that may not be reliable across different YouTube versions.

The simplified approach focuses solely on the layout change listener removal, which is the most stable way to prevent end screen cards from appearing. This reduces maintenance overhead and improves compatibility across versions.


patches/src/main/kotlin/app/morphe/patches/youtube/layout/hide/endscreensuggestedvideo/HideEndScreenSuggestedVideoPatch.kt
```diff
@@ -39,42 +36,16 @@ val hideEndScreenSuggestedVideoPatch = bytecodePatch(
...
-            }
+            endScreenMethod.addInstructionsWithLabels(
+                0,
+                """
+                    invoke-static {}, $EXTENSION_CLASS_DESCRIPTOR->hideEndScreenSuggestedVideo()Z
+                    move-result v0
+                    if-eqz v0, :show_end_screen_recommendation
+                    return-void
+                """,
+                ExternalLabel("show_end_screen_recommendation", endScreenMethod.getInstruction(0))
```
LisoUseInAIKyrios:
```suggestion
                    return-void
                    :show_end_screen_recommendation
                    nop
                """
```

LisoUseInAIKyrios:
Please confirm this works as expected when patching `20.21.37`, `20.40.45`, and `21.08.261` (or newer).

Thanks for the help!

ILoveOpenSourceApplications:
I would request @inotia00 to take a look at this, because this reworks "Hide end screen suggested video" and not "Hide end screen cards". 

Explanation here https://github.com/inotia00/ReVanced_Extended/issues/1958#issuecomment-2102519742.

utafrali:
Thanks for the detailed review!

@LisoUseInAIKyrios - I'll test this across the versions you mentioned to ensure compatibility.

@ILoveOpenSourceApplications - Good catch. I want to make sure this is targeting the correct behavior. Is this patch intended to hide the suggested video recommendations specifically, or should it be addressing the broader end screen cards functionality? Please clarify what the expected behavior should be so I can ensure the implementation is correct.

I'll also update the code with the suggested smali pattern.

inotia00:
If this PR is applied, a new side effect will occur, which can be reproduced by following these steps:

1. `Hide end screen suggested video` is enabled in Morphe settings.
2. `Autoplay next video` is enabled in YouTube settings.
3. Play any video.
4-1. If the app is in the background when the video ends, Autoplay works normally.
4-2. If the app is in the foreground when the video ends, Autoplay does not work.

This side effect should be mentioned in the settings.

ILoveOpenSourceApplications:
The current version of "Hide end screen suggested video" works flawlessly right? Then why merge this and create a flaw unless the above mentioned issue can be rectified. 

Just saying my opinion.

utafrali:
Thanks for the detailed testing and catching this side effect! You're absolutely right - introducing a regression where autoplay fails in foreground when both features are enabled isn't acceptable.

I can investigate a fix for this interaction between the two features, but given the complexity of state handling in this scenario, it might require more significant changes. Would you prefer that I:

1. Close this PR and revisit the approach to avoid the side effect, or
2. Keep it open while I work on fixing the autoplay interaction?

I appreciate you flagging this before merge - it's definitely something that needs to be addressed.

kitadai31:
@utafrali
First of all, do you even understand the difference between "End screen cards" and "End screen suggested video"?

End screen card:

Screenshot_20260226-192712_YouTube

End screen suggested video:

Image

The confusion between the two was originated in #635.
However, the author of #635 was definitely referring to the suggested videos.
Therefore, don't be misled by the issue title of #635.

FYI, "End screen" is an official name.
https://support.google.com/youtube/answer/6388789

inotia00:
Personally, I don't want to just close a PR that took your time.

This PR stems from #635.

I believe #635 was opened because the `Hide end screen suggested video` setting didn't provide sufficient explanation.

If the user dialog had provided sufficient explanation, #635 wouldn't have been opened.

I still think `Hide end screen suggested video` needs improvement, and I hope this PR can address those issues.

ILoveOpenSourceApplications:
> I believe #635 was opened because the `Hide end screen suggested video` setting didn't provide sufficient explanation.

It did explain it  sufficiently, just that "Restore old settings menus" removed `Playback` setting leaving the user with no way to access it.

> If the user dialog had provided sufficient explanation, #635 wouldn't have been opened.

#637*

> I still think `Hide end screen suggested video` needs improvement, and I hope this PR can address those issues.

It would be if improved in a way that no drawbacks are added on top of the existing method. Yes the current method maybe a bit complex but it does its job pretty well. So unless things can be improved in a positive way without racking up any negatives I still stick to what I said earlier.

utafrali:
Thanks for the clarification, @kitadai31. I appreciate you pointing out the distinction between "End screen cards" and "End screen suggested videos" - you're right, and I understand the difference now.

I also appreciate @inotia00's perspective on improving the setting explanation and @ILoveOpenSourceApplications's point about not adding drawbacks to the existing method. I'm listening to the feedback here.

If the consensus is that the current approach works well enough or that improvements should follow a different direction, I'm happy to adjust the PR accordingly or close it if it's not the right solution. What would be most helpful?

PR #650, opened 2026-02-26
feat(YouTube Music): Add `Change header` patch

### LINKED ISSUE #195: feat(YouTube Music): Add Custom Header Patch in YT Music
### Feature description

Add custom header patch for YouTube Music. So we can easily change the header by customizing >action_bar_logo, logo_music, ytm_music< 

Image

### Motivation

Where we can change custom branding icon, there custom header feature also should be added for YTM. It will look unique than official RV.

### Acknowledgements

ILoveOpenSourceApplications:
### Description

Code adapted from @inotia00 https://github.com/inotia00/revanced-patches/

Closes #195.

### Additional context

Screenshot_20260227_002851_YouTube Music

### Test results


ILoveOpenSourceApplications:
@MarcaDian, I'll be needing your guidance probably, as I'm walking in the dark here.

ILoveOpenSourceApplications:
It's starting to look promising. @MarcaDian, maybe you can try out a few custom headers? 

Personally, I don't use these patches so testers would be nice.

ILoveOpenSourceApplications:
Currently I've restricted the app to only support dark headers, cause YTM doesn't have a light mode. I need opinions on this.

MarcaDian:
Congratulations!

I didn't look at the code but it seems to work, I add this PR as source in the manager and patched Music.

The only thing that needs to be added is to restart the application after selecting the header.

ILoveOpenSourceApplications:
> Congratulations!

Thanks!.

> I didn't look at the code but it seems to work, I add this PR as source in the manager and patched Music.

Seems to be that way. What's your opinion on adding a dark and light as options? Is that needed or the current iteration fine?

Cause it'll look like this:
Screenshot_20260227_010919_YouTube Music

> The only thing that needs to be added is to restart the application after selecting the header.

Yeah, I noticed that as well. Will do that.

ILoveOpenSourceApplications:
I'm unsure regarding copyrighting `BaseChangeHeaderPatch.kt` and `ChangeHeaderPatch.kt` for YouTube after the changes and so haven't touched them. Please do the needful.

LisoUseInAIKyrios:
Will merge tomorrow if there's no new changes/discussion.

MarcaDian:
We will need to make changes to the manager to header creator, as the file name is different and does not have a header for the light theme.

LisoUseInAIKyrios:
Then this will wait to merge until manager can support it.

We should rename the patch option keys/filenames to match YouTube patch if that helps.

Are the header images the same size?

MarcaDian:
We need to discuss the general concept of the icon and header creator.

Right now they are hardcoded to use morphe values, but some third-party bundles have different file names. We can add a field to enter the file name (with preset morphe names)

All these changes add additional checks when parsing patches and their options (now we also need to check if the package is music to not show the creation of a light theme option)

MarcaDian:
It seems that the same name `morphe_header_dark` is used for YTM as for YT. 

The only thing I noticed from the YTM patch description is that the header sizes is different from YT, but during testing I didn't notice a difference, so we can use the existing xml drawables.

LisoUseInAIKyrios:
What about the template PNG images? Do they work with the existing sizes?

MarcaDian:
Do you mean custom options? We use vector xml for branding

LisoUseInAIKyrios:
I meant if a user provided custom branding header is used. It uses PNG files not drawable.

These files:
https://github.com/MorpheApp/morphe-patches/blob/main/patches/src/main/resources/change-header/drawable-xxxhdpi/morphe_header_custom_light.png

MarcaDian:
Here https://github.com/MorpheApp/morphe-patches/pull/650#issuecomment-3968675676 the first screenshot i using a png (png was for YT) I found in the trash of my phone )) But I tested this before refactoring it into a shared function

LisoUseInAIKyrios:
That looks ok. We can use the same image files and patch option description for both YouTube and YT Music.

ILoveOpenSourceApplications:
That's already the case though.

ILoveOpenSourceApplications:
The only issue from my understanding is that manager shows options for both dark and light whilst YTM only supports dark, so the light header selected is gone to waste. Isn't that right?

LisoUseInAIKyrios:
The light option would only apply to YouTube.

MarcaDian:
I'm working now on this changes in manager.

There are problem here
```
ChangeHeaderPatch: Header drawable not found: morphe_header_custom
Utils: Showing toast: ChangeHeaderPatch: Header drawable not found: morphe_header_custom
Setting: Clearing stored preference value (reset to default): morphe_header_logo
```

Patch option requires file called `morphe_header_custom_dark`

LisoUseInAIKyrios:
Push any changes to this PR as needed

MarcaDian:
Since third-party sources usually use their own names and options, I decided it would be better to show the header and icons creation button only for the default bundle. Let this be our feature.

LisoUseInAIKyrios:
I was pushing the change and it had a conflict, then saw you changed at the same time.

MarcaDian:
Oh, sorry, I changed it through the website.

LisoUseInAIKyrios:
No worries it was the same change I was going to do.

MarcaDian:
Now I'll check if everything works and post the changes for the manager.

The only thing this PR can be tested locally by temporarily changing the checking code for displaying the header creation button, since now it is displayed only for the default bundle.

Issue #204, opened 2026-01-13
feat(YouTube Music): Default filter in library


Altarentys:
### Feature description

Would like being able to set a filter category selected by default when opening the library tab.

![Image](https://github.com/user-attachments/assets/1b2fff6e-d756-4ad5-8d95-e36d225c2e3a)
-
![Image](https://github.com/user-attachments/assets/6c90e495-1afc-424c-81f6-0e276f9dd8ca)

### Motivation

When opening the Library tab, the space is cluttered with all the albums saved and artists followed, although most of the time I just wanna listen to my playlists which I have a hard time finding in all this mess.

Being able to activate the "Playlist" filter by default (or another one according to one's preference) would solve this issue.

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] This is a feature request for YouTube or YT Music

PR #653, opened 2026-02-27
feat(YouTube Music): Add `Change start page` patch

### LINKED ISSUE #56: feat(YT Music): Change start page patch
### Feature description

The YouTube mod has an option to choose what page opens when the app is loaded. I would like for this functionality to be added to YouTube music also..

### Motivation

For users who do not like using the explore options of YouTube music. Ink ow you can remove explore from the tray but the option for YouTube music to take the user direct to their playlists on open would be perfect for users as myself. 

### Acknowledgements

ILoveOpenSourceApplications:
### Description

Code adapted from @inotia00 https://github.com/inotia00/revanced-patches/

Closes #56.

### Additional context

Adds options to change the start page to the following:
```
Charts
Episodes for later
Explore
History
Library
Liked music
Playlists
Podcasts
Search
Subscriptions
``` 

### Test results

### DIFF
--- a/extensions/music/src/main/java/app/morphe/extension/music/settings/Settings.java
+++ b/extensions/music/src/main/java/app/morphe/extension/music/settings/Settings.java
@@ -4,8 +4,8 @@ import static java.lang.Boolean.FALSE;
 import static java.lang.Boolean.TRUE;
 import static app.morphe.extension.shared.settings.Setting.parent;
 
-import app.morphe.extension.music.patches.ChangeHeaderPatch;
-import app.morphe.extension.music.patches.ChangeStartPagePatch;
+import app.morphe.extension.music.patches.ChangeHeaderPatch.HeaderLogo;
+import app.morphe.extension.music.patches.ChangeStartPagePatch.StartPage;
 import app.morphe.extension.shared.Logger;
 import app.morphe.extension.shared.settings.BooleanSetting;
 import app.morphe.extension.shared.settings.EnumSetting;
@@ -19,7 +19,7 @@ public class Settings extends SharedYouTubeSettings {
     public static final BooleanSetting HIDE_GET_PREMIUM_LABEL = new BooleanSetting("morphe_music_hide_get_premium_label", TRUE, true);
 
     // General
-    public static final EnumSetting<ChangeStartPagePatch.StartPage> CHANGE_START_PAGE = new EnumSetting<>("morphe_change_start_page", ChangeStartPagePatch.StartPage.DEFAULT, true);
+    public static final EnumSetting<StartPage> CHANGE_START_PAGE = new EnumSetting<>("morphe_change_start_page", StartPage.DEFAULT, true);
     public static final BooleanSetting HIDE_CAST_BUTTON = new BooleanSetting("morphe_music_hide_cast_button", TRUE, true);
     public static final BooleanSetting HIDE_CATEGORY_BAR = new BooleanSetting("morphe_music_hide_category_bar", FALSE, true);
     public static final BooleanSetting HIDE_HISTORY_BUTTON = new BooleanSetting("morphe_music_hide_history_button", FALSE, true);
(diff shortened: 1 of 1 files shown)


patches/src/main/resources/addresources/values/music/arrays.xml
```diff
@@ -29,6 +29,34 @@
...
+        <item>CHARTS</item>
+        <item>EPISODES_FOR_LATER</item>
+        <item>EXPLORE</item>
+        <item>HISTORY</item>
+        <item>LIBRARY</item>
+        <item>LIKED_MUSIC</item>
+        <item>PODCASTS</item>
+        <item>SAMPLES</item>
+        <item>SEARCH</item>
+        <item>SUBSCRIPTIONS</item>
```
LisoUseInAIKyrios, line 57: Try a VPN to an obscure country and verify all of these do not show an error message.

For YouTube, some of the possible start pages are not available in some regions. The app usually handles the not available page correctly and shows the home screen instead, but some start pages show an error  if the page isn't available in the the country (those start pages are not a selectable option in YT Morphe)

ILoveOpenSourceApplications: I know that `Podcasts` isn't available in Russia.

LisoUseInAIKyrios: Does the app show an error, or does it show the regular home page?

If no error is shown that's ok. It's just the error pages we need to worry about.

ILoveOpenSourceApplications: Recommend me a good VPN which probably doesn't ask me to sign up and have to pay to connect to a Russian server.

ILoveOpenSourceApplications:
Every page except Search opens up as start page now. Will need to look into that.

Samples for some reason doesn't start playing and keeps on buffering, when switching to Home and back, it starts working. Need to check this as well.

ILoveOpenSourceApplications:
This can now close #204 as "Playlists" can be opened as the default start page.

ILoveOpenSourceApplications:
I can't get samples to open without some sort of flaws. With browseId it'll keep loading indefinetly and with URL, it'll open the home page for a split second before open samples page with a snackbar at the bottom saying "Signed in as Yahoo (iammario@gmail.com)". Trying to combine both crashed the app immediately. 

So what should I do here? Should I remove "Samples" as an option?

LisoUseInAIKyrios:
If it has issues then remove the option. Can maybe figure out something later

patches/src/main/resources/addresources/values/music/arrays.xml
LisoUseInAIKyrios: UrbanVPN

ILoveOpenSourceApplications: This one https://www.urban-vpn.com/?

Edit: That one doesn't have a free Russian server and wanted me to sign up as well. I hate that.

ILoveOpenSourceApplications:
Haven't linked #56 by the way.

patches/src/main/resources/addresources/values/music/arrays.xml
ILoveOpenSourceApplications: It was using just the Russian language and also that was for YouTube and not YouTube Music.

ILoveOpenSourceApplications:
Hmm, except "Search" the rest doesn't work on v7.29.52. Everything works on v8.10.52 and above.

LisoUseInAIKyrios:
Can use different preference values based on the app it's patching.

See spoof app version and how it uses different preference entries and entryValues:
```
  if (is_20_40_or_greater) {
      ListPreference("morphe_spoof_app_version_target")
  } else if (is_20_31_or_greater) {
      ListPreference(
          key = "morphe_spoof_app_version_target",
          entriesKey = "morphe_spoof_app_version_target_legacy_20_31_entries",
          entryValuesKey = "morphe_spoof_app_version_target_legacy_20_31_entry_values"
      )
  } else if (is_20_14_or_greater) {
      ListPreference(
          key = "morphe_spoof_app_version_target",
          entriesKey = "morphe_spoof_app_version_target_legacy_20_14_entries",
          entryValuesKey = "morphe_spoof_app_version_target_legacy_20_14_entry_values"
      )
  }
```

LisoUseInAIKyrios:
Oh, only one start page works on 7.29?

Then maybe restrict the patch entirely. If trying to patch below 8.10 then print a logger message that it doesn't work and gracefully return from patch execute() method.

ILoveOpenSourceApplications:
No need, it's fixed. All supported versions have been tested for now.

LisoUseInAIKyrios:
@ILoveOpenSourceApplications Ready to merge?

ILoveOpenSourceApplications:
Yes.

PR #664, opened 2026-02-28
chore: Use latest patches gradle plugin

brosssh:
### Description

This PR adds a new field in the generated `MANIFEST.MF` file in the bundle, the compatible patcher version. This is particular useful for external tools like [mine](https://github.com/brosssh/revanced-external-bundles) or URV, which needs to be aware of the patcher version used.  

Depends on https://github.com/MorpheApp/morphe-patches-gradle-plugin/pull/12

### Additional context

N/A

### Test results

image

PS, I don't know why there is a new line in the middle of the `License`, weird

### DIFF
--- a/patches/build.gradle.kts
+++ b/patches/build.gradle.kts
@@ -9,7 +9,6 @@ patches {
         contact = "na"
         website = "https://morphe.software"
         license = "GNU General Public License v3.0, with additional GPL section 7 requirements"
-        patcherVersion = libs.versions.morphe.patcher.get()
     }
 }
 


wchill:
How will you be handling other patch bundles that haven't updated the gradle plugin version or don't specify the patcher version? In the gradle plugin PR, that parameter is set to nullable.

brosssh:
I don't know how to handle bundles that doesn't specify patcher version, probably with some clunky workaround I still need to figure out.
Although it's marked as nullable, it clearly isn't as an exception is raised if it's not specified 🤷‍♂️. Same apply for the other fields I guess.

LisoUseInAIKyrios:
https://docs.oracle.com/javase/8/docs/api/java/util/jar/Attributes.html#getValue-java.lang.String-

?

```
    JarFile(jarPath).use { jar ->
        val manifest = jar.manifest ?: return null
        val attributes = manifest.mainAttributes
        return attributes.getValue("Patcher-Version") // Returns null if not found
    }
```

brosssh:
Try setting website to null and see what happens. It's probably how it's being implemented in the plugin.

brosssh:
Anyway, if the patcher version is not found I might guess a patcher version. If the guess it's incorrect, either fallback to some other patcher version or give an error.

LisoUseInAIKyrios:
If patcher version is not found then fallback to 1.1.1 because all known third party Morphe patches repos are using that.

brosssh:
For Morphe bundle yeah, that's easy. Gets harder on other bundles, don't know how I could guess the patcher used, but that's out of the scope of this PR.

brosssh:
Personally I don't even care adding support to _some_ other bundles, so not a big deal to me. Though, URV would need a way to figure it out, if it doesn't somehow already.

LisoUseInAIKyrios:
Patcher will update to 1.2.0 stable soon when arsclib PR's are merged.

This PR isn't needed anymore?

LisoUseInAIKyrios:
Actually, we can bump gradle plugin now

Issue #87, opened 2026-01-04
feat(YouTube): Hide "Related Videos/Recommended Videos" under current watching video


AkshayKanathur:
### Feature description

Please add a feature to hide Related / Recommended videos under current watching video 

### Motivation

Helpful to remove distraction. 

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] This is a feature request for YouTube or YT Music

LisoUseInAIKyrios:
Slightly different, but accomplishes the same thing:
`Player > Open videos in fullscreen portrait`

Can still use actions buttons, view comments, etc.

ghost:
These related videos under the player, and the posts that weave between them, are the only thing keeping Morphe from being a completely distraction-free app.

I know it's not easy, but any requests for this on past repos were ignored or dismissed, I'd love to finally see this become a reality (a reality *again*, since it was a YouTube update that broke it previously). Then I can finally stop using third party apps like PipePipe-X.

AngryHands:
It would be also useful to be able to remove the default "all" filter from the related frame. That could make the related videos useful as it could eliminate the need to return to the search results.

<details>

<img width="1080" height="2214" alt="Image" src="https://github.com/user-attachments/assets/a1f0fca8-0380-4fb5-8682-ce8c605e7e3a" />
</details>

AngryHands:
> Slightly different, but accomplishes the same thing: `Player > Open videos in fullscreen portrait`
> 
> Can still use actions buttons, view comments, etc.

Unfortunately, once I close the fullscreen video it is not possible to open the fullscreen video in portrait mode. The fullscreen video turns horizontal even while being viewed vertically and there is no way to rotate the video while being in fullscreen mode. The annoying related frame is visible one being exited out of the fullscreen mode to view comments etc.

AngryHands:
Being able to collapse the related frame would be very useful too (like hiding the preview comment of comments bar while having the access to the comments). I use this feature on desktop youtube site provided by open source for YouTube extension.

PR #685, opened 2026-03-02
feat(YouTube): Add `Hide releated videos` patch

inotia00:
### Description

Closes https://github.com/MorpheApp/morphe-patches/issues/87.

### Additional context

This PR's technical approach has several advantages over the RVX implementation (limiting the number of elements that can be drawn on screen) or using a custom filter (`video_lockup$relatedH`):

> [!Note]
> 1. The patch uses a proto parser to parse innerTube responses, which is the same technical approach YouTube uses to parse innerTube responses.
> 2. So there is no issue with related videos being hidden, or unintended elements being hidden.
> 3. The patch not only hides related videos, but also removes the ContinuationToken field, which prevents unnecessary spam API requests.
> 4. The proto parser's high performance minimizes client-side overhead, and since there are no unnecessary spam API requests, there's no server-side overhead.

_FYI: If related videos are hidden using a custom filter, the ContinuationToken field remains valid, resulting in numerous unnecessary spam API requests in a short period of time. This can lead to IP blocks by Google servers. For this reason, patches must be implemented using a proto parser, like this PR._

### Test results

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/settings/Settings.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/settings/Settings.java
@@ -188,8 +188,8 @@ public class Settings extends SharedYouTubeSettings {
     public static final BooleanSetting HIDE_PLAYER_PREVIOUS_NEXT_BUTTONS = new BooleanSetting("morphe_hide_player_previous_next_buttons", FALSE, true);
     public static final BooleanSetting HIDE_QUICK_ACTIONS = new BooleanSetting("morphe_hide_quick_actions", FALSE);
     public static final BooleanSetting HIDE_QUICK_ACTIONS_RELATED_VIDEOS = new BooleanSetting("morphe_hide_quick_actions_related_videos", FALSE, parentNot(HIDE_QUICK_ACTIONS));
-    public static final BooleanSetting HIDE_RELATED_VIDEOS = new BooleanSetting("morphe_hide_related_videos", FALSE, true);
-    public static final BooleanSetting HIDE_RELATED_VIDEOS_OVERLAY = new BooleanSetting("morphe_hide_related_videos_overlay", FALSE, true);
+    public static final BooleanSetting HIDE_PLAYER_RELATED_VIDEOS = new BooleanSetting("morphe_hide_player_related_videos", FALSE, true);
+    public static final BooleanSetting HIDE_PLAYER_RELATED_VIDEOS_OVERLAY = new BooleanSetting("morphe_hide_player_related_videos_overlay", FALSE, true);
     public static final BooleanSetting HIDE_SUBSCRIBERS_COMMUNITY_GUIDELINES = new BooleanSetting("morphe_hide_subscribers_community_guidelines", TRUE);
     public static final BooleanSetting HIDE_TIMED_REACTIONS = new BooleanSetting("morphe_hide_timed_reactions", TRUE);
     public static final BooleanSetting HIDE_VIDEO_TITLE = new BooleanSetting("morphe_hide_video_title", FALSE);
--- a/patches/src/main/kotlin/app/morphe/patches/youtube/layout/hide/relatedvideos/HideRelatedVideosPatch.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/youtube/layout/hide/relatedvideos/HideRelatedVideosPatch.kt
@@ -48,7 +48,7 @@ val hideRelatedVideosPatch = bytecodePatch(
 
     execute {
         PreferenceScreen.PLAYER.addPreferences(
-            SwitchPreference("morphe_hide_related_videos"),
+            SwitchPreference("morphe_hide_player_related_videos")
         )
 
         val continuationsField = with (WatchNextResponseParserFingerprint) {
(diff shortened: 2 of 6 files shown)


extensions/youtube/src/main/java/app/morphe/extension/youtube/settings/Settings.java
```diff
@@ -188,8 +188,9 @@ public class Settings extends SharedYouTubeSettings {
     public static final BooleanSetting HIDE_PLAYER_CONTROL_BUTTONS_BACKGROUND = new BooleanSetting("morphe_hide_player_control_buttons_background", FALSE, true);
     public static final BooleanSetting HIDE_PLAYER_PREVIOUS_NEXT_BUTTONS = new BooleanSetting("morphe_hide_player_previous_next_buttons", FALSE, true);
     public static final BooleanSetting HIDE_QUICK_ACTIONS = new BooleanSetting("morphe_hide_quick_actions", FALSE);
+    public static final BooleanSetting HIDE_QUICK_ACTIONS_RELATED_VIDEOS = new BooleanSetting("morphe_hide_quick_actions_related_videos", FALSE, parentNot(HIDE_QUICK_ACTIONS));
+    public static final BooleanSetting HIDE_RELATED_VIDEOS = new BooleanSetting("morphe_hide_related_videos", FALSE, true);
     public static final BooleanSetting HIDE_RELATED_VIDEOS_OVERLAY = new BooleanSetting("morphe_hide_related_videos_overlay", FALSE, true);
-    public static final BooleanSetting HIDE_RELATED_VIDEOS = new BooleanSetting("morphe_hide_related_videos", FALSE);
```
inotia00, line 191: The key `morphe_hide_related_videos` is already being used by another setting.

To minimize migration issues, it would be better to change the `morphe_hide_related_videos` key to something else.

patches/src/main/kotlin/app/morphe/patches/youtube/layout/hide/relatedvideos/HideRelatedVideosPatch.kt
```diff
@@ -0,0 +1,246 @@
...
+                    "V",
+                    AccessFlags.PRIVATE.value or AccessFlags.FINAL.value,
+                    annotations,
+                    null,
+                    MutableMethodImplementation(6),
+                ).toMutable().apply {
+                    addInstructionsWithLabels(
+                        0,
+                        """
+                            invoke-static { }, $EXTENSION_CLASS_DESCRIPTOR->hideRelatedVideos()Z
```
wchill, line 164: This is a long smali block, why can't this be done in Java?

inotia00: Among the subfields, there's an ElementRenderer field.

It appears the YouTube app calls the native library to parse the ElementRenderer field.

Therefore, rebuilding the parent field containing the ElementRenderer using a proto builder in the extension causes problems at the point where the native library is called (it's always replaced with the default instance).

Therefore, instead of using the proto builder, I chose to modify the YouTube field directly.

The subfields of ElementRenderer can be modified in the extension (another upcoming PR uses this approach).

> [!Note]
> 
> The InnerTube response received in proto format, when converted to a json file, is extremely large, exceeding 200MB.
> 
> Therefore, to reduce overhead, the YouTube app calls a native library.
>
> I haven't analyzed the native library, but it appears to directly reference the memory address instead of converting it to an object.

inotia00: On second thought, building the modified MessageLite class using proto builder failed, but if Java Reflection is used, the smali code can be reduced.

LisoUseInAIKyrios: Unless using reflections allows reducing smali to very little, I think it's better to keep it all in smali.

wchill: Another approach is you could always redeclare the necessary fields as public, then provide wrapper/stub code to set these in Java

LisoUseInAIKyrios: We can merge this now. It's working, and it may be unlikely we will change the smali logic soon. Can always refactor it later if desired.

extensions/youtube/src/main/java/app/morphe/extension/youtube/settings/Settings.java
LisoUseInAIKyrios: Can this new key be named `morphe_hide_player_related_videos`?

But any other name would work.

LisoUseInAIKyrios: Renamed to `morphe_hide_player_related_videos` and `morphe_hide_player_related_videos`

LisoUseInAIKyrios:
Will merge tomorrow if there's no new changes or concerns.

PR #722, opened 2026-03-05
feat(Spoof video streams): Add "Force player JavaScript hash" and "Player JavaScript hash"

inotia00:
### Description

Added `Force player JavaScript hash` and `Player JavaScript hash` settings.

### TL;DR

- This PR is a temporary workaround for the `StreamingDataRequest: getStream failure` exception.

- Once this PR is merged, users will be able to force the old player js hash, such as **9f4cc5e4**.

```
9f4cc5e4
```

> [!Note]
> 1. Change the default client to `TV`.
> 2. Enable the `Force player JavaScript hash` setting.
> 3. Click the `Player JavaScript hash` setting.
> 4. Delete the existing value, enter **9f4cc5e4**, and save.
> 5. Restart the app and the `TV` client will work again.

> [!Caution]
> This is a temporary workaround and may no longer be valid after one month. yt-dlp-ejs must be updated within a month.

### Additional Context

- yt-dlp-ejs is used to deobfuscate the `TV` client in Morphe.

- Recently, Google made Player JavaScript more difficult to parse. This caused yt-dlp-ejs, which had been working fine for six months, to fail deobfuscation for the first time: https://github.com/yt-dlp/ejs/issues/51.

- This caused Morphe to throw a `StreamingDataRequest: getStream failure` exception, and playback failed on `TV`.

player_js_version_readme

- As a temporary workaround, yt-dlp enforced the old player js hash (**9f4cc5e4**): https://github.com/yt-dlp/yt-dlp/pull/16123.

- This is only a temporary workaround, as GVS will revoke the player js hash after about a month.

- The yt-dlp team will likely update yt-dlp-ejs within a month, at which point Morphe will also use the latest yt-dlp-ejs.

### Test results

### DIFF
--- a/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/settings/preference/ForceJSHashResettableEditTextPreference.java
+++ b/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/settings/preference/ForceJSHashResettableEditTextPreference.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright 2026 Morphe.
+ * https://github.com/MorpheApp/morphe-patches
+ */
+
 package app.morphe.extension.shared.settings.preference;
 
 import android.content.Context;
--- a/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/settings/preference/ForceJSHashSwitchPreference.java
+++ b/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/settings/preference/ForceJSHashSwitchPreference.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright 2026 Morphe.
+ * https://github.com/MorpheApp/morphe-patches
+ */
+
 package app.morphe.extension.shared.settings.preference;
 
 import android.content.Context;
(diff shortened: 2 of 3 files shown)


wchill:
Can we set the default value of the preference to `9f4cc5e4` instead of empty string? We can always change it back once we update yt-dlp-ejs.

Checkbox can remain disabled by default, but if we're asking people to enable this setting to fix player issues, might as well preset the player hash for now.

inotia00:
I didn't do this for two reasons:

1. The `morphe_spoof_video_streams_js_hash` has been around since previous versions. Therefore, to avoid migration issues, we need to change the key, for example, to something like `morphe_spoof_video_streams_player_js_hash`.

2. I want users to understand that this is a temporary workaround and may not work after a month. That's why I made them follow these cumbersome steps.

Figim:
Hopefully ytdlp can solve this problem

inotia00:
There are two more fundamental workarounds:

1. Implement logic to more precisely analyze player.js through the AST, as YouTube.js does.

2. Fetch the deobfuscation function name from the server, as Grayjay did (Remote cipher).

Neither method is available right now (Not implemented in Java).

Issue #710, opened 2026-03-04
bug(YouTube): "Hide AI comments summary" also hides the top bar in the comments


wdjasasjsajafsasjf:
### Bug description

As explained in the title, enabling the option causes the top bar for sorting comments by "best, newest, timed" etc. disappears when enabling the option.

YouTube 20.40.45
Patches version 1.17.0-dev.1

<details>

![Image](https://github.com/user-attachments/assets/b8b61569-2eb1-47f4-b620-c7de699f837e)
</details>

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

LisoUseInAIKyrios:
A buffer filter might need an update after the recent refactoring.

PR #731, opened 2026-03-06
fix(YouTube - Hide layout components): Replace "Hide AI comments summary" with "Sanitize category bar"

inotia00:
### Description

Closes https://github.com/MorpheApp/morphe-patches/issues/710.

### Hide AI comments summary

- Hides only the `Topics` category button in the comment category bar.

- The category bar may be hidden in YouTube 20.22+ (A/B testing?): https://github.com/MorpheApp/morphe-patches/issues/710.

### Sanitize category bar

- Hides all category buttons except `Top` and `Newest` in the comment category bar (Category buttons like `Members` are also hidden).

- Not affected by A/B testing.

### Test results

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/CommentsFilter.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/CommentsFilter.java
@@ -1,11 +1,25 @@
+/*
+ * Copyright 2026 Morphe.
+ * https://github.com/MorpheApp/morphe-patches
+ *
+ * Original hard forked code:
+ * https://github.com/ReVanced/revanced-patches/commit/724e6d61b2ecd868c1a9a37d465a688e83a74799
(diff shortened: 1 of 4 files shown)


ILoveOpenSourceApplications:
Instead of the categories except `Top` and `Newest` getting hidden, can it be done in a way that the user can opt for which all to be hidden? 

Maybe something like the feed flyout menu filter will also be fine. Because there are users who will be using `Timed` and that'll also be hidden here with this toggle, right?

I'm just throwing out my own thinking here, if it's not possible then all is good.

inotia00:
This could theoretically be implemented if a proto parser is available.

Mapping .proto files takes more time than preparing other patches, so it's not a priority.

LisoUseInAIKyrios (approved):

Hmdmu98:
We need add "show play all button" in overly buttons like rvx

Issue #686, opened 2026-03-02
bug(YouTube): "Remove background playback restrictions" causes app crash on startup


ShoyebOP:
### Bug description

YouTube version: 20.40.45

Morphe manager verison- 1.9.0 & 1.10.0-dev.1

Patch version: 1.15.0

Open any YouTube videos causes the crash.

### Error logs

```shell
03-02 14:43:10.869  8193  8292 E AndroidRuntime: FATAL EXCEPTION: glide-active-resources
03-02 14:43:10.869  8193  8292 E AndroidRuntime: Process: com.google.android.youtube, PID: 8193
03-02 14:43:10.869  8193  8292 E AndroidRuntime: java.lang.IllegalStateException: Cannot obtain size for recycled Bitmap: android.graphics.Bitmap@1d1778d[640x480] ARGB_8888
03-02 14:43:10.869  8193  8292 E AndroidRuntime:       at fxe.a(PG:82)
03-02 14:43:10.869  8193  8292 E AndroidRuntime:       at frm.a(PG:3)
03-02 14:43:10.869  8193  8292 E AndroidRuntime:       at fnq.a(PG:3)
03-02 14:43:10.869  8193  8292 E AndroidRuntime:       at fpa.a(PG:7)
03-02 14:43:10.869  8193  8292 E AndroidRuntime:       at fxa.h(PG:2)
03-02 14:43:10.869  8193  8292 E AndroidRuntime:       at fpa.d(PG:1)
03-02 14:43:10.869  8193  8292 E AndroidRuntime:       at pev.m(PG:14)
03-02 14:43:10.869  8193  8292 E AndroidRuntime:       at fmt.c(PG:34)
03-02 14:43:10.869  8193  8292 E AndroidRuntime:       at bo.run(PG:569)
03-02 14:43:10.869  8193  8292 E AndroidRuntime:       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1156)
03-02 14:43:10.869  8193  8292 E AndroidRuntime:       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:651)
03-02 14:43:10.869  8193  8292 E AndroidRuntime:       at bo.run(PG:18)
03-02 14:43:10.869  8193  8292 E AndroidRuntime:       at java.lang.Thread.run(Thread.java:1119)


Another one-

[...]

PR #732, opened 2026-03-06
fix(Custom branding): Safely clone bitmaps to prevent native recycled crash in Glide

Pupy200mine:
### Description

Fixes a native `Cannot obtain size for recycled Bitmap` crash in Glide caused by [MediaSession](cci:2://file:///home/pupy200mine/Descargas/morphe-patches/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/youtube/patches/MediaSessionBitmapMitigationPatch.java:7:0-37:1) when trying to watch a video.

It uses a bytecode transformation to intercept `MediaMetadata.Builder.putBitmap` calls, safely deep-cloning the thumbnail Bitmap before YouTube's memory collector recycles it (or supplying a dummy 1x1 image if already recycled).

* Added an optional patch: "Fix notification thumbnail crash".
* The Java mitigation classes are placed in `shared-youtube/library` to solve the crash on both YouTube and YT Music.

### Additional context

It seems to be a bug that only affects a minority, since I couldn't find anyone with the same problem (I have a Pixel 6 + root + Evolution X ROM, a weird setup would make sense.) and prevents any video playback usage due to an instant app crash when opening a video or music on YouTube. It prevented me from using Morphe for several weeks.

### Test results

### DIFF
--- a/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/patches/CustomBrandingPatch.java
+++ b/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/patches/CustomBrandingPatch.java
@@ -1,6 +1,13 @@
+/*
+ * Copyright 2026 Morphe.
+ * https://github.com/MorpheApp/morphe-patches
+ *
+ * Original hard forked code:
+ * https://github.com/ReVanced/revanced-patches/commit/724e6d61b2ecd868c1a9a37d465a688e83a74799
+ */
+
 package app.morphe.extension.shared.patches;
 
-import android.app.Notification;
 import android.content.ComponentName;
 import android.content.Context;
 import android.content.pm.PackageManager;
@@ -100,16 +107,32 @@ public class CustomBrandingPatch {
     /**
      * Injection point.
      */
-    public static void setNotificationIcon(Notification.Builder builder) {
+    public static int getSmallIcon(int original) {
         try {
             final int smallIcon = getNotificationSmallIcon();
             if (smallIcon != 0) {
-                builder.setSmallIcon(smallIcon)
-                        .setColor(Color.TRANSPARENT); // Remove YT red tint.
+                return smallIcon;
             }
         } catch (Exception ex) {
-            Logger.printException(() -> "setNotificationIcon failure", ex);
+            Logger.printException(() -> "getSmallIcon failure", ex);
         }
+        return original;
+    }
+
+    /**
+     * Injection point.
+     */
+    public static int getColor(int original) {
+        try {
+            final int smallIcon = getNotificationSmallIcon();
+            if (smallIcon != 0) {
+                // Remove YT red tint.
+                return Color.TRANSPARENT;
+            }
+        } catch (Exception ex) {
+            Logger.printException(() -> "getColor failure", ex);
+        }
+        return original;
     }
 
     /**
--- a/patches/src/main/kotlin/app/morphe/patches/shared/layout/branding/BaseCustomBrandingPatch.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/shared/layout/branding/BaseCustomBrandingPatch.kt
@@ -2,6 +2,7 @@ package app.morphe.patches.shared.layout.branding
 
 import app.morphe.patcher.Fingerprint
 import app.morphe.patcher.extensions.InstructionExtensions.addInstruction
+import app.morphe.patcher.extensions.InstructionExtensions.addInstructions
 import app.morphe.patcher.extensions.InstructionExtensions.getInstruction
 import app.morphe.patcher.patch.PatchException
 import app.morphe.patcher.patch.ResourcePatch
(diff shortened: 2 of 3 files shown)


inotia00:
Thank you for your contribution.

#686 not reproducible on our end, so we can't confirm whether this PR fully resolves the issue.

Is #686 being reproduced on your side, and does this PR fix it?

Pupy200mine:
Removing the patch solves the problem, but as a result, it completely removes the notification thumbnail, both in the notifications menu and in the lock menu.

LisoUseInAIKyrios:
@Pupy200mine and this PR is confirmed to fix the problem for you?

patches/src/main/kotlin/app/morphe/patches/shared/misc/fix/bitmap/FixMediaSessionBitmapPatch.kt
```diff
@@ -0,0 +1,52 @@
...
+    override val methodName: String,
+    override val methodParams: Array<String>,
+    override val returnType: String,
+) : IMethodCall {
+    PutBitmapCompat(
+        "Landroid/support/v4/media/MediaMetadataCompat\$Builder;",
+        "putBitmap",
+        arrayOf("Ljava/lang/String;", "Landroid/graphics/Bitmap;"),
+        "Landroid/support/v4/media/MediaMetadataCompat\$Builder;",
+    ),
```
inotia00: YouTube 20.21.37, 20.40.45, 21.08.265, YouTube Music 7.29.52, 8.46.57, 9.09.52 all don't have this method call.

This doesn't seem necessary.

extensions/shared-youtube/library/src/main/java/android/support/v4/media/MediaMetadataCompat.java
```diff
@@ -0,0 +1,8 @@
+package android.support.v4.media;
+import android.graphics.Bitmap;
+
+public class MediaMetadataCompat {
```
inotia00: Since it is a class in YouTube and YouTube Music, it should be moved to a subdirectory of `stub`.

extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/youtube/patches/MediaSessionBitmapMitigationPatch.java
```diff
@@ -0,0 +1,38 @@
...
+        if (original == null) return null;
+        if (original.isRecycled()) {
+            Logger.printDebug(() -> "MediaSessionBitmapMitigation: Intercepted recycled bitmap");
+            return Bitmap.createBitmap(1, 1, Bitmap.Config.ARGB_8888);
+        }
+        try {
+            Bitmap.Config config = original.getConfig();
+            return original.copy(config != null ? config : Bitmap.Config.ARGB_8888, false);
+        } catch (Exception e) {
+            Logger.printDebug(() -> "MediaSessionBitmapMitigation: Could not copy bitmap: " + e.getMessage());
```
inotia00: Can be replaced with `Logger.printException()`

LisoUseInAIKyrios:
@Pupy200mine Please enable "allow changes by maintainers".

The destination branch must be morphe-patches `dev`


Edit: nevermind GitHub is showing me a new UI and the edit button is now half hidden

inotia00:
@Pupy200mine I did some refactoring.

Could you test if the patch I refactored works on your device?

https://github.com/MorpheApp/morphe-patches/actions/runs/22754957661/artifacts/5793996603

Pupy200mine:
After using the patches I can confirm that everything is working correctly.

patches/src/main/kotlin/app/morphe/patches/music/misc/backgroundplayback/BackgroundPlaybackPatch.kt
```diff
@@ -12,7 +13,8 @@ val backgroundPlaybackPatch = bytecodePatch(
 ) {
     dependsOn(
         sharedExtensionPatch,
-        settingsPatch
+        settingsPatch,
+        fixRecycledBitmapPatch,
```
LisoUseInAIKyrios, line 17: @inotia00 Is this an issue with background playback, or custom branding?

Should this fix be a dependent patch here or custom branding?

inotia00: I think custom branding is the cause of the issue.

LisoUseInAIKyrios: Ok I see this is already a dependent patch for the base custom branding.

LisoUseInAIKyrios: Will merge this later today

Issue #102, opened 2026-01-05
bug(YouTube): Cannot hide player action buttons "Report", "Thanks", "Ask AI", "Hype", "Stop Ads", etc 


Janekknaf:
### Bug description

It used to be avaible in RVX by inotia00 
<img width="1280" height="2856" alt="Image" src="https://github.com/user-attachments/assets/29627e08-735c-4946-92df-589f35685e5b" />

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

ILoveOpenSourceApplications:
`Adds options to hide action buttons (such as the Download button) under videos. Patching version 20.21.37 or lower can hide more player button types.`

LisoUseInAIKyrios:
Currently there is no known way to hide those last few buttons.

It took a lot of work/effort just to accomplish the hiding that is currently present, because YT 20.31 dramatically changed it's internal code and many new fixes were required.

If you want to hide them, then patch `20.21.37`

LisoUseInAIKyrios:
Re-opening for others to find. And if this can ever be done in the future.

Issue #280, opened 2026-01-24
feat(YouTube): Expand buttons when others are hidden


ILoveOpenSourceApplications:
### Feature description

Currently for action buttons, when one button is hidden, there's a gap left behind. Similarly, in channel pages having only two buttons, if any button is hidden, the there's a gap where the button is hidden.

But there's an exception. If a channel page has three buttons, and two of the three buttons are hidden, except the Subscribe button that is, then the Subscribe button expands to take up the entire space leaving only a slight gap behind.

| None of the buttons hidden | Subscribe button not hidden | Join button (or any other button) not hidden |
|--------|--------|--------|
| ![Image](https://github.com/user-attachments/assets/8db2d06b-cffa-4fb2-8137-a3320c6efc0b) | ![Image](https://github.com/user-attachments/assets/7048dd27-e170-48e0-af0a-31bf207a7730) | ![Image](https://github.com/user-attachments/assets/1dc4c1af-92d3-42e8-b881-f477016ee067) | 

### Motivation

If the logic behind why this happens can be identified and applied to the case of action buttons or channel page with two buttons, then that's gonna solve one of the long lasting issues of gaps when buttons are hidden. Though I don't know how useful this will be in the case of video action buttons.

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] This is a feature request for YouTube or YT Music

LinuxGuy-cyber:
Still having this issue.

<details><summary>Details</summary>
<p>

<img width="1080" height="507" alt="Image" src="https://github.com/user-attachments/assets/8ae208f4-db0b-4e90-bd2e-313ae6f2d1b2" />
<img width="1080" height="469" alt="Image" src="https://github.com/user-attachments/assets/d7795de9-7872-45ca-87d5-4ff89cb0e51f" />
<img width="1080" height="518" alt="Image" src="https://github.com/user-attachments/assets/2f4ecba1-70f9-49f2-8895-90005089a2b5" />

</p>
</details>

PR #736, opened 2026-03-06
fix(YouTube - Hide video action buttons): Some action buttons cannot be hidden in YouTube 20.22+

inotia00:
### Hiding video action buttons like Ask, Hype, Report, and Thanks is now supported in YouTube 20.22+

- Starting with YouTube 20.22, the way apps read proto buffers has changed.

- For this reason, hiding some action buttons has not been available in YouTube 20.22+ until now.

- As a workaround, I spent about a week writing a proto file to parse the `/next` endpoint response, and the patch is now supported regardless of YouTube version.

- Closes https://github.com/MorpheApp/morphe-patches/issues/102.

#### Before
before_pr

#### After
after_pr

- While [the issue was opened in the RVX issue repo](https://github.com/inotia00/ReVanced_Extended/issues/17) in November 2022, the `Hide action buttons` patch was first implemented in Vanced Extended around June 2022 (with help from @0xrxl), and the issue has been around since then.

- Personally, I'm satisfied that the proto parser has solved an issue that I couldn't fix for about 4 years.

### Test results

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/VideoActionButtonsFilter.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/VideoActionButtonsFilter.java
@@ -1,29 +1,142 @@
+/*
+ * Copyright 2026 Morphe.
+ * https://github.com/MorpheApp/morphe-patches
+ *
+ * Original hard forked code:
+ * https://github.com/ReVanced/revanced-patches/commit/724e6d61b2ecd868c1a9a37d465a688e83a74799
+ */
+
 package app.morphe.extension.youtube.patches.components;
 
-import app.morphe.extension.youtube.patches.VersionCheckPatch;
+import app.morphe.extension.youtube.patches.VideoInformation;
 import app.morphe.extension.youtube.settings.Settings;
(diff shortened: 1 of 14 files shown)


ILoveOpenSourceApplications:
Seems this also will fix #280?

inotia00:
Channel pages use the `/browse` endpoint instead of the `/next` endpoint, so a completely new mapping is needed.

0xrxL:
If you want, you can also add what I call: "Lite Mode", with the only first three action buttons visible and the save button placed directly in the player.

Screenshot_20260306_152918

Hmdmu98:
can hide buttons label?

ANewDawn:
Could you add the subscribe button to the list of actions to hide. 

I removed the subscribe button with custom filter "subscribe_button" but seems it doesn't move empty space.

imvoldemort:
@0xrxL how did you get the new player overlay buttons?

0xrxL:
> @0xrxL how did you get the new player overlay buttons?

I wrote a logic that takes all the properties (including fade) from the fullscreen button and applies them (with an Observer) to a new button, placed at a different X position. @LisoUseInAIKyrios said that he will add it in future.

PR #762, opened 2026-03-07
feat(Reddit): Add experimental support for `2026.10.0`

LisoUseInAIKyrios:
### Description

Adds experimental support for latest Reddit.

See: 
https://github.com/MorpheApp/morphe-documentation/blob/main/docs/morphe-resources/questions.md#20-can-i-patch-the-latest-versions-of-apps-can-i-patch-newer-than-the-recommended-version

Tested with 2026.04.0, 2026.10.0, and verified to patch older app targets (did not try using older than 2026.04.0).

Did not verify "Remove subreddit dialog" works but the patch changes look correct.

### Test results

### DIFF
--- a/patches/src/main/kotlin/app/morphe/patches/reddit/layout/subredditdialog/Fingerprints.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/reddit/layout/subredditdialog/Fingerprints.kt
@@ -13,17 +13,6 @@ import app.morphe.patcher.string
 import com.android.tools.smali.dexlib2.AccessFlags
 import com.android.tools.smali.dexlib2.Opcode
 
-private val SHOW_DIALOG_FILTERS = listOf(
-    methodCall(
-        opcode = Opcode.INVOKE_VIRTUAL,
-        name = "show"
-    ),
-    opcode(
-        Opcode.MOVE_RESULT_OBJECT,
-        location = MatchAfterImmediately()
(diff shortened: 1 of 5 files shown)


optimatrix0:
In my view, losing support for very old versions wouldn’t really be a big deal. For example, something like 2024.17.0 is probably a version that almost nobody actually wants to keep using anymore.

So honestly, if dropping some legacy compatibility helps improve newer versions, I personally don’t see a problem with that. The number of users on those versions is probably very small anyway. And if someone really wants them, they could always host alternative sources with older builds or something like that.

Software moves forward, after all. Maintaining extremely wide compatibility with versions from two years ago feels inherently unnecessary from a user’s perspective. What really matters is good support for current stable releases and experimental builds.

I haven’t updated to this new version yet, but I’m glad to see you working on fixes. I’m currently on version 8.0 from this year, so I’ll probably stick with it for a while. Also, the patch-only updates approach already helps keep versions usable for a long time.

Anyway, thanks for the work. I just wanted to drop my completely unnecessary opinion here 😄

LisoUseInAIKyrios:
The oldest supported version is already 2025.48.0

https://github.com/MorpheApp/morphe-patches/blob/63a809c854ace4983c309bae1bedef88585626a4/patches/src/main/kotlin/app/morphe/patches/reddit/shared/Constants.kt

PR #816, opened 2026-03-11
fix(YouTube - Hide layout components): Resolve "Hide community posts" not working when selecting a channel from subscribed channels bar in Subscriptions tab

### LINKED ISSUE #814: bug(YouTube): Community posts are not hidden
### Bug description

Community posts are still shown (all kinds of) in the Subscriptions feed even after disabling it.

<details>

Image
Image

</details>

### Acknowledgements

ILoveOpenSourceApplications:
### Description

Closes #814.

### Additional context

Previously, opening a channel from the subscribed channels bar met the exact same conditions used to exempt channel page, causing community posts to bypass the filter. This update adds an explicit check for the Subscriptions tab to ensure posts are properly hidden in this context.

### Test results

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/LayoutComponentsFilter.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/LayoutComponentsFilter.java
@@ -81,6 +81,7 @@ public final class LayoutComponentsFilter extends Filter {
     private final StringFilterGroup chipBar;
     private final StringFilterGroup channelProfile;
     private final StringFilterGroupList channelProfileGroupList;
+    private final StringFilterGroupList communityPostStringFilterGroup;
 
     public LayoutComponentsFilter() {
         exceptions.addPatterns(
@@ -134,6 +135,16 @@ public final class LayoutComponentsFilter extends Filter {
                 "text_post_responsive_root.e",
                 "poll_post_responsive_root.e",
                 "shared_post_root.e"
+        );
+        communityPostStringFilterGroup = new StringFilterGroupList();
+        communityPostStringFilterGroup.addAll(
+                new StringFilterGroup(
+                        null,
+                        // home
+                        "horizontalCollectionSwipeProtector=null",
+                        // subscriptions
+                        "heightConstraint=null"
+                )
         );
 
         final var subscribersCommunityGuidelines = new StringFilterGroup(
@@ -401,17 +412,7 @@ public final class LayoutComponentsFilter extends Filter {
         }
 
         if (matchedGroup == communityPosts) {
-            // Always hide community posts when on the Subscriptions tab
-            if (NavigationButton.getSelectedNavigationButton() == NavigationButton.SUBSCRIPTIONS) {
-                return true;
-            }
-
-            // Allow community posts on channel profile page,
-            // or if viewing an individual channel in the feed.
-            if (NavigationBar.isBackButtonVisible() && !NavigationBar.isSearchBarActive()
-                    && PlayerType.getCurrent() != PlayerType.WATCH_WHILE_MAXIMIZED) {
-                return false;
-            }
+            return communityPostStringFilterGroup.check(contextInterface.toString()).isFiltered();
         }
 
         if (exceptions.matches(path)) return false; // Exceptions are not filtered.


extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/LayoutComponentsFilter.java
```diff
@@ -393,13 +393,18 @@ boolean isFiltered(String identifier, String accessibility, String path, byte[]
...
         }
 
-        if (matchedGroup == communityPosts
-                && NavigationBar.isBackButtonVisible()
-                && !NavigationBar.isSearchBarActive()
-                && PlayerType.getCurrent() != PlayerType.WATCH_WHILE_MAXIMIZED) {
+        if (matchedGroup == communityPosts) {
+            // Always hide community posts when on the Subscriptions tab
+            if (NavigationButton.getSelectedNavigationButton() == NavigationButton.SUBSCRIPTIONS) {
+                return true;
```
LisoUseInAIKyrios: This can be reached if the nav tab is subscription but the player is open.

Can also be reached if searching while on subscription tab.

Are either of these situations relevant to community posts?

ILoveOpenSourceApplications: It's returning true, so in all those conditions it'll be hidden, right? Or will there be an issue due to this check being there?

LisoUseInAIKyrios: What if a channel page is visited while the navigation is on subscription? Or an individual channel is selected in the subscription tab?

Currently community posts are allowed in those situations. This change will hide the community posts.

This check could be incorporated into the existing search/back button check, or the check is after the search/back check.

ILoveOpenSourceApplications: Because both the subscribed channels bar feed and a channel page trigger `isBackButtonVisible() == true` while on the Subscriptions tab, distinguishing them is tricky. Is there a specific path identifier we can check, so we can whitelist actual channel pages without letting the subscribed channels bar feed bypass the filter?

LisoUseInAIKyrios: Maybe this isn't a bug at all?

If someone is wanting to view the channel from the subscription filter bar, isn't that the same as viewing a channel page itself?

ILoveOpenSourceApplications: I don't know how to answer that. Depends on the eye of the user perhaps.

ILoveOpenSourceApplications: How about something like?
```kotlin
if (matchedGroup == communityPosts) {
            if (path != null && path.contains("channel_profile")) {
                return false; 
            }

            if (PlayerType.getCurrent() == PlayerType.WATCH_WHILE_MAXIMIZED || NavigationBar.isSearchBarActive()) {
                return true;
            }

            if (!NavigationBar.isBackButtonVisible()) {
                return true;
            }

            if (NavigationButton.getSelectedNavigationButton() == NavigationButton.SUBSCRIPTIONS) {
                return true;
            }

            return false;
        }
```

LisoUseInAIKyrios: Is `NavigationButton.getSelectedNavigationButton() == NavigationButton.SUBSCRIPTIONS` still needed?

If the rest of that works then it seems ok.

inotia00:
Instead of checking the navigation button state, it would be better to use a ConversionContext filter like RVX:

https://github.com/inotia00/revanced-patches/blob/54ce1d4808b12903602a1a0d9a721ee835093c38/extensions/shared/src/main/java/app/revanced/extension/youtube/patches/components/FeedComponentsFilter.java#L39-L42

To do this, the LithoFilterPatch needs a field for the conversionContext in addition to the path and identifier fields.

inotia00:
The last two commits implement the same approach used in RVX into Morphe (Additional refactoring may be required).

If you encounter any unintended bugs, you can revert these commits.

PR #873, opened 2026-03-16
feat(YouTube): Add `Reload video` patch

### LINKED ISSUE #398: feat(YouTube): Refresh / reload TV spoof client video button
### Motivation

Refresh the video button
I hope the video refresh button always adds a display function

Image

### Acknowledgements

inotia00:
### Description

Closes https://github.com/MorpheApp/morphe-patches/issues/398.

### Additional context

The patch was implemented as an independent patch instead of relying on `Spoof video streams` for the following reasons:

1. Root users cannot use the `Spoof video streams` patch.
2. `Reload video` might be useful for someone who wants to refresh related videos.

### Test results

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/LoopVideoPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/LoopVideoPatch.java
@@ -11,7 +11,6 @@ public class LoopVideoPatch {
         boolean shouldLoop = status != null && "ENDED".equals(status.name())
                 && Settings.LOOP_VIDEO.get();
         // Instead of calling a method to loop the video, just seek to 00:00.
-        if (shouldLoop) VideoInformation.seekTo(0);
-        return shouldLoop;
+        return shouldLoop && VideoInformation.seekTo(0);
     }
 }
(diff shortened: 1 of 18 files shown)


inotia00:
PR was marked as draft because it was not tested in the experimental version.

0xrxL:
Add a way to get status of loading after a while (5/7 seconds) and reload it automatically. 👍

MondayNitro:
Does the reloaded video play from the same point where reload button was clicked?

Domiiniik:
I'd be a great option if it will automatically try to reload the video if its buffering for longer then x period because Youtube can sometimes glitch out and buffer indefinitely until you reload. That's the only usefull use case I see for it. I might be wrong tho I dunno.

imPratyoosh:
> Does the reloaded video play from the same point where reload button was clicked?

In RVX this was added for JavaScript clients specifically, once playback has started most probably you don't need to reload again

0xrxL:
> > Does the reloaded video play from the same point where reload button was clicked?
> 
> In RVX this was added for JavaScript clients specifically, once playback has started most probably you don't need to reload again

Sometimes it is also useful for regular clients, who inexplicably take longer to load some videos, that once restarted will be loaded quickly. It's not a so common issue, but happen.

error-reporting:
> PR was marked as draft because it was not tested in the experimental version.

I think also the reload button doesn't show for yt shorts which needs it the most on rvx so it may or may not be that useful unless the button could be shown in shorts too

PR #874, opened 2026-03-16
fix(YouTube): Prompt to restart app on first launch of clean install

LisoUseInAIKyrios:
### Description

To fix various bugs that only happen on the first app session of a clean install, the app now prompts to restart once after the initial launch.

Code adapted from https://github.com/inotia00/revanced-patches

### Test results


sr-shishir:
So from now there will launch two prompts? One is for experimental (if experimental version patched) & the other one for restart the app right?

LisoUseInAIKyrios:
Experimental dialog will show (but it's not a restart dialog, just a confirmation), and then another prompt to restart the app.

Will check now that the dialogs behave correctly and don't show both at once.

sr-shishir:
Yes that's I'm talking about ❝The Experimental Dialog❞. It’s good but maybe users get annoyed for opening 1st dialog window and then restart prompt. Who knows. 🤷‍♂️

LisoUseInAIKyrios:
Both dialogs are needed. Only one is a restart.

PR #887, opened 2026-03-16
chore: Add credits dialog

MarcaDian:
### Description

Add a dedicated Credits dialog with current and prior contributor sections including date ranges. Redesign the About dialog with modern Morphe brand style.
___

about

credits

### DIFF
--- a/extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/MorpheAboutPreference.java
+++ b/extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/MorpheAboutPreference.java
@@ -87,24 +87,28 @@ public class MorpheAboutPreference extends Preference {
 
         final boolean preferred;
         final String name;
+        @Nullable
+        final String subText;
         final String url;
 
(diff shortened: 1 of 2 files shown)


sr-shishir:
This was very necessary. It should have been added earlier when the Morphe project launched.

goatkou:
I’m not very familiar with this, but would it be possible to replace the letters with the projects’ icons instead?
Picsart_26-03-17_04-57-30-105

LisoUseInAIKyrios:
The prior project icons are copyrighted and trademarked, and generally considered not open source. We can't use them.

sr-shishir:
It says ReVanced 2022–2025, but now it’s 2026 and the ReVanced project is still ongoing. If Osum sees this he’ll go crazy. 😂

MarcaDian:
This data is the last time the codes were used.

LisoUseInAIKyrios:
Correct, we [hard forked](https://producingoss.com/en/forks.html) ReVanced and have not brought in a single line of code from it since. The times of 2022-2025 is correct.

sr-shishir:
I see.

PR #891, opened 2026-03-17
perf: Reduce patches bundle size

wchill:
### Description

Reduces Morphe patches bundle size (~31MB down to ~6MB). Also removes need to reload JS context after every 12 videos (impact on memory usage to be tested; this may need to be reintroduced and the number adjusted).

Closes #379.

### Test results

- [X] Tested on both the **minimum** and **maximum** supported versions
- [X] Tested on **experimental** supported versions (Optional)

PR #919, opened 2026-03-18
fix(Spoof video streams): Handle Number.toLocaleString() for webview versions that have issues

wchill:
### Description

Fixes the issue mentioned in https://github.com/MorpheApp/morphe-patches/pull/891#issuecomment-4083794067

### Additional context

This appears to be caused by an issue in the underlying WebView's support for internationalization.

The following code (unminified) errors:
```javascript
    ne = function (z) {
        var K = null;
        try {
            K = z.toLocaleString("en", { style: "percent" });
        } catch (t) {
            K = z.toLocaleString(void 0, { style: "percent" });
        }
        return K;
    };
```
By patching `Number.prototype.toLocaleString` to effectively be a no-op, the underlying WebView's ICU library is not called and thus should not error.

### Test results


wchill:
Prebuilt bundle for testing available here for the next 24 hours https://wormhole.app/Pp7xNA#PNZjKLPbfs08jzSZQG3cVA

Issue #163, opened 2026-01-10
feat(GmsCore support): Add a setting to stop prompting to disable microG battery optimization


ArcherLegend:
### Feature description

With RVX you could disable the pop up asking you to disable battery optimisation for microG:

-on rvx- builder from inotia00 you could flag that in the settings.txt file generated by the builder before applying patches
- inside Anddea's fork there is a specific patch to have that setting disabled directly in app.

Please implement at least one of both.

### Motivation

On many devices like Samsung devices microG works with no issues even if battery optimisation is on and microG is in the suspension (not the advanced suspension) state so there is no need to run it unrestricted and get worse battery life.

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] This is a feature request for YouTube or YT Music

LisoUseInAIKyrios:
It would be helpful to have comparisons of microG battery usage when background optimizations are allowed/denied in some kind of controlled environment test.

But, since microG is typically using less than 0.1% of battery when optimizations are disabled, it's very difficult to compare because any potential change is too small to mater.

ArcherLegend:
One my device it makes a difference and it is a setting that has already easily been implemented in various ways inside RVX different branches so I don't think that it is something too hard or useless to add as there are already implementations out there with RVX.

NBruderman:
> It would be helpful to have comparisons of microG battery usage when background optimizations are allowed/denied in some kind of controlled environment test.
> 
> But, since microG is typically using less than 0.1% of battery when optimizations are disabled, it's very difficult to compare because any potential change is too small to mater.

I saw multiple complaints on reddit that Morphe microG causes battery drain compared to the revanced ones. Personally, I use the root method, so I skip the microG use entirely, so I can't confirm or debunk it

LisoUseInAIKyrios:
Most users confuse YouTube battery usage with microG battery usage.

I haven't seen a single user show any microG release using more than 1% battery usage. Many devices falls under 0.1% and Android doesn't show at all in battery stats because it's so low.

LisoUseInAIKyrios:
For anyone with _YouTube_ battery issues, see the [prior post](https://github.com/ReVanced/revanced-patches/issues/3092#issuecomment-2294892894) I made about this:

#### Changing battery optimization settings for MicroG will not change the battery usage of patched _YouTube_.
If anyone wants to reduce battery usage of _YouTube_ they can try:

[...]

PR #921, opened 2026-03-18
feat(GmsCore support): Add a setting to not prompt to disable MicroG battery optimization

MarcaDian:
### Description

Closes https://github.com/MorpheApp/morphe-patches/issues/163

### Test results

### DIFF
--- a/extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/BaseSettings.java
+++ b/extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/BaseSettings.java
@@ -30,7 +30,7 @@ public class BaseSettings {
     public static final StringSetting EXPERIMENTAL_APP_CONFIRMED = new StringSetting("morphe_experimental_app_target_confirmed", "", false, false);
 
     /** If the GmsCore battery optimization dialogs should be shown. */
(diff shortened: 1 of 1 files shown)


patches/src/main/resources/addresources/values/shared/strings.xml
```diff
@@ -142,4 +142,7 @@ Disabling battery optimizations for MicroG will not negatively affect battery us
 
 Tap the continue button and allow optimization changes."</string>
     <string name="gms_core_dialog_continue_text">Continue</string>
+    <string name="morphe_gms_core_battery_optimization_dialog_title">Battery optimization dialog</string>
+    <string name="morphe_gms_core_battery_optimization_dialog_summary_on">MicroG battery optimization dialog is shown</string>
+    <string name="morphe_gms_core_battery_optimization_dialog_summary_off">MicroG battery optimization dialog is not shown</string>
```
LisoUseInAIKyrios, line 147: This needs a user dialog warning that lists all known problems if MicroG does not have battery optimizations disabled:

- App may randomly close
- Notifications may not work
- Background playback may not work

MarcaDian: 👌

PR #928, opened 2026-03-19

### LINKED ISSUE #923: feat(YouTube): Set default client to "Android Reel"
### Feature description

Image

Please Add default Client to - Android Reel 

### Motivation

Good

### Acknowledgements

inotia00:
### `Android VR` client versions have been updated

- Google has recently discontinued support for clients released in 2023 and 2024.

- **YouTube Studio 23.47.101** and **YouTube 19.47.53**, released in 2023 and 2024, now show a **Playback error (App is outdated?)** toast message.

- Since YouTube VR 1.47.48 was released in **July 2023** and YouTube VR 1.54.20 in **February 2024**, it is highly likely that support will be discontinued in the near future.

- For this reason, the client versions of YouTube VR have been updated.

### Default client has been changed to `Android Reel`

- NewPipe has been using only **Android Reel** for two years and works for most users.

- I have tested it for a few weeks, and it seems there won't be any major issues even if it is used as the default client.

### Test results

### DIFF
--- a/extensions/music/src/main/java/app/morphe/extension/music/patches/spoof/SpoofVideoStreamsPatch.java
+++ b/extensions/music/src/main/java/app/morphe/extension/music/patches/spoof/SpoofVideoStreamsPatch.java
@@ -1,8 +1,8 @@
 package app.morphe.extension.music.patches.spoof;
 
 import static app.morphe.extension.music.settings.Settings.SPOOF_VIDEO_STREAMS_CLIENT_TYPE;
-import static app.morphe.extension.shared.spoof.ClientType.ANDROID_VR_1_47_48;
-import static app.morphe.extension.shared.spoof.ClientType.ANDROID_VR_1_54_20;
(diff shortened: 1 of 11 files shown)


0xrxL:
Suggestion:

Make an interactable list for clients (with the ability to move every line up/down), with the first default clients at the top. In this way at every fail the immediately next client will be picked.

LisoUseInAIKyrios:
Users need a lot of guidance on what to do with spoof streams settings. Letting them rearrange the preferred order requires intimate knowledge of why they want one preferred order over another, which very few will have.

A simpler option may be to have a different spoof client setting for Shorts, especially since Shorts doesn't always play nice with JS client on budget devices.

imPratyoosh:
> A simpler option may be to have a different spoof client setting for Shorts, especially since Shorts doesn't always play nice with JS client on budget devices.

I believe this was an exiting feature in rvx about 1 or 2 years earlier but got merged into one later on

PR #960, opened 2026-03-22
perf(Litho filtering): Use Boyer-Moore-Horspool algorithm instead of KMP

YT-Advanced:
### Description

Boyer-Moore-Horspool algorithm is faster than KMP in byte matching. Moreover, patterns used to search is mostly unique chars (no repeating), so we can't gain any benefit from KMP
- KMP Average Time: 280632.126ns
- BMH Average Time: 98565.4416ns
### Test results

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/FilterGroup.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/FilterGroup.java
@@ -197,14 +197,12 @@ class ByteArrayFilterGroup extends FilterGroup<byte[]> {
     public FilterGroupResult check(final byte[] bytes) {
         int matchedLength = 0;
         int matchedIndex = -1;
-        
         if (isEnabled()) {
             int[][] tables = skipTables;
             if (tables == null) {
-                buildSkipTables(); // Lazy load
+                buildSkipTables(); // Lazy load.
                 tables = skipTables;
             }
-            
             for (int i = 0, length = filters.length; i < length; i++) {
                 byte[] filter = filters[i];
                 matchedIndex = indexOf(bytes, filter, tables[i]);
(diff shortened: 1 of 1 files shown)


LisoUseInAIKyrios:
It would be helpful to see perf logging to show before/after.

To keep the test consistent can use copy pasted buffer string passed into ReturnYouTubeDislikesFilter, and then run simple `System.currentTimeMillis()` or `System.nanoTime()` checks before/after calling. Can stick this test code in some startup hook and run before/after this perf change.

fromhgbwithluv:
> To keep the test consistent can use copy pasted buffer string passed into ReturnYouTubeDislikesFilter/

Uhm in case of ReturnYouTubeDislikesFilter, the difference isn't very noticeable, since my code is simply trying to take advantage of the JVM's unroll-loop.

LisoUseInAIKyrios:
Ok then try using filter group on a large buffer and record the performance change.

fromhgbwithluv:
> Ok then try using filter group on a large buffer and record the performance change.

I tried to extract a very large protobuffer and run benchmark, BMH is significant faster than KMP. However, the difference in reality is negligible imo (screenshot below).

P/s: In fact, the parts of Morphe that use KMP are relatively few; I only see Hide fullscreen ads, hide membership button, and hide mix playlist that utilize KMP.

## KMP
image


## BMH
image

LisoUseInAIKyrios:
Use nanoTime() logging

LisoUseInAIKyrios:
If there isn't much appreciable benefit to changing the algorithm, and since it's already using so little time searching with the existing algorithm, then I'm not sure if the change is worth it. See rule 2 https://www.cs.unc.edu/~stotts/COMP590-059-f24/robsrules.html

But if nanotime shows any appreciable improvement then we could merge because the changes are self contained and maybe future code might use this code more often.

fromhgbwithluv:
> Use nanoTime() logging

### Result
- KMP Average Time: 280632.126ns
- BMH Average Time: 98565.4416ns

LisoUseInAIKyrios:
So, roughly 2.5x speed up.

This isn't a change from a simple algorithm to a complex one, but a change from one complex to another complex.

So there isn't anything to lose if everything works. We can merge this.

PR #997, opened 2026-03-24
feat: Add search and file export to debug logs

### LINKED ISSUE #929: feat: Export debugging logs to a text file
### Feature description

RVX has the option to export debugging logs to a text file, instead of clipboard, so it would nice to have that as option.

### Motivation

For some users this method is preferably, since clipboard might not always work for some people.  

### Acknowledgements

### LINKED ISSUE #992: bug(YouTube): When exporting to a file in the settings export, it merges over the existing content without replacing the existing file
### Bug description

When exporting to a file in the settings export/import function, if an existing file contains settings, an error occurs where the settings are merged with the existing content instead of being replaced with the new content.

<p>

Image
Image
Image
Image
Image

</p>
</details> 

### App version patched

youtube 21.12.524

### Error logs

```shell

```

### Acknowledgements

ILoveOpenSourceApplications:
### Description

https://github.com/user-attachments/assets/5982cfb4-087d-49a0-9186-3cf7e79e01a5

Closes #929 and #992.

### Additional context

N/A

### Test results


MarcaDian:
I didn't notice it in the video, but it would be cool if the entered text was highlighted.

ILoveOpenSourceApplications:
Should it match the light gray similar to that of the Morphe search or some other color?

ILoveOpenSourceApplications:
Is this good @MarcaDian?

<p>

| Dark | Light |
|--------|--------|
| image | image | 

</p>
</details>

MarcaDian:
Awesome 😎

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/AbstractPreferenceFragment.java
```diff
@@ -510,9 +507,11 @@ private EditText getEditText(Context context) {
     public void exportActivity() {
         try {
             Setting.exportToJson(getActivity());
-
+            Context context = getActivity();
+            String appName = context.getApplicationInfo().loadLabel(context.getPackageManager()).toString();
```
LisoUseInAIKyrios: Can use https://github.com/MorpheApp/morphe-patches/blob/283625288e579b2685ef046b46e0805377a05744/extensions/shared/library/src/main/java/app/morphe/extension/shared/Utils.java#L144

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/AbstractPreferenceFragment.java
```diff
@@ -560,7 +561,7 @@ protected static void showLocalizedToast(String resourceKey, String fallbackMess
 
     private void exportTextToFile(android.net.Uri uri) {
         try {
-            OutputStream out = getContext().getContentResolver().openOutputStream(uri);
+            OutputStream out = getContext().getContentResolver().openOutputStream(uri, "rwt");
```
LisoUseInAIKyrios: These file input/output streams can use a try close block:

```
try (OutputStream out = getContext().getContentResolver().openOutputStream(uri)) {

     if (out != null) ....
}
```

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/LogBufferManager.java
```diff
@@ -23,19 +43,55 @@ public final class LogBufferManager {
...
-     * The buffer is limited to approximately {@link #BUFFER_MAX_BYTES} or {@link #BUFFER_MAX_SIZE}
-     * to prevent excessive memory usage.
-     *
-     * @param message The log message to append.
+    /** * A thread-safe, dynamic list of log prefixes that should be aggressively deduplicated.
      */
+    private static final Set<String> SPAMMY_PREFIXES = new CopyOnWriteArraySet<>(Arrays.asList(
+            "LithoFilterPatch:",
+            "SpoofVideoStreamsPatch:",
+            "EnableDebuggingPatch:"
```
LisoUseInAIKyrios: These could use the class simple name:

`SpoofVideoStreamsPatch.class.getSimpleName()`

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/LogBufferManager.java
```diff
@@ -23,19 +43,55 @@ public final class LogBufferManager {
...
+            for (String prefix : SPAMMY_PREFIXES) {
+                if (log.startsWith(prefix)) {
+                    isSpammy = true;
+                    break;
+                }
+            }
+
+            if (isSpammy) {
+                if (uniqueNoisyLogs.add(log)) {
+                    filteredOutput.append(log).append("\n");
```
LisoUseInAIKyrios:
```suggestion
                    filteredOutput.append(log).append('\n');
```

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/LogBufferManager.java
```diff
@@ -58,25 +114,24 @@ public static void appendToLogBuffer(String message) {
...
-    public static void exportToClipboard() {
+    private static void exportToClipboard(String logsToExport) {
         try {
             if (!BaseSettings.DEBUG.get()) {
                 Utils.showToastShort(str("morphe_debug_logs_disabled"));
                 return;
             }
 
-            if (logBuffer.isEmpty()) {
+            if (logsToExport == null || logsToExport.trim().isEmpty()) {
```
LisoUseInAIKyrios:
```suggestion
            if (logsToExport == null || logsToExport.isBlank()) {
```

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/LogBufferManager.java
```diff
@@ -85,6 +140,67 @@ public static void exportToClipboard() {
...
+    private static void exportToFile(Context context, String logsToExport) {
+        try {
+            if (!BaseSettings.DEBUG.get() || logsToExport == null || logsToExport.trim().isEmpty()) {
+                Utils.showToastShort(str("morphe_debug_logs_none_found"));
+                return;
+            }
+
+            pendingLogsToExport = logsToExport;
+
+            String appName = context.getApplicationInfo().loadLabel(context.getPackageManager()).toString();
```
LisoUseInAIKyrios: Use Utils name method as above.

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/LogBufferManager.java
```diff
@@ -110,4 +226,189 @@ public static void clearLogBuffer() {
...
+                    logViewer,
+                    str("morphe_debug_export_logs_copy"),
+                    () -> exportToClipboard(logViewer.getText().toString()),
+                    () -> {},
+                    str("morphe_debug_export_logs_clear"),
+                    LogBufferManager::clearLogBuffer,
+                    true
+            );
+
+            int margin = (int) android.util.TypedValue.applyDimension(android.util.TypedValue.COMPLEX_UNIT_DIP, 16f, context.getResources().getDisplayMetrics());
```
LisoUseInAIKyrios: Use the DIM class to get this margin value.

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/LogBufferManager.java
ILoveOpenSourceApplications: I tried to implement this, but `LogBufferManager` is in the shared module while these patches are in the shared-youtube and youtube modules. Since shared can't import classes from the extension modules, I had to leave them as hardcoded strings to avoid breaking the module boundaries.

LisoUseInAIKyrios: Ahh, yes. No way around this then.

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/LogBufferManager.java
LisoUseInAIKyrios: Is there a need to check for duplicate log entries?

This will ignore the same log message if it repeats more than once.

ILoveOpenSourceApplications: I can remove it, given we are already handling aggressive deduplication for known noisy patches via the `SPAMMY_PREFIXES`.

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/LogBufferManager.java
```diff
@@ -23,16 +43,49 @@ public final class LogBufferManager {
...
+
+    @NonNull
+    private static String getFilteredLogs() {
+        java.util.LinkedHashSet<String> uniqueNoisyLogs = new java.util.LinkedHashSet<>();
+        StringBuilder filteredOutput = new StringBuilder();
+
+        for (String log : logBuffer) {
+            boolean isSpammy = false;
+
+            for (String prefix : SPAMMY_PREFIXES) {
```
LisoUseInAIKyrios: I'm still seeing the 'spammy' outputs in the logs. Is this logic working or is it enabled somehow?

But the spam output is part of the logs and is relevant. The default log output should include everything in the logs, including spammy stuff.

ILoveOpenSourceApplications: It's fixing spam to some extent, but logs being logs, if there's even a minor change, that still makes it unique and thus bypasses the filter.

LisoUseInAIKyrios: I think all the logs can be shown and there is no filtering. Imagine a user posts a log but it's "less spammy" but is missing data. Troubleshooting could be very difficult if stuff is left out.

The debug flags mostly show only on app startup (start of the logs), and the litho logs are important.

ILoveOpenSourceApplications: The problem with the in-app debug logs, I believe, is that the user has to go through You tab evey time ro export, which changes the logs from what they want to capture to`horizontal_shelf` and all the button icons `yt_outline` due to history shelf.

LisoUseInAIKyrios: But the logs will still contain all the prior entries. So any new logs generated by the you tab will be a few extra entries at the bottom and can be ignored by the user/us.

ILoveOpenSourceApplications: You can check and see if it's making any difference, if not, go ahead and remove it. I just wanted to reduce spam logs so that the user can copy paste the exact logs associated with their issue and not `horizontal_shelf.e` for the nth time.

LisoUseInAIKyrios: We can start asking users to export to a file so all logs are present, and we sift thru to find the important stuff.

ILoveOpenSourceApplications: If the export file will have more than what's being shown in-app, then yes, that's a good solution. But is that the case?

LisoUseInAIKyrios: Right now I'm seeing everything in app (which I think it should show everything).

The file output can include more log entries because copying has a size limit. Need to double check the limit for file output is larger.

Maybe the log file should be directly wrote to in a temp app fder and then save file simply opens a file save picker with that file. Then it's all logs since startup.

ILoveOpenSourceApplications: If so, then that would be a good alternative.

LisoUseInAIKyrios: For now let's use the existing smaller in-memory log buffer. Can change to a temporary file if situations come up where the memory log isn't large enough.

MarcaDian:
I remember when we added the new window design I added code to move the buttons to a new line when they took up a lot of space. Now it doesn't work.

MarcaDian:
I fixed the adaptability of the dialog buttons and used existing code instead of custom buttons.

extensions/shared/library/src/main/java/app/morphe/extension/shared/settings/preference/AbstractPreferenceFragment.java
```diff
@@ -422,34 +420,6 @@ public static void showRestartDialog(Context context) {
...
-        LinearLayout.LayoutParams params = new LinearLayout.LayoutParams(0, height, 1.0f);
-        params.setMargins(marginLeft, 0, marginRight, 0);
-        btn.setLayoutParams(params);
-        btn.setOnClickListener(listener);
-
-        return btn;
-    }
     public void showImportExportTextDialog() {
         try {
             Activity context = getActivity();
```
MarcaDian, line 430: @LisoUseInAIKyrios If we set the `activity` here instead of the `context`, the dialog does not use the Morphe locale. Can we set the `context` here?

LisoUseInAIKyrios: Yes can use the utils context if that works.

MarcaDian: Clarification. The problem is not in this place. All dialogs with edit text and the sponsor block section also do not use the Morphe locale.

Also search settings and About dialog.

LisoUseInAIKyrios: Some places an activity may be needed and not just any context. 

But utils context is usually an activity too but not always depending on which extension hook most recent has run.

If an activity is needed then we could modify the utils setActivity to apply the Morphe language like the regular context does.

https://github.com/MorpheApp/morphe-patches/blob/283625288e579b2685ef046b46e0805377a05744/extensions/shared/library/src/main/java/app/morphe/extension/shared/Utils.java#L435

PR #1037, opened 2026-03-29
feat(YouTube - Setings): Move OPEN_VIDEOS_FULLSCREEN_PORTRAIT above EXIT_FULLSCREEN

ghost:
### Description

This makes more sense to keep these settings close to each other, because one is opening in fullscreen portait and one is exiting fullscreen.

Once OPEN_VIDEOS_FULLSCREEN_LANDSCAPE is added, you can add it here as well. It's nice to keep these all together, it makes it easier for users as well.

Closes #

### Additional context

N/A

### Test results


LisoUseInAIKyrios:
The settings in the UI are not shown in the order of the settings.java

There is a way to group together settings in the UI, but it's done differently than here.

Make a feature request to do this and someone else can make the code change needed.

PR #1068, opened 2026-04-01
feat: Use shared morphe-patches-library for common code

### LINKED ISSUE #248: feat: Seperate repository for patch Utils
### Feature description

I think it would make sense to move the utilities (`patches/src/main/kotlin/app/morphe/util`) to a separate repo and publish a package for consumption.

### Motivation

One of consistent DX annoyances with Revanced was that the patch utility functions are colocated in the org's patches repo. This means everyone creating there own custom patches has to copy at least some of that code (`BytecodeUtils` is a common example), and it's no longer connected upstream. Similarly, any helper functions that I write becomes more cumbersome to PR and have little visibility to other devs.

I think that makes even more sense with Morphe, which is officially only supporting youtube patches at this time.

### Acknowledgements

- [ ] This is a feature request for YouTube or YT Music

wchill:
### Description

Deduplicates code and allows other 3rd party patch developers to contribute to and use a shared library without having to copy paste.

Closes #248

### Additional context

For other developers wishing to also take a dependency on the library, update the gradle patches plugin to 1.3.0 or higher and then add `app.morphe:morphe-patches-library` and `app.morphe:morphe-extensions-library` as dependencies. (These dependencies are not added automatically by the plugin, in case you don't want to use them.)

https://github.com/MorpheApp/morphe-patches-library

You should also migrate away from context receivers at this time and use context parameters instead. A future version of the gradle patches plugin will bump the Kotlin version to 2.3.0+ which removes support for context receivers entirely.

### Test results

### TODO

- [x] Finalize and merge https://github.com/MorpheApp/morphe-patches-library/pull/2
- [x] Finalize and merge https://github.com/MorpheApp/morphe-patches-gradle-plugin/pull/14
- [x] Bump gradle plugin + library versions in this PR


wchill:
Still need to remove hexPatch from morphe-patches.

CheckStringResources might also be good to move, but it uses a hardcoded list of app names right now.

Moving Morphe about dialog + preference code not specific to YouTube to the library might also make sense.

LisoUseInAIKyrios:
All the patches kotlin preference code is a mess and I think should be thrown away. It requires too many intermediate steps that serve no good purpose:

- Patches programmatically creates objects that represent preference xml > serialize custom kotlin classes to preference xml > deserialize xml to Preference objects at app runtime > Extension code does more processing of the preferences to sync with Setting objects > done

The way Reddit does it is better because it skips all the xml. It's also way more flexible because not all apps can easily be resource patched:
- Extension patches programmatically creates preference objects based on settings > done

LisoUseInAIKyrios:
`CheckStringResources` can be extracted and the app specific config (app languages, resource types) can be set by the patches to use it.

wchill:
~~Reddit still needs to be updated to use Crowdin translations, so that should be figured out before we add preference code to the library, then.~~

Also it's still taking a dependency on AbstractPreferenceFragment etc, so those still need to be copied

settings.gradle.kts
```diff
@@ -48,3 +48,16 @@ mapOf(
...
+// It is a multi-module project, so each artifact maps to a specific subproject.
+file("../morphe-patches-library").let { libDir ->
+    if (libDir.exists()) {
+        includeBuild(libDir) {
+            dependencySubstitution {
+                substitute(module("app.morphe:morphe-patches-library")).using(project(":patch-library"))
+                substitute(module("app.morphe:morphe-extensions-library")).using(project(":extension-library"))
+            }
+        }
+    }
```
brosssh, line 62: I'm not sure I understand why this is needed, I don't have it [here](https://github.com/brosssh/morphe-patches) and it works just fine.

LisoUseInAIKyrios: This allows compiling and building changes to morphe-patches-library locally without publishing to maven local. It's optional and not required.

PR #1091, opened 2026-04-04
fix(Change package name): `DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION` is not updated correctly

FlaminSarge:
### Description
Regression from 60a2f9e5b27ca126c4cce3a7207c197675f10d80

Closes #

### Additional context

Found this while trying to debug why GmsCore patch wasn't updating C2D_MESSAGE permission package name in YT 21.x (which turned out to be just me forgetting to -f).
Note that GmsCore patch hid this issue because it would also make the same rename (but correctly).

### Test results
Verified manifest has the updated name on the receiverNotExported permission for 20.45.36. I don't think this needs testing on the other versions given it's a manifest-specific bugfix.


FlaminSarge:
In testing, found another bug with this: the boolean options don't actually read from the options given because they're fetching .value immediately instead of later on when they need to be checked. Gonna update this PR with that fix too.

LisoUseInAIKyrios (approved):
Looks good, thanks for catching this.

Will merge tomorrow if there is no new reviews or concerns.

PR #1092, opened 2026-04-04
feat(YouTube - Seekbar): Add "Enable livestream DVR" setting

FlaminSarge:
Adds a setting under "Seekbar" to enable seeking on live streams that don't have seeking enabled.

### Additional context

As far as I can tell, this one method in VideoStreamingData is the only thing preventing the seek bar from fully functioning for playbackType=LIVE (not LIVE_DVR). I have not looked into whether videos that lack VideoStreamingData need additional handling.
Uses the string-based fingerprint for VideoStreamingData because the class is obfuscated in 21.x (not obfuscated in 20.x though).

FAIR WARNING: The JADX decomp was analyzed by Claude Code in order to find this method (and anything else that could be affecting liveDvr handling). I tried several other overrides in my initial testing (LiveModel was the one I had found first just by searching through JADX's results, then PlayerResponseModelImpl) which either did nothing or would enable the seek bar but not actually seek backwards, before settling on this.
To that end, the fingerprint used here (literal 8 with CONST_4 opcode at +2) is the only thing I could find that would match this method without being mistaken for the other method that checks playbackType 8 and 9 (Claude agreed but I don't fully trust its judgment).

The setting name/description/location are up for debate, too, I guess.

JS reference that informed that this was even possible: https://greasyfork.org/en/scripts/485020-ytbetter-enable-rewind-dvr/code

### Test results
Tested on 20.21.37, 20.45.36, 21.13.163, and 21.14.482.
Tested toggling the setting while an active stream was playing too: disabling the setting from enabled snaps the stream back to live if it's not already a DVR stream (once it's back in focus upon returning from the settings page), and enabling the setting from disabled allows it to seek without having to reopen the video.


ILoveOpenSourceApplications:
If this can be a setting in seekbar patch rather than a patch of its own, that would be better I think.

patches/src/main/kotlin/app/morphe/patches/youtube/video/livedvr/Fingerprints.kt
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2026 Morphe.
+ * https://github.com/MorpheApp/morphe-patches
+ *
+ * See the included NOTICE file for GPLv3 §7(b) and §7(c) terms that apply to this code.
+ */
+
+package app.morphe.patches.youtube.video.livedvr
```
FlaminSarge: Does this package also need updating to move to seekbar?
Might be worth updating the i18n keys too.

patches/src/main/kotlin/app/morphe/patches/youtube/video/livedvr/Fingerprints.kt
```diff
@@ -0,0 +1,30 @@
...
+import app.morphe.patcher.InstructionLocation.MatchAfterImmediately
+import app.morphe.patcher.literal
+import app.morphe.patcher.opcode
+import app.morphe.patches.youtube.video.quality.VideoStreamingDataToStringFingerprint
+import com.android.tools.smali.dexlib2.Opcode
+
+// Returns false only for playbackType 8 (non-DVR live).
+// Forcing true enables seeking on live streams.
+// Note: another no-param boolean method in the same class also checks literal(8),
+// but adds a type 9 check, placing CONST_4 at +4 there, vs +2 here.
```
FlaminSarge: This comment might need updating since the fingerprint is now 8 -> if-eq -> 1

patches/src/main/kotlin/app/morphe/patches/youtube/video/livedvr/Fingerprints.kt
LisoUseInAIKyrios: Yeah this can move to there.

FlaminSarge:
One caveat I somehow didn't catch during testing on an emulator but is happening on my personal device: if Spoof Video Streams is disabled, DVR still gets force-enabled but has trouble rewinding past the point that the stream started playing from (for streams that had DVR disabled). It either snaps back to live, or can't go back any further than when the stream started being viewed. I think it's related to SABR handling, possibly that the server never even sends segments before that point.

I could have sworn the only patches I had on the emulator were Force Live DVR + GmsCore, though, but perhaps it was an old enough device that SABR wasn't even supported so it never ran into that. A newer version emulator runs into the issue, though.

LisoUseInAIKyrios:
A settings user dialog message could be added, informing the feature may not work with some devices and some spoof video clients.

PR #1110, opened 2026-04-07
feat(YouTube - Seekbar): Add "Expand livestream DVR duration" setting

FlaminSarge:
This sets the seekbar to up to 7 days in the past for livestreams that go longer than 12 hours (YT normally limits it at 12 hours).
The setting has a warning on it that it may not function as intended without "Spoof video streams" enabled.

Closes #

### Additional context

Once again, setting name/description/location/warnings up for debate.

The warning regarding "Spoof video streams" is due to SABR preventing the app from receiving data past 12 hours as far as I can tell, and since that patch disables SABR it lets this work without issue.

FAIR WARNING: As with #1092, the JADX decomp was fed to Claude Code to discover this method to override. The initial fingerprint it came up with wasn't great though (just the 3 opcodes in the method without any location/field access handling), so I tried to make it more resilient other methods potentially getting the same set of opcodes in app updates, with the risk of being not as adaptable if the method itself changes in updates.

Also the reference for 7 days (in addition to my own testing) is https://greasyfork.org/en/scripts/485020-ytbetter-enable-rewind-dvr/code. As far as I could tell, video data past 7 days is a dice roll on whether it's present or not, for very long livestreams. Most often, it's missing.

### Test results
Tested on 20.21.37, 20.45.36, and 21.14.482 with "Spoof video streams" enabled. Doesn't extend past the livestream start time, either, if it's shorter than 7 days.

Screenshot 2026-04-06 at 8 58 28 PM


patches/src/main/kotlin/app/morphe/patches/youtube/interaction/seekbar/SeekbarPatch.kt
```diff
@@ -10,14 +10,16 @@ val seekbarPatch = bytecodePatch(
             "slide to seek instead of playing at 2x speed when pressing and holding, " +
             "tapping the player seekbar to seek, " +
             "hiding the video player seekbar, " +
-            "and enabling seeking in livestreams."
+            "enabling seeking in livestreams, " +
+            "and expanding the livestream DVR window."
```
FlaminSarge: wasn't sure whether to use 'duration' or 'window'

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/ExpandLivestreamDvrDurationPatch.java
```diff
@@ -0,0 +1,18 @@
+package app.morphe.extension.youtube.patches;
+
+import app.morphe.extension.youtube.settings.Settings;
+
+@SuppressWarnings("unused")
+public class ExpandLivestreamDvrDurationPatch {
+
```
LisoUseInAIKyrios: This could be combined into the other one Livestream dvr patch and rename it to be something reflective of both features.

patches/src/main/kotlin/app/morphe/patches/youtube/interaction/seekbar/ExpandLivestreamDvrDurationPatch.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    execute {
+        PreferenceScreen.SEEKBAR.addPreferences(
+            SwitchPreference("morphe_expand_livestream_dvr_duration")
+        )
+
+        FormatStreamModelMaxDvrDurationFingerprint.method.apply {
+            val returnIndex = implementation!!.instructions
+                .indexOfFirst { it.opcode == Opcode.RETURN_WIDE }
+            val returnReg = getInstruction<OneRegisterInstruction>(returnIndex).registerA
```
LisoUseInAIKyrios: This index is available using the fingerprint instruction matches (.first().index)

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/ExpandLivestreamDvrDurationPatch.java
FlaminSarge: Just the patch class itself, or combine the setting too? (Would not recommend combining the setting).

LisoUseInAIKyrios: Just the extension/patch classes. Can be single SeekbarDVRPatch for patches and the same single patch for the extensions.

patches/src/main/kotlin/app/morphe/patches/youtube/interaction/seekbar/ExpandLivestreamDvrDurationPatch.kt
FlaminSarge: It'd be `last()` to get the return instruction index, right? (Just checking, still getting familiar with this, coming from Source Engine RE). Seems like I could also use findInstructionIndicesReversedOrThrow like the other patch but seems like overkill when the method is only 3 instructions.

LisoUseInAIKyrios: Yes that's correct.

https://github.com/FlaminSarge/morphe-patches/blob/bde22491df3e8263a40b316f13ad69d6c0ea114f/patches/src/main/kotlin/app/morphe/patches/youtube/interaction/seekbar/Fingerprints.kt#L161

PR #1111, opened 2026-04-07
chore(YouTube - Seekbar): Add warning to "Enable livestream DVR" setting regarding "Spoof video streams"

FlaminSarge:
Adds a warning to "Enable livestream DVR" that it may not function correctly unless "Spoof video streams" is also enabled.

### Additional context

As with #1110, when the app is using SABR the feature may not work correctly because SABR doesn't send any data from before the user started viewing the stream, and "Spoof video streams" disables SABR as far as I can tell.
N/A

### Test results
Tested alongside #1110 

### DIFF
--- a/patches/src/main/resources/addresources/values/youtube/strings.xml
+++ b/patches/src/main/resources/addresources/values/youtube/strings.xml
@@ -572,7 +572,7 @@ Info: May not work on live streams"</string>
     <string name="morphe_livestream_dvr_summary_on">Seeking is allowed on all livestreams</string>
     <string name="morphe_livestream_dvr_summary_off">Seeking may not be available on livestreams</string>
     <!-- 'Spoof video streams' should be the same translation used for 'morphe_spoof_video_streams_screen_title'. -->
-    <string name="morphe_livestream_dvr_user_dialog_message">This setting may cause unexpected behavior if \'Spoof video streams\' is not enabled.</string>
+    <string name="morphe_livestream_dvr_user_dialog_message">Enabling this may cause unexpected livestream seeking behavior if \'Spoof video streams\' is not enabled.</string>
 
     <!-- interaction.seekbar.enableSlideToSeekPatch -->
     <string name="morphe_slide_to_seek_title">Enable slide to seek</string>
--- a/patches/src/main/resources/addresources/values/youtube/strings.xml
+++ b/patches/src/main/resources/addresources/values/youtube/strings.xml
@@ -571,6 +571,8 @@ Info: May not work on live streams"</string>
     <string name="morphe_livestream_dvr_title">Enable livestream DVR</string>
     <string name="morphe_livestream_dvr_summary_on">Seeking is allowed on all livestreams</string>
     <string name="morphe_livestream_dvr_summary_off">Seeking may not be available on livestreams</string>
+    <!-- 'Spoof video streams' should be the same translation used for 'morphe_spoof_video_streams_screen_title'. -->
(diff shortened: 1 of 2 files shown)


patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -571,6 +571,8 @@ Info: May not work on live streams"</string>
     <string name="morphe_livestream_dvr_title">Enable livestream DVR</string>
     <string name="morphe_livestream_dvr_summary_on">Seeking is allowed on all livestreams</string>
     <string name="morphe_livestream_dvr_summary_off">Seeking may not be available on livestreams</string>
+    <!-- 'Spoof video streams' should be the same translation used for 'morphe_spoof_video_streams_screen_title'. -->
+    <string name="morphe_livestream_dvr_user_dialog_message">This setting may cause unexpected behavior if \'Spoof video streams\' is not enabled.</string>
```
LisoUseInAIKyrios: Does it cause unexpected behavior, or does it simply not work?

If it doesn't work then it can say something like that.

error-reporting: I think Android VR av1 client may have sabr enabled so this client should be checked further before Adding warning to "Enable livestream DVR" setting regarding "Spoof video streams"

FlaminSarge: On the emulator, it would let you attempt to seek but then snap back to live immediately. On my personal device, it would let you rewind, but only to where you started viewing stream. It would also occasionally 'close' the rewind window to new checkpoints, so to speak. Definitely 'unexpected behavior'.

We could potentially just disable the patch entirely if Spoof video streams isn't enabled, but like @error-reporting said, some of the spoof clients might have SABR enabled so further testing might be needed for that.

LisoUseInAIKyrios: "unexpected behavior" could be "unexpected livestream seeking behavior"

FlaminSarge: That sounds about right. I'll leave it to you for the final text.

> I think Android VR av1 client may have sabr enabled

I tried setting this up, and live rewind was working alongside Android VR AV1 being enabled, but it looks like it may also just not use AV1 in some cases even with that option enabled, so I don't know if there's a definitive way to test this.

```suggestion
    <string name="morphe_livestream_dvr_user_dialog_message">Enabling this may cause unexpected livestream seeking behavior if \'Spoof video streams\' is not enabled.</string>
```

error-reporting: how's the Android VR AV1 test results on your newer emulator goes also the tv client may need to be tested too

FlaminSarge:
@LisoUseInAIKyrios FYI I'm not sure if it was one of the spoof video stream or fix-SABR commits recently, but fewer spoof targets seem to disable SABR now, breaking the livestream dvr patches. Thus far I believe Android Reel and Android VR started using SABR (and stopped working), and the only one I could get working was visionOS.
Don't have the bandwidth right now to dig into it further.

LisoUseInAIKyrios:
"Enable Livestream DVR" could add a Setting availability to restrict the feature to require stream spoofing with visionOS.

FlaminSarge:
> "Enable Livestream DVR" could add a Setting availability to restrict the feature to require stream spoofing with visionOS.

Could do that, or potentially have it auto-disable SABR with those settings enabled (I'm sure there's a way to force-disable SABR somehow, just haven't had a chance to look).

Issue #894, opened 2026-03-17
feat(YouTube): Add "Cancel" button and improve text in "Restart required" dialog (Match RVX)


mysterious-myself:
### Feature description

First off, thank you for adding the automated "Restart required" prompt! This is definitely a very good feature to have, but the current forced implementation is not ideal for all users. Currently, it forces an immediate restart with no option to dismiss the prompt. Additionally, the text inside the prompt is left-aligned which feels slightly unpolished.
![Image](https://github.com/user-attachments/assets/3922f718-c27d-4d72-8424-d5366dc36367)

### Proposed Solution

1) Add a "Cancel" button to allow users to dismiss the prompt and delay the restart.
2) Center-align the dialog text to make it look cleaner.
3) Change the secondary text to match RVX's clearer phrasing: "Restart to load the layout normally" instead of "Restart to finish setting up". (Not necessary)
4) Alternatively, simply adopt the exact dialog implementation from RVX, which already has a beautifully centered UI, the improved text, and the necessary Cancel button.

### Motivation

While the auto-restart is a good idea in theory, having a forced, un-cancelable restart actually breaks the setup flow for users dealing with specific `base-YouTube bugs`. Especially since I am facing this myself. For example, on certain patched or unpatched YouTube versions, the `Dark Theme` glitches out if the app layout is altered before Dark Theme is manually toggled on during a fresh install.

Because the current YouTubr Morphe dialog has no Cancel button, users are completely trapped: the app forces a restart and applies the new layout before the user can reach the settings menu to enable `Dark Theme`. This permanently triggers the dark mode bug. Allowing users to hit "Cancel," set up their base app settings (like Dark Theme), and then manually restart is a critical necessity.

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] This is a feature request for YouTube Morphe

LisoUseInAIKyrios:
The user must restart or numerous bugs will be present.

What app versions have this dark mode bug you are referring to? Post a video showing the problem.

FaiyazAhamadRabbi:
I tested Youtube v20.21.37

LisoUseInAIKyrios:
Post a video recording showing the issue.

FaiyazAhamadRabbi:
@LisoUseInAIKyrios Plz take a look

https://github.com/user-attachments/assets/f6c1da23-43d7-4b21-b6d6-b91c176800e4

[...]

PR #1126, opened 2026-04-09
fix(YouTube): Bold player buttons break under certain conditions

### LINKED ISSUE #1125: bug(YouTube): Bold player buttons break under certain conditions
### Bug description

It works well in most situations, but breaks in specific cases.

Image

### Prerequisites (Shared)

> [!Note]
> 
> - Two or more player buttons are enabled (e.g. `Show speed dialog button` and `Show video quality button` are enabled)
> 
> - `Restore old player button style` is disabled

### Steps for reproduction (1)

> [!Note]
> 
> - Patch YouTube **20.21.37**.
> 
> - Bold player buttons are broken even without changing any settings (Bold icons were not added in YouTube 20.21.37).
> 
> - It is better to change the default value of `Restore old player button style` to **TRUE**, and the `Restore old player button style` setting should only be added in YouTube 20.31+.

### Steps for reproduction (2)

> [!Note]
> 
> - Patch YouTube **20.45.36**.
> 
> - Manually add the following flags in `Feature flags manager`: **45643739**, **45686474**, **45709810**, **45713296**.
>
> - If `Restore old player button style` is turned off, the patch returns the **original** instead of overriding: https://github.com/MorpheApp/morphe-patches/blob/f5ded79beb21bd7833284a670e4a3b4c2fb9fc51/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/LegacyPlayerControlsPatch.java#L86-L91
> 
> - In other words, on accounts where these experimental flags are **FALSE**, the issue is reproduced without using `Feature flags manager`.
> 
> - The `usePlayerBottomControlsExploderLayout(boolean)` function must return **TRUE** instead of the **original**.

### App version patched

20.21.37, 20.45.36

### Error logs

```shell
N/A
```

### Acknowledgements

inotia00:
### Description

Closes #1125.

### Additional context

~~The default value for `Restore old player button style` has been changed to **TRUE**.~~

~~Unpatched YouTube also does not use bold icons when the app is first installed.~~

~~To minimize unintended issues, it is better to allow users to manually disable `Restore old player button style`.~~

`Restore old player button style` setting is no longer available if the app version is spoofed to **20.28** or earlier.

According to the description in the `Spoof app version target`, the bold icon should not be used in YouTube 20.28 or earlier:

```
20.28.41 - Disable bold icons
```

### Test results

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/LegacyPlayerControlsPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/LegacyPlayerControlsPatch.java
@@ -13,6 +13,7 @@ import app.morphe.extension.shared.Logger;
 import app.morphe.extension.shared.Utils;
 import app.morphe.extension.shared.settings.Setting;
 import app.morphe.extension.youtube.settings.Settings;
+import app.morphe.extension.youtube.settings.YouTubeActivityHook;
 
 @SuppressWarnings("unused")
 public class LegacyPlayerControlsPatch {
@@ -25,7 +26,7 @@ public class LegacyPlayerControlsPatch {
     }
 
     public static final boolean RESTORE_OLD_PLAYER_BUTTONS =
-            Settings.RESTORE_OLD_PLAYER_BUTTONS.get() || !IS_20_31_OR_GREATER || isSpoofingToLessThan("20.31.00");
+            Settings.RESTORE_OLD_PLAYER_BUTTONS.get() || !YouTubeActivityHook.useBoldIcons(true);
 
     public static WeakReference<View> fullscreenButtonRef = new WeakReference<>(null);
 
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/LegacyPlayerControlsPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/LegacyPlayerControlsPatch.java
@@ -1,5 +1,8 @@
 package app.morphe.extension.youtube.patches;
 
+import static app.morphe.extension.youtube.patches.VersionCheckPatch.IS_20_31_OR_GREATER;
+import static app.morphe.extension.youtube.patches.spoof.SpoofAppVersionPatch.isSpoofingToLessThan;
+
 import android.view.View;
 import android.view.ViewTreeObserver;
 import android.widget.ImageView;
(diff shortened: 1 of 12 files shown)


LisoUseInAIKyrios:
> Unpatched YouTube also does not use bold icons when the app is first installed.

What about after the user restarts when prompted? The restart prompt is forced and has no way to bypass.

inotia00:
That also seems to be one of the good workarounds.

LisoUseInAIKyrios:
If restarting fixes it, using bold icons as default is preferred.

LisoUseInAIKyrios (approved):

mysterious-myself:
> > Unpatched YouTube also does not use bold icons when the app is first installed.
> 
> What about after the user restarts when prompted? The restart prompt is forced and has no way to bypass.
@LisoUseInAIKyrios @inotia00 

Just a quick heads-up regarding the restart workaround! There are a couple of conflicts with relying on a forced restart for this:

***1• The "Cancel" button bypass: In Issue*** #894, a Cancel button was added to the Restart dialog. If a user clicks it to bypass the restart, won't they be stuck with the broken player buttons on their first launch?

***2• The v20 Dark Theme bug:*** The reason that Cancel button was added (for old android version 8 & 9) in the first place is because forcing an immediate restart caused a dark theme issue on older v20 base versions (like 20.21.37). Forcing it again will likely bring that bug back.

### A possible solution
 In RVX Patches, there is an options menu in the Manager for the "Player Icon" patch that lets users choose between 
**• Thin (by default)**
**• Bold**
**• Rounded**
If a user applies that patch, the icons change permanently without any rendering errors (even on older bases like v20.05.46).

Perhaps the rendering logic from that specific RVX patch could be adapted here instead of relying on the restart prompt? Just wanted to point this out before anything gets finalized!

LisoUseInAIKyrios:
The cancel restart is still possible with Android 10 and lower.

If we do a change player icons it's highly desired to do it with bytecode patching and not by replacing resources, so an in-app settings can change the icons.

The Morphe icons it's easy to change because we already pick between bold and non bold at runtime, but all the other built-in YT player icons may need changes too and that means some patch changes.

mysterious-myself:
@LisoUseInAIKyrios Ah, that makes complete sense! Doing it via bytecode patching so we can get an in-app toggle in the settings is definitely the superior way to handle it, rather than just statically replacing the resources.

Thanks for taking the time to break down the technical side of how you're approaching the built-in YT icons versus the Morphe ones. Really looking forward to seeing how the patch changes turn out!

PR #1128, opened 2026-04-09
feat(Reddit): Add `Hide Reddit search` patch

### LINKED ISSUE #457: feat: [ Reddit ] Remove "Reddit search" from Android Text Selection Menu
### Feature description

**Desired Implementation**
Ideally, this could be implemented as:
1.  **A Toggle:** A setting in Morphe Section (e.g., "Enable text selection menu integration") so users can turn it on/off dynamically.
2.  **Or a Standalone Patch:** A patch named "Remove Text Selection Menu" that strips the intent entirely during the patching process.

### Motivation

**Problem Description**

The official Reddit app registers a `PROCESS_TEXT` intent filter in `AndroidManifest.xml'. This forces a "Reddit search" button to appear in the system-wide text selection menu (the floating toolbar when selecting text in any app). This clutters the menu for users who do not use this feature.

**proposed Solution** Please add a patch that removes the `android.intent.action. PROCESS TEXT` intent filter from the application manifest.

Image

----

Image

### Acknowledgements

inotia00:
### Description

Closes #457.

### Additional context

N/A

### Test results

### DIFF
--- a/patches/src/main/kotlin/app/morphe/patches/reddit/layout/search/HideRedditSearchPatch.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/reddit/layout/search/HideRedditSearchPatch.kt
@@ -14,7 +14,8 @@ import app.morphe.util.findElementByAttributeValueOrThrow
 @Suppress("unused")
 val hideRedditSearchPatch = resourcePatch(
     name = "Hide Reddit search",
-    description = "Permanently hides the Reddit search in the contextual menu.",
+    description = "Permanently hides the Reddit search in the contextual menu. " +
+            "This patch does not work with root mounting",
     default = false
 ) {
     compatibleWith(COMPATIBILITY_REDDIT)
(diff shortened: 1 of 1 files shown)


patches/src/main/kotlin/app/morphe/patches/reddit/layout/search/HideRedditSearchPatch.kt
```diff
@@ -0,0 +1,40 @@
...
+import app.morphe.patcher.patch.resourcePatch
+import app.morphe.patches.reddit.misc.fix.signature.spoofSignaturePatch
+import app.morphe.patches.reddit.shared.Constants.COMPATIBILITY_REDDIT
+import app.morphe.util.findElementByAttributeValueOrThrow
+
+@Suppress("unused")
+val hideRedditSearchPatch = resourcePatch(
+    name = "Hide Reddit search",
+    description = "Permanently hides the Reddit search in the contextual menu.",
+    default = false
```
inotia00, line 19: Since the patch modifies AndroidManifest.xml, it cannot be turned on or off in the Morphe settings.

Therefore, the patch has been excluded by default.

Someone might install Reddit via a mount method (ROOT environment), in which case the patch might not work.

I am not sure if I should add this to the description.

PR #1135, opened 2026-04-09
fix(YouTube - Hide ads): Resolve "Hide YouTube Premium promotions" affecting YouTube Doodles

ILoveOpenSourceApplications:
### Description

To prevent this from happening:

image

### Additional context

N/A

### Test results


LisoUseInAIKyrios:
Try with the refactoring. I can't reproduce any promotional banners so I cannot confirm.

ILoveOpenSourceApplications:
I also can't reproduce promotional banner. But if it's not hiding YouTube doodles, then that's a win.

ILoveOpenSourceApplications:
I was able to hide promotions, so looks good.

PR #930, opened 2026-03-19
fix(YouTube - Hide layout components): Resolve "Hide horizontal shelves" hiding other components

inotia00:
### Description

Closes https://github.com/MorpheApp/morphe-patches/issues/914.

### Test results

- [X] Tested on both the **minimum** and **maximum** supported versions
- [X] Tested on **experimental** supported versions (Optional)

PR #1153, opened 2026-04-11
fix(YouTube - Hide layout components): Resolve history shelf can be hidden on initial app startup

### LINKED ISSUE #1151: bug(YouTube): History shelf is hidden until app is restarted
### Bug description

Sometimes when I open the YouTube app and tap the account icon in the bottom right corner to view my history, I can't see it. As you can see in the attached image. Even swiping down multiple times doesn't make it appear. I need to close and reopen the app to view my history normally.

Image

### App version patched

20.45.36

### Error logs

```shell

```

### Acknowledgements

utafrali:
### Description

The history shelf was getting hidden on startup because the nav bar hadn't initialized yet. `getSelectedNavigationButton()` returns null during early setup, and comparing null against `NavigationButton.LIBRARY` would evaluate to true, causing the shelf to be filtered out. Added a null check so we don't hide the shelf until we actually know which tab is selected.

Closes #

### Additional context

N/A

### Test results


ILoveOpenSourceApplications:
The changes you've made are based off a really old build. The horizontal shelves filter are in a file of its own.

utafrali:
ah got it, didn't realize the structure changed - what file is it in now? I'll update the changes

ILoveOpenSourceApplications:
See #930.

utafrali:
got it, I'll check #930 and update the PR

ILoveOpenSourceApplications:
The changes are done with the help of an LLM, and since you're not giving enough context, it's not actually doing anything good. 

Are you even testing the changes?

utafrali:
My bad, didn't properly test this. Closing it.

Issue #587, opened 2026-02-19
bug(YouTube): PiP background playback black bar


adriel806:
### Bug description

When I watch a video in PiP video, the video is cut in half by a black bar. I'm using patch 1.14 and  the recommended version.

![Image](https://github.com/user-attachments/assets/95775e3b-27f6-4b15-b0fa-8c2335db8c99)

. 

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

LisoUseInAIKyrios:
What version YouTube did you patch?

Did you turn on `Disable translucent status bar` or `Disable translucent navigation bar`?

adriel806:
> What version YouTube did you patch?
> 
> Did you turn on `Disable translucent status bar` or `Disable translucent navigation bar`?

The translucent status bar and navigation bar are disabled. The version I'm using is 20.40.45.

LisoUseInAIKyrios:
Sorry, is translucent status/bar active? Are they translucent or not translucent?


Disabling translucent can cause these black bars in the player.

adriel806:
![IMG_20260219_091839.jpg](https://github.com/user-attachments/assets/5824332e-81f7-462b-ae47-686e35781273)

LisoUseInAIKyrios:
Are you using RVX?

`Enable translucent navigation bar` is a RVX feature and is not the same as the Morphe setting that disables but doesn't force enable.

adriel806:
I thought I was using Morphe WTF 😆 sorry about that

adriel806:
> Sorry, is translucent status/bar active? Are they translucent or not translucent?
> 
> 
> Disabling translucent can cause these black bars in the player.

Enabling or disabling the translucent navigation bar has no effect. The navigation bar remains opaque... in both night and light mode.

LisoUseInAIKyrios:
The feature isn't available on all devices, and YouTube controls who gets it.

If you've been version spoofing or patching old versions of YT then it seems less likely the feature is enabled by YT.

Forcing the translucent on causes issues for some devices, including crashing with old android releases.

PR #1173, opened 2026-04-13
fix(YouTube): Background playback PiP shows a black bar after using livestream chat

FlaminSarge:
### Description

Long-standing YT bug causes a white or black bar to appear on the PIP player if the user ever interacts with the text entry box on livestream chat e.g. https://www.reddit.com/r/youtube/comments/1lkt6ek/ https://www.reddit.com/r/youtube/comments/1jrr3ty
This hides that bar when PIP is activated.

Once again, patch name/language/description/etc up for debate.

Closes #

### Additional context

navigationBarBackground is the culprit, apparently YT doesn't properly hide it when text entry is dismissed.
Initially I planned to restore the visibility value on it, but it seemed like YT handled it properly on its own so I went with just the simplest way of hiding it when PIP is activated and letting YT handle it when it comes back into focus.
Initial patch was for PipController's onPipModeChanged, but that involved some jank to get the activity out of the PipController object, and moving it up to MainActivity seemed to work fine.
There's another overload of onPictureInPictureModeChanged that might need a patch as well but that might only be used in very old versions of YT.

FAIR WARNING: Once again, decompile fed to Claude Code to figure out where possible culprits were, though ultimately it didn't actually find it and I had to go digging through the window views to find that navBarBackground was exactly 126px tall.

### Test results
Tested on 20.45.36, 21.14.482, 20.21.37 


patches/src/main/kotlin/app/morphe/patches/youtube/misc/fix/pipchatbar/FixPipChatBarPatch.kt
```diff
@@ -0,0 +1,26 @@
...
+    name = "Fix PiP chat bar",
+    description = "Hides the bar that appears over the PiP video after using live chat text entry.",
+) {
+    dependsOn(sharedExtensionPatch)
+
+    compatibleWith(COMPATIBILITY_YOUTUBE)
+
+    execute {
+        PipModeChangedFingerprint.method.addInstruction(
+            1,
```
FlaminSarge: On 21.x this index shouldn't be more than like 3, the p0 register gets clobbered after that point. This may also work at index 0, anyways.

LisoUseInAIKyrios: If it works at index 0 that's ideal.

These changes could be part of `Remove background playback restrictions` as it has other background playback fixes for bugs in unpatched YT.

FlaminSarge: Not entirely on board with grouping it with that, since if someone were to look through the patch list they wouldn't really be able to tell that `Background playback fixes` would include PiP fixes.

Definitely think it needs a rename/package move though, "FixPipChatBar" was the best I could come up with but definitely isn't ideal since it implies a 'chat bar' and not 'a bugged bar because you opened chat'.

LisoUseInAIKyrios: Now I remember. The fixes are not part of background playback patch but instead a dependent patch of settings, which is the simplest way to have mandatory patches so users don't exclude these fixes then report bugs the patches fix.


This patch can be the same. Remove the patch name and add it as a dependency like the other fixes.

https://github.com/MorpheApp/morphe-patches/blob/07d416a5fee8f502a538e513e41e0319b7fbc9dc/patches/src/main/kotlin/app/morphe/patches/youtube/misc/settings/SettingsPatch.kt#L209-L212

FlaminSarge: Done, but patch name still iffy for me.

LisoUseInAIKyrios: Name looks ok to me.

Will merge tomorrow. Many thanks.

FlaminSarge:
Fixes #587

tyoung2221:
If I use one of the 3 versions you tested the fix with, then will it still work if I proceed to spoof the app version? For example using 20.45 but then spoofing to 20.37?

PR #1175, opened 2026-04-13
fix(Settings): Bypass click debounce for toggleable preferences

ILoveOpenSourceApplications:
### Description

N/A

### Additional context

The debounce logic originally added to prevent double-opening of sub-menus and dialogs was inadvertently affecting `SwitchPreference` and `CheckBoxPreference` items. This caused them to feel unresponsive when toggled rapidly. 

This fix retrieves the clicked item from the adapter and conditionally bypasses the debounce check if the item is a `TwoStatePreference`, restoring smooth toggle interactions while keeping the double-click protection intact for standard list items.

### Test results

### DIFF
--- a/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/settings/BaseActivityHook.java
+++ b/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/settings/BaseActivityHook.java
@@ -50,6 +50,12 @@ public abstract class BaseActivityHook extends Activity {
     }
 
     /**
+     * Set this to true in the future if stock YouTube ever fully implements
+     * Android's predictive back gestures for its settings menus.
+     */
+    private static final boolean ENABLE_PREDICTIVE_BACK_ANIMATION = false;
+
+    /**
      * Initializes the activity by setting the theme, content view and injecting a PreferenceFragment.
      */
     public static void initialize(BaseActivityHook hook, Activity activity) {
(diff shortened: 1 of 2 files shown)


LisoUseInAIKyrios:
What preferences did this problem originally affect?

Was it just preference screens or some other specific preference?

ILoveOpenSourceApplications:
Preference screens and custom dialogs would create duplicates if clicked twice in quick successions.

<p>

https://github.com/user-attachments/assets/f9cf24e5-b5e7-4bc2-b1f6-7a84654fa808

</p>
</details>

LisoUseInAIKyrios:
Is it just preference screens? If so then the denounce fix could be selectively applied to just preference screen instances.

Edit: I just noticed it's also presences that open dialogs. So selectively applying may not be simple.

ILoveOpenSourceApplications:
Yeah, because of this, it's much safer to apply the debounce globally and explicitly opt-out `TwoStatePreference`. Toggleable preferences are practically the only items in a settings list where rapid, repeated user clicks are expected and don't push new views to the screen.

Domiiniik:
BTW its also worth to mention that there is this bug where "Morphe Settings" tab is the only tab having predictive back enabled when they shouldn't (stock youtube settings AND Morphe settings categories tabs don't have it) so "Morphe Settings" also shouldn't for consistency (YTM doesn't have this problem either)

https://github.com/user-attachments/assets/ae00e4ae-39ac-4e94-a599-bf4f6a09b607

ILoveOpenSourceApplications:
That's because Morphe is injected in the same way how Settings opens for YouTube, not like one of its sub menus like General. You can check see that both of those have predictive back gestures.

Edit: I spoke too soon.

PR #1187, opened 2026-04-15
feat(YouTube - Hide layout components): Add "Hide Settings button" setting for player

### LINKED ISSUE #1181: feat(YouTube): Add an option to hide the gear icon in player
### Feature description

Option that hides the gear icon which apperas in top-right corner next to the captions button in the video player.

We can hide the player flyout menu items, but not the gear icon itself.

### Motivation

Because depending on use-case it may not be needed. For instance I only really care about the captions button, which can be used outside the gear icon, and the options in the gear icon I don't need to change because other patches already do it for me (e.g. playback speed and video quality).

### Acknowledgements

ILoveOpenSourceApplications:
### Description

Closes #1181.

### Additional context

N/A

### Test results


extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/DescriptionComponentsFilter.java
```diff
@@ -82,6 +82,11 @@ public DescriptionComponentsFilter() {
                 "hype_points_factoid"
         );
 
+        final StringFilterGroup videoDetails = new StringFilterGroup(
+                Settings.HIDE_VIDEO_DETAILS_SECTION,
+                "linear_layout.e"
```
LisoUseInAIKyrios: This seems very generic and may hide unrelated stuff.

Is there anything else in the path of buffer that this can also search for?

ILoveOpenSourceApplications: All the components sounds generic for this one like `section_header` which will only hides the header, `list_item` which hides the content, and then there's `linear_layout` which hides it entirely.

LisoUseInAIKyrios: Yes, but `linear_layout` is an android layout container, and it's possible other places may use this.

It might be better to add a buffer filter with `section_header.e` to reduce the chance of a false positive.

ILoveOpenSourceApplications: Sure, that works.

ILoveOpenSourceApplications: Is the current one better or is the below one better?
```
        if (matchedGroup == videoDetails) {
            return path.contains("section_header.e") && path.contains("list_item.e");
        }
```

Issue #1001, opened 2026-03-24
bug(YouTube Music): Setting start page to Charts while navigation bar is hidden, makes going to the settings impossible


ghost:
### Bug description

There is no possible way to go the settings while start page is set to Charts and navigation bar is hidden. Clicking back button in the top left corner in Charts window minimizes the app (like double click back button does). And Charts window doesnt have any way to open up profile, which would also allow to open up settings.

As I see it, user is forced to clear app data and start over, because there is no possible way to open up YouTube Music or Morphe settings when start page is Charts and navigation bar is hidden.

Also when start page is set to Charts, and navigation bar is hidden. It will make YouTube Music really limited. You cant go to the Home Page or your Library, or any other page. You can only subscribe to channels, and search.

Should there be a warning about this in the Strings? This is not a bug per se, more of a unfortunate combination of settings that breaks usage.

### App version patched

9.11.54

### Error logs

```shell
No errors.
```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

ILoveOpenSourceApplications:
> As I see it, user is forced to clear app data and start over, because there is no possible way to open up YouTube or Morphe settings when start page is Charts and navigation bar is hidden.

Repatch the app again, excluding either `Change start page` or `Navigation bar` patch and change the values in "Import/Export" setting. No need to clear your app data necessarily.

ghost:
> > As I see it, user is forced to clear app data and start over, because there is no possible way to open up YouTube or Morphe settings when start page is Charts and navigation bar is hidden.
> 
> Repatch the app again, excluding either `Change start page` or `Navigation bar` patch and change the values in "Import/Export" setting. No need to clear your app data necessarily.

Ahh yes, this would be a workaround.

ILoveOpenSourceApplications:
The back button in YouTube takes the user back to Home feed, the same doesn't happen in YouTube Music. I'll look into this so that pressing the back button takes the user back to Home feed in YouTube Music.

ghost:
> The back button in YouTube takes the user back to Home feed, the same doesn't happen in YouTube Music. I'll look into this so that pressing the back button takes the user back to Home feed in YouTube Music.

It would be great if it had the same method to go to the Home feed as YouTube has.

ILoveOpenSourceApplications:
This is intended as the start page behavior should be identical to that of Home page when it's the default start page.

PR #1195, opened 2026-04-16
refactor(Change start page): Extract common Java code to `shared-youtube` module and some cleanup

ILoveOpenSourceApplications:
### Description

N/A

### Additional context

N/A

### Test results


LisoUseInAIKyrios:
The patch was crashing on startup with 9.15 (back button register is not the expected type), and after fixing the crash the patch doesn't seem to work if `Always change start page` is turned off.

Patch doesn't seem to always change start page either. I'm not sure if the problem is fixing 9.15 or if something else is going on.

LisoUseInAIKyrios:
Neither `overrideIntentActionOnCreate` or `overrideBrowseId` are being called when the app is resumed.

ILoveOpenSourceApplications:
I made the changes based off of the patch for YouTube. I guess it still needs more testing.

ILoveOpenSourceApplications:
"`WARNING: Track crossfade is not yet available for YouTube Music 9.x. Patch YouTube Music 8.44.54ΓÇô8.50.51 for crossfade.`"

LisoUseInAIKyrios:
The crossfade warning is correct, it doesn't support 7.x or 9.x

ILoveOpenSourceApplications:
I don't know if this issue was present back then or not, but if the start page is "Episodes for later" or "Liked music", pressing back button takes the user to a blank page.

LisoUseInAIKyrios:
If those are problematic, then can either remove those start page types or selectively not apply the back button change to those start pages.

ILoveOpenSourceApplications:
Also, are these issues present for only android 13 and above or is there for those below as well?

ILoveOpenSourceApplications:
Thinks are looking good now finally. @LisoUseInAIKyrios, can you check from your side as well?

ILoveOpenSourceApplications:
What does "Always change start page" do actually?

LisoUseInAIKyrios:
If the app is closed, then reopened, the start page is applied again.

Edit: This no longer seems true anymore even for YouTube.

ILoveOpenSourceApplications:
So should that option be removed? Or does it work on minimum version of YouTube?

LisoUseInAIKyrios:
Need to look into what is going on with YouTube, because the feature is useful.

ILoveOpenSourceApplications:
I still don't fully understand what it does when turned on vs off as I don't personally use the patch, so please do the needful.

LisoUseInAIKyrios:
20.21.37 doesn't call the existing startup hooks so opening an already running app from the launcher does not open the desired start page.

But using "always change startup page" does fix the back button navigating back to home screen.

I'm not sure if the side effect of always changing the start page is still relevant and I haven't been able to reproduce the "back button may not work on the toolbar" side effect anymore.

I think we can try removing the "always override back button" setting and force it always on.

Later in a different PR we can fix the start page not changing when using launcher and the app is already running.

ILoveOpenSourceApplications:
Sure, the scope of this PR was to fix back button not taking to home in YTM, and that's fixed so the other issue can be the scope of another PR.

LisoUseInAIKyrios:
Odd, but YT Music search start page is not always applied. Sometimes it still uses the home page on a cold launch.

LisoUseInAIKyrios:
@ILoveOpenSourceApplications Please double check the current code. With the always override now always on, I'm not sure if any of the code I removed is still needed.

For me, the app behaves how it should where opening the app with a different start page and pressing back button exits the app (and does not navigate to the home screen). This is the same behavior as YouTube and logically is correct because pressing back on the default start page exits the app, and allows hiding the home feed entirely.

ILoveOpenSourceApplications:
Episodes and Liked is opening when exiting using the back button/gesture despite not being the start page. Need to look into that.

ILoveOpenSourceApplications:
Last time I checked it was https://github.com/MorpheApp/morphe-patches/issues/1001#issuecomment-4125141062 for YouTube as well. Are you sure that pressing back button should exit the app in intended?

LisoUseInAIKyrios:
When using default start page, opening the app then pressing back button exits the app. 

The same behavior should be present if the start page is changed to something different.

ILoveOpenSourceApplications:
It wasn't doing that the last time I checked, or not atleast for every start page available.

LisoUseInAIKyrios:
YT back button would return to home feed, if "always override" was off. But always override can be always on because the toolbar back button issue doesn't seem to happen anymore.

ILoveOpenSourceApplications:
Back button within YT just keeps looping back into the same page.

LisoUseInAIKyrios:
Which start page does that happen with?

ILoveOpenSourceApplications:
Any page with a back button on the top left.

ILoveOpenSourceApplications:
YTM pages "Liked Music" and "Episodes for later" are once again broken.

LisoUseInAIKyrios:
Calling `activity.finish()` breaks cold launching the app if the start page is search.

ILoveOpenSourceApplications:
@inotia00, hey, when you have some free time would you mind taking a look at this? Some start pages no matter what I do won't exit when pressing the in-app back button. I guess there's two approach that can be taken here, one is to fix it properly, second is to hide the back button of whatever start page that has it. The default device back button always exits the app just like how its supposed to be.

ILoveOpenSourceApplications:
Went from a fix to a refactor.

MarcaDian:
Please do not touch `Settings.java`, the formatting is intentionally done in multiple lines for convenience.

LisoUseInAIKyrios (approved):

PR #1208, opened 2026-04-17
fix(YouTube - Hide layout components): Hide preview comment dots

0xrxL:
### Description

Screenshot_20260417_235807

### Additional context

These dots are a part of a touch selector that works only when the preview comment is visible. So, I added a way to hide them when also preview comment is hidden.

### Test results


0xrxL:
@LisoUseInAIKyrios wchill tried and it works. Give a little code check and than merge. 👍

PR #1243, opened 2026-04-20
fix(YouTube Music - Track crossfade): Stop playback when app is closed and improvements to auto-play crossfade transitions

### LINKED ISSUE #1203: bug(YouTube Music): App keeps playing sound even after closing the app
### Bug description

Ever since I installed the patch that enables track transitions, when I close the app, the music keeps playing, and when I reopen it, the player (and the mini player) no longer appear. I have to force-close the app to stop the sound.

### App version patched

8.51.51

### Error logs

```shell

```

### Acknowledgements

VazerOG:
### Description

  Fixes three issues with the Track crossfade patch:

  **1. Music keeps playing after app is closed**
  Added `onActivityStop()` / `onActivityStart()` lifecycle hooks injected into `MusicActivity`. When the activity stops, any active crossfade is aborted and the auto-advance
  monitor is halted. Playback resumes correctly when the activity restarts.

  **2. Auto-advance fade-out not perceptible**
  The outgoing track was staying at or near full volume when a track naturally ended into the next. Root cause: `onPendingPlayerReady` was applying the full configured fade
  duration regardless of how much audio the outgoing track had left. Variable new-player READY latency (100–500ms+) meant the fade could start too late to complete before the
  track ended. Fixed by dynamically shortening the fade-out duration to match actual remaining audio (`max(150ms, actualRemaining)`) when it is less than the configured fade
  duration.

  **3. Incoming track volume blip at transition start**
  On auto-advance, the native `playNextInQueue` path calls `MedialibPlayer.loadVideo(... volume=1.0 ...)`, which reset the new player's volume to 1.0 before our crossfade
  animation began — causing a brief full-volume blip from the incoming track. Fixed by changing `onBeforePlayNext` from a void hook to a blocking boolean hook: the original call
  is blocked, crossfade state is set up, native is re-invoked in a controlled window (via `internalPlayNext` flag), and volume is synchronously re-enforced to 0 before polling
  begins.

  Closes #1203

  ### Additional context

  Tested on the following versions:
  - **8.44.54** — PASS (minimum supported)
  - **8.47.56** — PASS (experimental)
  - **8.50.51** — PASS (experimental)

  9.x versions are not supported by this patch and were not tested.

  ### Test results


extensions/music/src/main/java/app/morphe/extension/music/patches/CrossfadeManager.java
```diff
@@ -718,6 +760,11 @@ public void run() {
                 if (!crossfadeInProgress) return;
                 if (newPlayer != pendingInPlayer) return;
 
+                // Keep new player silent while waiting for READY. The native
+                // playNextInQueue (auto-advance) runs after our void hook and
+                // resets the player to volume 1.0 — re-enforce on every tick.
+                try { newPlayer.patch_setVolume(0.0f); } catch (Exception ignored) {}
```
LisoUseInAIKyrios: These catch blocks can log to debug, just in case something interesting happens.

patches/src/main/kotlin/app/morphe/patches/music/interaction/crossfade/CrossfadePatch.kt
```diff
@@ -273,6 +279,22 @@ val crossfadePatch = bytecodePatch(
             """,
         )
 
+        val musicActivityClass = mutableClassDefBy(MusicActivityOnCreateFingerprint.classDef.type)
```
LisoUseInAIKyrios:
```suggestion
        val musicActivityClass = MusicActivityOnCreateFingerprint.classDef
```

sr-shishir:
Expecting that fix. Cause crossfade was not working functionally on v8.47.56. I used this settings & the songs were not Cross fading properly. Every single song ends to it’s last & then go to the next one.

VazerOG (commented):
Looks good.

LisoUseInAIKyrios:
Will merge tomorrow if there's no new concerns.

extensions/music/src/main/java/app/morphe/extension/music/patches/CrossfadeManager.java
```diff
@@ -678,7 +708,7 @@ public static boolean onPauseVideo() {
             return false;
         }
 
-        logInfo("onPauseVideo during crossfade — aborting crossfade, allowing pause");
+        Logger.printDebug(() -> "onPauseVideo during crossfade — aborting crossfade, allowing pause");
         abortCrossfadeNow();
         return false;
     }
```
LisoUseInAIKyrios, line 710: This method always returns false, but the patches smali has logic for both true and false return values.

Is there a typo here and one of these paths needs to return true?

ILoveOpenSourceApplications:
I did a bunch of cleanup, please check if everything works fine. Also, if some of the comments removed are actually useful and not AI generating for context, do restore them as required.

extensions/music/src/main/java/app/morphe/extension/music/patches/CrossfadeManager.java
VazerOG: I am at Google Next '26 in Las Vegas right now. I will look at this as soon as I can. I suspect it's a typo.

VazerOG:
Just tested out everything. I like the requirement to restart the app after enabling Crossfade in Morphe > Player > Crossfade. 

Tested on: 
- 8.44.54
- 8.47.56
- 8.50.51

Crossfade functionality is perfect.

9.x: 
Still struggling with the backend players and the frontend UI state getting disconnected... will take more time to figure out. 

This can be pushed as-is if there are no other concerns from anyone else.

LandonMoran:
Why is this one here cuz it was another PR with the same kind of name that was here before

LisoUseInAIKyrios:
@LandonMoran this is a fix, not a new feature.

extensions/music/src/main/java/app/morphe/extension/music/patches/CrossfadeManager.java
LisoUseInAIKyrios: Can this be ignored for now?

VazerOG: Yes. Ignore for now.

Issue #1018, opened 2026-03-26
feat(YouTube): Add options to hide fullscreen overlay individual action buttons


helel-ben-shahar:
### Bug description

After turning off action button like ask, clip, remix etc. from settings it's still showing in landscape mode but hidden in portrait mode. 

![Image](https://github.com/user-attachments/assets/14afce06-f6cd-4224-80c1-d9fce1dae500)
![Image](https://github.com/user-attachments/assets/e3a4b9b7-b2ea-4919-abf9-88b7b9c246e8)

### App version patched

YouTube Version: 20.45.36

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

ILoveOpenSourceApplications:
Enabling "Hide quick actions" will hide the entire bottom area, but currently there's not way to individually hide those buttons.

l205306:
https://github.com/MorpheApp/morphe-patches/issues/880

kamik-hub:
We can individually hide icons in RVX btw
![Screenshot_2026-03-26-21-58-17-545_app.rvx.android.youtube.jpg](https://github.com/user-attachments/assets/ff9c7e47-af30-4ac3-a33e-dd11d3b33b5f)

BlackGold8282:
@ILoveOpenSourceApplications 

https://github.com/inotia00/revanced-patches/blob/dev/extensions/shared/src/main/java/app/revanced/extension/youtube/patches/components/QuickActionFilter.java

iBasim:
In Morphe  "Action buttons" under Player is merged into one setting.
I want to split it into two separate options like in RVX (see attached image)
1- Hide/show action buttons under video (non-fullscreen).
2- Hide/show action buttons in fullscreen mode.

Reason: Some buttons I want to hide only in fullscreen mode; others only under videos (non-fullscreen).

<img width="1440" height="2466" alt="Image" src="https://github.com/user-attachments/assets/24723fa7-2828-4556-bd99-2950e7c08b10" />

PR #1264, opened 2026-04-22
refactor(YouTube - Hide video action buttons): Group player action button settings together into a single preference screen

### LINKED ISSUE #1262: bug(YouTube): "Hide Quick actions" setting is not working
### Bug description

In the 'player->Quick actions->the feature does not hiding all the components properly.......look at the photo

I'm using the latest 1.25.0 dev1 patches btw

Image
Image

### App version patched

21.16.245

### Error logs

```shell

```

### Acknowledgements

ILoveOpenSourceApplications:
### Description

Closes #1262

### Additional context

N/A

### Test results

### DIFF
--- a/patches/src/main/resources/addresources/values/youtube/strings.xml
+++ b/patches/src/main/resources/addresources/values/youtube/strings.xml
@@ -636,8 +636,9 @@ Adjust volume by swiping vertically on the right side of the screen"</string>
     <string name="morphe_swipe_change_video_summary_off">Swiping in fullscreen mode will not change to the next/previous video</string>
 
     <!-- layout.buttons.action.hideButtonsPatch -->
-    <string name="morphe_hide_buttons_screen_title">Action buttons</string>
-    <string name="morphe_hide_buttons_screen_summary">Hide or show action buttons under videos</string>
+    <string name="morphe_action_buttons_screen_title">Action buttons</string>
+    <string name="morphe_action_buttons_screen_summary">Hide or show action buttons in fullscreen and portrait</string>
+    <string name="morphe_portrait_buttons">Portrait buttons</string>
     <string name="morphe_disable_like_subscribe_glow_title">Disable Like and Subscribe glow</string>
     <string name="morphe_disable_like_subscribe_glow_summary_on">Like and Subscribe button will not glow when mentioned</string>
     <string name="morphe_disable_like_subscribe_glow_summary_off">Like and Subscribe button will glow when mentioned</string>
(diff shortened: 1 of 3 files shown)


patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -942,7 +942,7 @@ To show the Audio track menu, change \'Spoof video streams\' to \'Android Reel\'
     <string name="morphe_hide_settings_button_summary_off">Settings button is shown</string>
 
     <string name="morphe_quick_actions_screen_title">Quick actions</string>
-    <string name="morphe_quick_actions_screen_summary">Hide or show fullscreen quick actions related components</string>
+    <string name="morphe_quick_actions_screen_summary">Hide or show quick actions in fullscreen</string>
```
LisoUseInAIKyrios: Not specific to this PR, but is the name "quick actions" used by YouTube?

Except for the recommended video, these are the same as the portrait player action buttons.

Is something like "Fullscreen action buttons" better?

ILoveOpenSourceApplications: It's called `quick_actions` in litho and I think that's where the name came from. Are there any official documents on what else these are called?

LisoUseInAIKyrios: As far as I can find, YouTube does not have any name for these buttons in general.

If we wanted to make the settings simpler, these hide settings could be consolidating into the player action button settings (and move the fullscreen 'more video' somewhere else). Because I'm not sure if someone would want to hide the same button in one mode but not the other (portrait vs fullscreen). Less settings is desirable to keep things simpler.

ILoveOpenSourceApplications: Could do that, yes. But https://github.com/MorpheApp/morphe-patches/issues/1018#issuecomment-4232910441. Users love the extra options

LisoUseInAIKyrios: At the expense of increasing the "what do all these settings do?"

I think we can rename this to something like "Fullscreen action buttons" and group both preference screens together in the player menu.

ILoveOpenSourceApplications: A preference category?

LisoUseInAIKyrios: Either two preference categories in the same "Action buttons" preference screen (one category for portrait, another for fullscreen).

Or two different preference screens grouped together using a no title preference group so they appear together in the Player menu.

ILoveOpenSourceApplications: <details><summary>Details</summary>
<p>

image
image


</p>
</details>

ILoveOpenSourceApplications: So this is good?

PR #1269, opened 2026-04-23
fix(YouTube - Open Shorts in regular player): Back button does not work when viewing Shorts in the regular player

0xrxL:
Current implementation doesn't work for 21.xx versions


LisoUseInAIKyrios:
This needs an adjustment because if open Shorts in regular player is off the behavior is different:

21.17.480, Lpzq;

Before
```
        if (pycVar.o) {
            fzVar.finish();
        } else {
            ...
        }
```

After
```
        boolean z2 = pycVar.o;
        if (OpenShortsInRegularPlayerPatch.overrideBackPressToExit()) {
            fzVar.finish();
        } else {
          ...
        }
```

0xrxL:
I totally overturned the structure of the patch, to avoid losing mind by adapting the patch for every little change implemented by google over all the compatible versions. Now the patch can be applied dinamically and the logic is more small.

LisoUseInAIKyrios:
@0xrxL Try again with the latest. The bytecode changes appear correct for all supported versions.

There seems to be a different issue with 21.17 where the fullscreen button sometimes does not do anything when a Short is opened in fullscreen mode (nothing happens when pressed by the user or artificially clicked to exit fullscreen mode), but this has nothing to do with the changes here.

0xrxL:
> @0xrxL Try again with the latest. The bytecode changes appear correct for all supported versions.
> 
> There seems to be a different issue with 21.17 where the fullscreen button sometimes does not do anything when a Short is opened in fullscreen mode (nothing happens when pressed by the user or artificially clicked to exit fullscreen mode), but this has nothing to do with the changes here.

Tested and works on 21.17. You can merge it. 👍

PR #1274, opened 2026-04-23
feat(YouTube): Add `Save to watch later` patch

### LINKED ISSUE #18: feat(YouTube): Add a custom player Save button
### Feature description

Add a Save button as overlay button on video player.

### Motivation

Recently with the adoption of new the compact action buttons layout (the one with the channel avatar behind like/dislike button), the list of action buttons fetched with YouTube API has been made definitively non-functional.

In brief: now Google apply a different layout for every video where: Clip, Remix and Save buttons aren't placed at the same index position for every video and, from the moment that Remix button doesn't appear inside the aformentioned list (must be added manually, as already documented limitation by @inotia00), the filtering is obviously broken.

This lead me to take the decision to perform the filtering of all buttons, with the exceptions of first three (channel avatar, like/dislike, share), and add the save button directly on player, to also circumvent the limitation of Saving videos marked as kids in a playlist (this at least on uDrop).

The unique problem is to find the logic to save a video inside a playlist. For this reason I chose to open this feature request (maybe @inotia00 already found the way with one of his patches).

### Acknowledgements

0xrxL:
Based on @inotia00 code

Closes: https://github.com/MorpheApp/morphe-patches/issues/18

This patch needs of a new class to extend the capacities of current client stream fetching, in a way to be able to get other types of data from a video ID, like: a playlist, a creator channel, name of available audio tracks and more.

For the moment I kept only two of these, that represents two future patches:

- A button to save a video in watch later playlist.
- A mod to redirect the circle live avatar to the creator channel instead to the current live video.

The last patch will be added later, once this PR will be correctly refactored and merged.


0xrxL:
@MarcaDian @LisoUseInAIKyrios @ILoveOpenSourceApplications 👍👍👍

0xrxL:
P.S: Need to be refactored, because I added the logic in the method of copyURL button.

delt4-haze (commented):
👌

0xrxL:
@LisoUseInAIKyrios Any ETA for abstraction? I had planned to add the second PR this week, if you would be able to merging this one before.

extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/spoof/requests/VideoDetailsRequest.java
```diff
@@ -0,0 +1,193 @@
...
+
+                                String jsonBuilderString = jsonBuilder.toString();
+                                JSONObject jsonResponse = new JSONObject(jsonBuilderString);
+
+                                switch (infoToFetch) {
+                                    case "channelID" -> {
+                                        return jsonResponse
+                                                .getJSONObject("videoDetails")
+                                                .getString("channelId");
+                                    }
```
fromhgbwithluv: The `channelId` is accessiable in VideoInformation.getChannelId(), so why we need to fetch it from youtube api

LisoUseInAIKyrios: Yes. Code can be consolidated. @0xrxL I will work on this in the next few days.

0xrxL: > The `channelId` is accessiable in VideoInformation.getChannelId(), so why we need to fetch it from youtube api

Because I needs of it to override the behaviour of channel avatar in feed, when it redirect to the current live video instead to the creator channel.

delt4-haze: ​Really appreciate the attention to detail here. That fix for the channel avatar redirect is a lifesaver. Thanks for working on this!

LandonMoran: Wasn't there one an old YouTube quite a while ago?

0xrxL: > Wasn't there one an old YouTube quite a while ago?

Only in RVX, originally shared by me to Inotia (I'm referring only to the live avatar patch).

0xrxL:
I did my best to refactor the code by avoiding duplicates. I don't know if I managed to achieve this goal, but I tried.

Anyway the new code is fully functional. Only a good icon for the button is missing (it currently still use the one from reload button).

0xrxL:
This PR needs more refactors here and there. Not major things, only some methods/classes names and debug stuff.

0xrxL:
ThreadsPoolExecutor's threads amount must be increased to 4 or 5.

LisoUseInAIKyrios:
The thread pool executor has no thread limit. The min size is 3 but that's not strictly required and is only to reduce thread creation cycles from RYD/SB/Spoofing that is called on every new video.

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/SaveToWatchLaterPatch.java
```diff
@@ -0,0 +1,86 @@
...
+                    } else {
+                        Utils.showToastShort(str("morphe_save_to_watch_later_already_exists_toast"));
+                    }
+                }
+
+                Logger.printInfo(() -> saveToWatchLaterResponse);
+            }
+        } catch (Exception ignore) {
+            Utils.showToastShort(str("morphe_save_to_watch_later_error_toast"));
+        }
```
LisoUseInAIKyrios: Is this a normal scenario? If this should not happen this can be an exception log message.

0xrxL: Yeah it's normal. Because if the API can't be reached for whatever reason (temporary missing network for example), that's show you that error toast.

LisoUseInAIKyrios:
Seems ready. Will merge later today if there's no new concerns.

0xrxL:
> Seems ready. Will merge later today if there's no new concerns.

Check if this PR also works for you (just to be 100% safe).

LisoUseInAIKyrios:
Yes I'm using it now and it works great.

extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/spoof/requests/PlayerRoutes.java
```diff
@@ -64,28 +86,47 @@ static String createInnertubeBody(ClientType clientType, String videoId) {
...
+            }
+            String platform = clientDevice.getClientPlatform();
+            if (platform != null && !platform.isEmpty()) {
+                client.put("platform", platform);
             }
-            if (clientType.clientPlatform != null) {
-                client.put("platform", clientType.clientPlatform);
+            JSONObject user = new JSONObject();
+            user.put("lockedSafetyMode", false);
+            if (detailsToFetch != null && detailsToFetch.isEmpty()) {
```
LisoUseInAIKyrios: Should this be:
`if (detailsToFetch != null && !detailsToFetch.isEmpty()) {`

0xrxL: No, because language parameters should not be included in video detail requests.

So the check is right, and allow language only for stream.

LisoUseInAIKyrios: But language is set only if the details to fetch is both not null and it's empty.

Is that right? What values are passed in here?

0xrxL: > But language is set only if the details to fetch is both not null and it's empty.
> 
> Is that right? What values are passed in here?

detailsToFetch is empty for stream requests and filled with "getChannelID" or "saveToWatchLater" string for details requests, based on which details you want to fetch (the channelID to override live avatar behaviour or saveToWatchLater button request to save the video in the watch later playlist).

Otherwise you can also change the condition of check, to verify if clientTypeDetails is not null, instead if detailsToFetch is empty. In both cases are empty/null values when stream requests are performed.

0xrxL:
@LisoUseInAIKyrios I was even thinking about a refactor of these two fields. These can be converted in an Enum, instead to be kept as strings. What do you think?

<details>

Screenshot_20260506_111346_Edge
</details>

LisoUseInAIKyrios:
Yes an enum makes sense there.

LisoUseInAIKyrios:
Seems ready. Will merge tomorrow.

0xrxL:
> Seems ready. Will merge tomorrow.

As usual: check if the button still works for you after this refactor. 👍

extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/spoof/requests/PlayerRoutes.java
```diff
@@ -133,14 +170,18 @@ static String createInnertubeBody(ClientType clientType, String videoId) {
...
+            String clientVersion) throws IOException {
         var connection = Requester.getConnectionFromCompiledRoute(YT_API_URL, route);
 
         connection.setRequestProperty("Content-Type", "application/json");
-        connection.setRequestProperty("User-Agent", clientType.userAgent);
+        connection.setRequestProperty("User-Agent", userAgent);
         // Not a typo. "Client-Name" uses the client type id.
-        connection.setRequestProperty("X-YouTube-Client-Name", String.valueOf(clientType.id));
-        connection.setRequestProperty("X-YouTube-Client-Version", clientType.clientVersion);
+        connection.setRequestProperty("X-YouTube-Client-Name", clientName);
```
inotia00, line 183: There was a typo starting from this commit..

Valid instructions:
```
// Not a typo. "Client-Name" uses the client type id.
connection.setRequestProperty("X-YouTube-Client-Name", String.valueOf(clientType.id));
```

Invalid instructions:
```
// Not a typo. "Client-Name" uses the client type id.
connection.setRequestProperty("X-YouTube-Client-Name", clientName);
```

**Update**: Fixed in http://github.com/MorpheApp/morphe-patches/pull/2094/commits/b51290fe96ec17e45d27109ee4c2ced6013c49f3.

PR #1301, opened 2026-04-26
feat(YouTube - Sponsorblock): Add highlight segment submission

### LINKED ISSUE #1100: feat(YouTube): Add support to create/submit SponsorBlock highlights
### Feature description

Would allow for submitting highlights for sponsorblock. 

Submitting highlights would go based on whatever part you marked as start/end, for example if you just marked the start, the highlight would be the "start", and vice versa. When clicking the submit button here: 
Image without marking 2 locations, it would assume you're submitting a highlight, and would show 

 "Is the highlight timestamp correct?
xx:xx:xx:xxx
If you're not submitting a highlight, tap on "cancel" and mark the second location.

`Cancel`     `Submit`" 

If the user marks both start and end, it'll ask the user which timestamp is the highlight.

### Motivation

This would allow for submitting all segments

### Acknowledgements

MarcaDian:
Users can now submit SponsorBlock highlight (`poi_highlight`) segments directly from the new segment UI.
### How it works:

- One location marked - tapping `Publish` shows a confirmation dialog with that timestamp and an option to submit it as a highlight, or `Cancel` to mark the second location first.
- If both a start and end are selected, then choosing 'Highlight' from the category selection dialog will use the start time as the highlight.

Highlight segments are submitted with `category=poi_highlight` and `actionType=poi`, with `startTime == endTime` as required by the SponsorBlock API.
___


MarcaDian:
I'm pretty sure everything was working before the last change.
But I'm not sure if I broke anything by moving everything to `CustomDialog`.
Because I don't use segment creation or highlights.

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -1444,6 +1444,21 @@ Ready to submit?"</string>
...
+
+%s
+
+If you\'re not submitting a highlight, tap Cancel and mark the second location"</string>
+    <string name="morphe_sb_new_segment_highlight_choose_timestamp">"Which timestamp is the highlight?
+
+Start: %1$s
+End: %2$s
+
+If you\'re not submitting a highlight, tap Cancel and preview the segment first"</string>
```
LisoUseInAIKyrios: I think if the user selects both a start and end, to prevent accidentally submitting a segment as a highlight it can prompt to preview as it previously has, otherwise users may submit non highlights as highlights because they couldn't figure out how to preview the segment.

Highlight can be added to the dialog list of segment categories shown in the last step of submitting. Maybe to reduce confusion it should use the start segment without asking because an end time logically doesn't make sense to be the highlight the user may have intended.

MarcaDian: I have never sent segments and do not use this feature. The only time, during a test, I checked whether the highlight was sent.

Therefore, I cannot suggest what behavior is correct, I only tried to implement what the user wrote in the request.

LisoUseInAIKyrios:
There seems to be an issue with setting the current time as a start/end time, and the time is always a little off from where the current video is.

I'm not sure how long it's been partially broken and I'm surprised nobody filed a bug report.

LisoUseInAIKyrios:
Set video time is fixed.

If someone has a tv that supports linking to YouTube, please verify the SponsorBlock set start/end time function using the current video time still works.

Update: Verified it works.

WHITE-KNIGHT5:
🫡 Morphe

Issue #1311, opened 2026-04-28
bug(YouTube Music): Crossfade does not work when screen is locked or the player is minimized


sr-shishir:
### Bug description

On patch 1.25.0-dev.7 The crossfade feature in YouTube Music does not function when the screen is locked or the player is minimized. Under these conditions, playback continues without crossfade, resulting in abrupt transitions between tracks.  

**Steps to Reproduce**:  
1. Enable crossfade in YouTube Music settings > morphe > player > crossfade
2. Play any playlist or album.  
3. Lock the device screen or minimize the player.  
4. Wait for one track to end and the next to begin. 

**Expected Behavior**:  
Crossfade should smoothly transition between tracks regardless of whether the screen is locked or the player is minimized.  

**Actual Behavior**:  
Crossfade fails to apply, and tracks switch abruptly without blending. 

Also 

> _**crossfade on manual skip**_

 is working while it’s turned off.

### App version patched

8.47.56

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

sr-shishir:
@VazerOG Facing this issue

DavidPv07:
same issue here. i  take log

[youtube_crossfade_off_screen.txt](https://github.com/user-attachments/files/27166268/youtube_crossfade_off_screen.txt)

VazerOG:
Under review.

sr-shishir:
@VazerOG Crossfade again fails in lock screen & minimized player. But this time is different. 

➤ Firstly after playing any song when I minimize the player & lock screen then it crossfade the next track smoothly but after that the rest songs don't crossfade anymore. It switches abruptly like previous issue.

Check this tracks timeline 


<details>

<img width="3264" height="3264" alt="Image" src="https://github.com/user-attachments/assets/cfed0120-75e3-4e72-a36c-f5c7c8de09b0" />

</details>

➤ Also in minimized player mode it behaves same like previous issue.

Check this timeline too

<details>

<img width="3264" height="3264" alt="Image" src="https://github.com/user-attachments/assets/b9688bc9-70e4-444f-b297-c49294e838d9" />

</details>

I don’t know what's wrong with this crossfade issue with every YT Music client player. 😥

[...]

PR #1312, opened 2026-04-28
fix(YouTube Music - Track crossfade): Crossfade does not work when screen is locked

VazerOG:
## Description

Fixes two bugs reported in #1311 where crossfade fails when the screen is locked or the player is minimized, and where the `Crossfade on manual skip` setting is bypassed for skips near the end of a track.

### **Bug 1**: Crossfade stops working with screen locked or player minimized

`Activity.onStop()` fires on both screen lock and app minimize. `onActivityStop()` was calling `stopAutoAdvanceMonitor()`, killing the polling loop. Both `onBeforePlayNext` and the monitor's inner guard also checked `!activityRunning`, blocking all crossfade while the activity was stopped.

> Fix: remove `stopAutoAdvanceMonitor()` from `onActivityStop()` and remove the `!activityRunning` check from `onBeforePlayNext` and the monitor loop. The monitor continues running in the background regardless of activity state.

### **Bug 2**: `Crossfade on manual skip` setting bypassed within last 5 seconds of a track

`onBeforeStopVideo` used a position-based `isAutoAdvance` heuristic (remaining < 5000ms) to decide which setting to check. Skipping within the last 5 seconds set `isAutoAdvance = true`, routing to `CROSSFADE_ON_AUTO_ADVANCE` instead of `CROSSFADE_ON_SKIP` — so crossfade would fire on manual skips even with the setting disabled.

> Fix: `onBeforeStopVideo` is always a manual skip — true auto-advance is handled exclusively by `onBeforePlayNext`. Remove the heuristic and always check `CROSSFADE_ON_SKIP`. Remove the now-unused `AUTO_ADVANCE_THRESHOLD_MS` constant.

### **Bug 3**: Discovered that `session control` was trying to use a stale instance of the shuffle button found at first-launch and was not relocating the shuffle button as it should. 

> Fix: added function `longPressAttachRetry` to re-attach the longPress handler to the new shuffle button when one isn't found.

## Additional context

 Tested on the following versions: 

- 8.44.54
- 8.47.56
- 8.50.51

### Test results

### DIFF
--- a/extensions/music/src/main/java/app/morphe/extension/music/patches/CrossfadeManager.java
+++ b/extensions/music/src/main/java/app/morphe/extension/music/patches/CrossfadeManager.java
@@ -225,7 +225,6 @@ public class CrossfadeManager {
     private static final int READY_TIMEOUT_MS = 10000;
     private static final int STATE_READY = 3;
     private static final int REASON_DIRECTOR_RESET = 5;
-    private static final long AUTO_ADVANCE_THRESHOLD_MS = 5000;
     private static final long MONITOR_POLL_MS = 100;
     // Extra lead time to absorb poll granularity + new-player READY latency (~120-200ms typical).
     // Ensures the fade-out completes before the old track's audio content runs out.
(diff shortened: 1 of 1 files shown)


DavidPv07:
How I can build MPP file with this patches

LisoUseInAIKyrios:
You can download and manually add the mpp file from the checks tab here.

Edit: This change is now merged. Patch as usual with pre-release.

VazerOG:
Greetings @DavidPv07

To build the .mpp file yourself you'll need JDK 17 and Gradle:

 ## Clone the repo
 ```
git clone https://github.com/MorpheApp/morphe-patches.git
cd morphe-patches
```

 ## Build
  ./gradlew build

 ## The .mpp artifact will be at: patches/build/libs/patches-<version>.mpp

  Then patch your APK using [Morphe CLI](https://github.com/MorpheApp/morphe-cli):

```
 java -jar morphe-cli-<version>-all.jar patch \
    -p patches/build/libs/patches-<version>.mpp \
    -o ytmusic-patched.apk \
    /path/to/base-ytm.apk
```

You'll need to acquire the appropriate version of the APK compatible with your device from APK Mirror or other trusted mirror platform, if you don't already have one. 

  I don't know what your skill-level is, but this can be quite complex for a novice user, but a bit of Googling and chatting with AI might help. 

  If you just want to use the latest release without building from source, pre-built .mpp files are attached to each [GitHub release](https://github.com/MorpheApp/morphe-patches/releases) when they're actually released. This will be in the next dev release shortly if the maintainers approve the changes.

I suspect this will be in version 1.25.0-dev8 initially.

LisoUseInAIKyrios:
The changes look good here. @VazerOG Ready to merge?

VazerOG:
> The changes look good here. @VazerOG Ready to merge?

Ready when you are, captain.

PR #1322, opened 2026-04-30
feat(YouTube - System media controls): Add options to hide prev/next and seekbar in notification media controls

btraas:
### Description

2 new settings for YouTube under Player - > System media controls (new section): 

1.  Hide Prev/Next in notification
- Previous and next buttons will be removed in the persistent media notification and lock screen
- Previous and next headphone controls have no effect
- Setting turned off by default. Both these effects are mentioned in the setting description
- Requires media to stop/start or restart app to apply changes. Prompts user to restart app when toggling on/off

2. Disable Seekbar in notification
- Seekbar is disabled in the persistent media notification and lock screen. It becomes a read-only progress bar instead.
- Requires media to stop/start media or restart app to apply changes. Prompts user to restart app when toggling on/off

### Test results

Tested on 20.21.37 and 20.47.62 with Samsung S23 Ultra, Android 14. Builds fine, works as intended

### DIFF
--- a/patches/src/main/kotlin/app/morphe/patches/youtube/misc/medianotification/MediaNotificationControlsPatch.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/youtube/misc/medianotification/MediaNotificationControlsPatch.kt
@@ -1,0 +1,45 @@
+package app.morphe.patches.youtube.misc.medianotification
+
+import app.morphe.patcher.extensions.InstructionExtensions.addInstructions
+import app.morphe.patcher.extensions.InstructionExtensions.getInstruction
+import app.morphe.patcher.patch.bytecodePatch
+import app.morphe.patches.youtube.misc.extension.sharedExtensionPatch
+import app.morphe.patches.youtube.misc.settings.settingsPatch
+import app.morphe.patches.youtube.shared.Constants.COMPATIBILITY_YOUTUBE
+import com.android.tools.smali.dexlib2.iface.instruction.FiveRegisterInstruction
+
+private const val EXTENSION_CLASS =
+    "Lapp/morphe/extension/youtube/patches/MediaNotificationControlsPatch;"
+
+val mediaNotificationControlsPatch = bytecodePatch(
+    name = "Media notification controls",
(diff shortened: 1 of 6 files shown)


patches/src/main/kotlin/app/morphe/patches/youtube/misc/medianotification/MediaNotificationControlsPatch.kt
```diff
@@ -0,0 +1,51 @@
...
+        playerOverlayButtonsSettingsPatch
+    )
+
+    compatibleWith(COMPATIBILITY_YOUTUBE)
+
+    execute {
+        addPlayerOverlayPreferences(
+            SwitchPreference("morphe_hide_notification_media_prev_next"),
+            SwitchPreference("morphe_hide_notification_media_seekbar"),
+        )
```
LisoUseInAIKyrios: Should this go in the player overlay controls menu or somewhere else?
This is slightly different from the other player overlay controls.

@ILoveOpenSourceApplications Your thoughts?

ILoveOpenSourceApplications: I agree. It doesn't make sense to be within player overlay as this hides the notification control buttons.

Should a new preference category be added for this?

LisoUseInAIKyrios: A preference category or a new preference screen? We only have categories in a few places, and if a screen has them then all preferences need to be in a category otherwise the non category preferences blend into other categories.

I think this is ok to be in a new player subscreen (player > Media controls)

btraas: I wasn't quite sure where to put it, wherever you guys think it would be best! Thanks for considering this PR, I'd love for this to be included in the official patches!

I hit these buttons and seekbar by accident all the time, and it's super annoying, especially for livestreams where the official app "forgets" where you were and jumps back to live.

PR #1342, opened 2026-05-02
feat(YouTube - Miniplayer): Add "Disable resuming miniplayer" setting

inotia00:
### Description

continue-watching-miniplayer

YouTube Premium users can use a feature called `Continue watching` (a feature that opens the last video being watched when the app cold starts).

This feature cannot be turned off, and some YouTube Premium users want to disable it.

This PR adds an option to turn off this feature via `Disable resuming miniplayer`.

Since this patch targets YouTube Premium users, it is not necessary for most users.

### Test results

### DIFF
--- a/patches/src/main/resources/addresources/values/youtube/strings.xml
+++ b/patches/src/main/resources/addresources/values/youtube/strings.xml
@@ -1583,11 +1583,10 @@ Limitation: Using the back button on the toolbar may not work"</string>
 
     <!-- layout.miniplayer.miniplayerPatch -->
     <string name="morphe_disable_resuming_miniplayer_title">Disable resuming miniplayer</string>
-    <string name="morphe_disable_resuming_miniplayer_summary_on">"'Continue watching' will not resume on app startup"</string>
-    <string name="morphe_disable_resuming_miniplayer_summary_off">"'Continue watching' will resume on app startup
-
-• 'Continue watching' is the YouTube Premium feature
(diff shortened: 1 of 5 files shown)


patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -1582,6 +1582,12 @@ Limitation: Using the back button on the toolbar may not work"</string>
     <string name="morphe_shorts_autoplay_background_summary_off">Shorts background play will repeat</string>
 
     <!-- layout.miniplayer.miniplayerPatch -->
+    <string name="morphe_disable_resuming_miniplayer_title">Disable resuming miniplayer</string>
+    <string name="morphe_disable_resuming_miniplayer_summary_on">"'Continue watching' will not resume on app startup"</string>
+    <string name="morphe_disable_resuming_miniplayer_summary_off">"'Continue watching' will resume on app startup
+
+• 'Continue watching' is the YouTube Premium feature
+• This setting does not force 'Continue watching' to be enabled"</string>
```
inotia00: TODO:

- [x] Review this explanation again to see if it is appropriate.
- [x] Test if it works on an experimental YouTube target (it will probably work well).

LisoUseInAIKyrios: I changed it to a user dialog message because the miniplayer screen is already quite full of UI text.

21.17 works

LisoUseInAIKyrios (approved):

PR #1346, opened 2026-05-02
feat(YouTube - Sanitize sharing links): Add setting to change Share links to shortened URLs

FlaminSarge:
### Description

Adds setting to replace `youtube.com/live/VIDEOID` and `youtube.com/shorts/VIDEOID` urls with `youtu.be/VIDEOID` urls.
This only affects live/shorts as normal links already get shortened by stock YT.
Closes #

### Additional context

Motivation: YT recently introduced a bug where embed metadata for live urls (technically /e/ and /v/ urls too, but those aren't really used) doesn't work properly (at least in Discord). Adding this option so that shared livestreams can use the short link and get correct metadata that way.

### Test results

- [x] Tested on both the **minimum** and **maximum** supported versions - 20.21.37, 20.47.62
- [x] Tested on **experimental** supported versions (Optional) - 21.17.480

### DIFF
--- a/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/patches/SanitizeSharingLinksPatch.java
+++ b/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/patches/SanitizeSharingLinksPatch.java
@@ -1,6 +1,7 @@
 package app.morphe.extension.shared.patches;
 
 import android.net.Uri;
+import android.text.TextUtils;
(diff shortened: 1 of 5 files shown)


FlaminSarge:
Could probably be applied to YTMusic too but for now just doing it for base YT.

LisoUseInAIKyrios:
Looks good. Will merge in a day or two if there are no concerns.

PR #1389, opened 2026-05-06
feat: Add miscellaneous link handling preference to help change 'open with' links

### LINKED ISSUE #605: feat(YouTube): Add open default app settings
### Feature description

Most Xiaomi phones now support `HyperOS`, but unlike other Android devices, `HyperOS` doesn’t provide a built‑in option in App info to set default apps. That means users are forced to install third‑party apps like ❝activity launcher❞ just to change defaults, which feels inconvenient and wastes time. So add on option to set `YouTube Morphe` as default while open any `YouTube` web addresses. This feature  will very helpful for Xiaomi users.

Image
**This is a concept image**

### Motivation

Why should this feature be implemented?
Because users expect the ability to set default apps directly within the system settings. HyperOS currently lacks this, forcing reliance on third‑party tools, which undermines convenience and user satisfaction. 

A user wants to open links, media files, or documents with their preferred app (e.g., choosing Chrome instead of the built‑in browser, or VLC instead of the stock video player). Without a native option, they cannot assign these defaults easily. 

The benefits are:-
- Efficiency: Saves time by removing the need for external apps.  
- Consistency: Ensures a smoother experience across apps and workflows.  
- Security: Reduces dependency on third‑party apps that may pose risks.  
- User control: Empowers users to personalize their device according to their habits.

It’s a core usability feature in modern operating systems. By not including it, HyperOS feels incomplete compared to competitors. Implementing it would improve user trust, reduce frustration, and align Xiaomi devices with standard Android expectations.  

### Acknowledgements

MarcaDian:
### Description

Adds a new preference in the `Misc` settings screen that guides users through reassigning app links from the original app to the patched one. Tapping it opens a two-step dialog: first to clear links from the original app, then to enable them for the patched app.
___


MarcaDian:
Tested YouTube and YT Music.

LisoUseInAIKyrios:
If the 'open with' links are already configured correctly, then perhaps the preference can show a different text or possible the preference is disabled entirely.

MarcaDian:
Good idea. However, reliably detecting the current link handling state is tricky - the `DomainVerificationManager` API that would allow us to check if the patched app is already the verified handler is only available on API 31+, and there's no clean equivalent for older versions. Given our min SDK is 26, we'd need a fallback that may not be accurate anyway. I'd rather keep it simple for now and leave it as a future improvement if there's demand for it.

LisoUseInAIKyrios:
Disabling the preference could be an API 31+ only behavior, and for older devices it always shows the same 'click here to fix this' text.

We can add this later if we want or  there is confusion about this feature.

MarcaDian:
The last change may be overcode, it updates the preferences upon entering the screen and not upon starting the fragment.

LisoUseInAIKyrios:
Seems ready to merge

Issue #1284, opened 2026-04-25
bug(YouTube): Shorts player has no buttons / buttons are hidden


LisoUseInAIKyrios:
### Bug description

For some users, the Shorts player has no player overlay buttons except sometimes a few buttons at the top of the screen.

Issue affects 20.31.42+

20.21.37 seems to be unaffected.

### App version patched

20.31.42

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

LisoUseInAIKyrios:
I am quite sure there was/is a previous issue on this but I cannot find it.

LisoUseInAIKyrios:
After the issue showed up with 20.31.42 after upgrading from 20.21.37, I started excluding patches until the problem went away.

After excluding `Spoof app version` the problem was fixed, but then after re-including the patch again the problem was still fixed.

The debug logs did not show the litho filter was hiding anything, so the problem is something else.

I suspect it might be the experimental "next gen" Shorts layout, and the old Shorts player layout data is missing:
https://github.com/MorpheApp/morphe-patches/blob/73c2c58f27f62e1221215ce54e39fe9859feeaf5/patches/src/main/kotlin/app/morphe/patches/youtube/layout/hide/shorts/HideShortsComponentsPatch.kt#L272-L277

ReplacementFit4095:
>there was/is a previous issue

Is it this one? https://github.com/MorpheApp/morphe-patches/issues/183

I remember that issue and link to it sometimes whenever people have problems with the Shorts UI

LisoUseInAIKyrios:
I remember a different issue that was unrelated to version spoofing.

OrdinaryPerson0:
https://github.com/MorpheApp/morphe-patches/issues/183#issuecomment-4215873109 The comment in this issue report points to a different problem with Shorts player. Could this be possible?

Also can have the same issue with my non-root install. I only see this issue with Shorts recommended under a video. And only happens sometimes. But i remember having this issue even RV times not after 20.31.42

DARKL17:
#981 The button is hidden

error-reporting:
<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/dee4b433-bb5c-49f7-bc47-91344209ae41" />
those overlay had returned for my account recently

ILoveOpenSourceApplications:
[...]

Issue #663, opened 2026-02-28
feat(YouTube): Shorts: add `Enable Clear mode` and `Change Shorts repeat state`


Kantotero:
### Feature description

Both of these patches were present in Inotia's RVX. In the first patch it removes all the distracting buttons, icons, and text display in clear mode, kinda similar to clear display feature in Tiktok.
![Image](https://github.com/user-attachments/assets/c9efc528-faed-4c02-a6ca-eed1229f25bd)

Meanwhile change repeat state in Shorts features the ability to pause the Shorts instead of the default looping of YT shorts in the latest versions.
![Image](https://github.com/user-attachments/assets/f4290753-73be-4540-8c4b-29219664bf7e)

### Motivation

It would be really convenient to have a clean and distraction-free streaming of YT shorts. And the feature to stop  a YT short from looping just like in RVX

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] This is a feature request for YouTube, YT Music, or Reddit

LisoUseInAIKyrios:
Shorts already has a repeat and autoplay option.  See https://github.com/MorpheApp/morphe-patches/issues/91#issuecomment-3708342565

PR #1421, opened 2026-05-09
feat(YouTube): Add `Dim Shorts overlay` patch

Nicsilver:
Adds a new **Dim Shorts overlay** patch that reduces the brightness of the Shorts UI to help prevent OLED burn-in. Dims the overlay, toolbar buttons, nav bar tabs, and seekbar based on a configurable opacity setting. Also includes an immersive mode option that hides the status bar while Shorts is open.

| 100% (no patch) | 50% opacity + Immersive | 20% opacity + Immersive |
|-----------------|-------------|-------------|
|  |  |  |

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/DimShortsOverlayPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/DimShortsOverlayPatch.java
@@ -1,0 +1,186 @@
+package app.morphe.extension.youtube.patches;
+
+import android.app.Activity;
+import android.content.Context;
+import android.content.ContextWrapper;
+import android.os.Build;
+import android.view.View;
+import android.view.ViewParent;
+import android.view.ViewTreeObserver;
+import android.view.Window;
+import android.view.WindowInsetsController;
+import android.widget.ImageView;
+
+import app.morphe.extension.shared.Logger;
+import app.morphe.extension.youtube.settings.Settings;
+import app.morphe.extension.youtube.shared.NavigationBar.NavigationButton;
+import app.morphe.extension.youtube.shared.ShortsPlayerState;
+
+@SuppressWarnings("unused")
+public class DimShortsOverlayPatch {
+
+    private static int reelTimeBarId = 0;
+
+    /**
+     * Injection point.
+     */
+    public static void dimShortsPlayerOverlay(View shortsOverlayView) {
+        shortsOverlayView.addOnAttachStateChangeListener(new View.OnAttachStateChangeListener() {
+            private ViewTreeObserver registeredObserver;
+            private ViewTreeObserver.OnPreDrawListener preDrawListener;
+            private Window window;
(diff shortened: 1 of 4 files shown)


ILoveOpenSourceApplications:
@Nicsilver, can you also add some screenshots for comparison?

Nicsilver:
> @Nicsilver, can you also add some screenshots for comparison?

There we go :)

ILoveOpenSourceApplications:
Do you think #663 is within the scope of this PR?

LisoUseInAIKyrios:
This almost handles that if the overlay value is set to 0, but I imagine the non visible buttons may still work.

Nicsilver:
I mean I'm not 100% sure what clear mode is, but if it actually removes the buttons, it's different from this. With this patch at 0% opacity buttons are all still clickable.

ILoveOpenSourceApplications:
They want the buttons to auto hide just like in the normal video player, where tapping once shows the button and tapping again will hide them.

Nicsilver:
Yeah ok. That's outside the scope of this imo

Nicsilver:
I think this is ready for review? :)

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/DimShortsOverlayPatch.java
```diff
@@ -0,0 +1,186 @@
...
+@SuppressWarnings("unused")
+public class DimShortsOverlayPatch {
+
+    private static int reelTimeBarId = 0;
+
+    /**
+     * Injection point.
+     */
+    public static void dimShortsPlayerOverlay(View shortsOverlayView) {
+        shortsOverlayView.addOnAttachStateChangeListener(new View.OnAttachStateChangeListener() {
```
LisoUseInAIKyrios, line 28: I'm confident this can be done without state change and predraw listeners.

I've been working on the a fix for the Shorts player showing no buttons (caused by the '[next gen' Shorts player](https://github.com/MorpheApp/morphe-patches/issues/1284#issuecomment-4319127301) that does not use Litho), and each of the Shorts components are hooked individually.

That means hiding components in the new Shorts player is done directly to the button views, much like the old YouTube 17.x releases from 4 years ago before litho was extensively used. Any opacity changes can be done then at the same time.

LisoUseInAIKyrios:
Thanks for the effort, but the changes here will collide with and immediately become obsolete with the upcoming Shorts 'next gen' player patch changes.

Nicsilver:
Makes sense @LisoUseInAIKyrios — I dug into it and can see you're force-disabling the experimental player flags while the next gen support is being worked on. Worth noting this patch does a bit more than just the Shorts overlay — it also dims the bottom nav buttons and toolbar buttons app-wide when Shorts is open, and adds an immersive mode that hides the status bar. Since you mentioned opacity changes could be done alongside the individual button hooks, would it be possible to fold this in when the next gen player work is ready? Happy to adapt to whatever hook structure you're building around.

PR #1433, opened 2026-05-11
fix(YouTube - Spoof video streams): Resolve Shorts freezing and feed scrolling issues

### LINKED ISSUE #1413: bug(YouTube): Home / feed scrolling stuttering
### Bug description

In the latest dev patch, after patching YouTube, the scrolling is stuttering too much. I'm on an Android 16 Realme GT 7 Pro device. I have a video too, but I don't know why I'm unable to upload it here it's showing failed. 

To reproduce the bug, just scroll up and down slowly and wait in between - it'll stutter. I've also attached my enabled patches list.

[YouTube_Settings_2026-05-08.txt](https://github.com/user-attachments/files/27530194/YouTube_Settings_2026-05-08.txt)

### App version patched

version 21.19.280

### Error logs

```shell

```

### Acknowledgements

### LINKED ISSUE #1425: bug(YouTube): Shorts content not loading when scrolled
### Bug description

Youtube App version: 20.47.62
Patched version: 1.26.0
Steps to reproduce the bug:
1. Open youtube shorts
2. Sometimes first video already wont load, it shows a freeze frame of that video, the like dislike toolbar loads, no loading circle too

<details>

Image
</details>
Something like that...

4. Or you need to scroll a bit more to make it happen 
5. After scrolling a bit deeper some videos can still load, and then it will freeze again after scrolling a bit more
6. It seems like it is only happening to more recent YouTube versions, versions below 20.47.62 works ok and doesn't bug out that badly(but i want the new features so i dont really want to use old versions)
Some things that i have already tried:
1. Changing my spoof client to all of the clients even logging in to android vr(for those people so ask to use android reel client, its the same so dont sweat it)
2. Making sure my dns is off (i use an adblocker dns)
3. Using the microg that was provided by morphe and updated it. 
4. Turning off return youtube dislike(only for shorts as normal long form content loads nornally)
5. Clearing cache

I really hope that the morphe developers can fix this issue.

Edit: it turns out sometimes( rarely) the short would load but it takes like 3 minutes to load and my wifi speed is quite fast (i did comparison with unpatched yt it loads the videos smoothly)
Edit²: It turns out the short that loads after 3 mins dont have a selection for quality change and it is at a low quality too, could it be the video was fetched from some source eventually, that doesn't have other quality?
### App version patched

20.47.62

### Acknowledgements

AboodFares:
Fixes two feature flags causing issues in recent YouTube versions.

Flag 45638126 was causing Shorts to freeze and fail to load when scrolling
on 20.47.62+. Flag 45782902 was causing the home feed to stutter while
scrolling on 21.19+.

Both flags were tracked down by users in #1425 and #1413 using the debug
feature flag manager. The fix follows the same approach already used in the
codebase for disabling problematic experimental Shorts flags.

### Test results


LisoUseInAIKyrios:
For either, do Shorts work correctly if stream spoofing is turned off?

if Shorts work with spoofing off, then those flag overrides can be part of spoof video streams patch (and the flags are forced off only when stream spoofing is active)

AboodFares:
I don't have an Android device to test this. Could you check with the reporter in #1425 to confirm if Shorts work with spoofing disabled?

LisoUseInAIKyrios:
These flags seem to interfere with stream spoofing. The override can be moved to `SpoofVideoStreams` and it only forces off if stream spoofing is enabled

AboodFares:
Moved 45638126 to the spoof patch, it'll only force off when spoofing is active now

sajeshyadav:
When I increase the playback speed, the app crashes.

LisoUseInAIKyrios:
Please use a computer and collect the crash logs as described here https://github.com/MorpheApp/morphe-documentation/blob/main/docs/morphe-resources/questions.md#40-capture-logs

Sommergoldhaehnchen:
This patch crashes app when trying to change playback speed!

LisoUseInAIKyrios:
It does. Will look into it and fix or revert.

Update: Issue was from a refactoring and not this

PR #1470, opened 2026-05-13
chore(YouTube - Open channel of live avatar): added response log

0xrxL:


LisoUseInAIKyrios:
If we want to make a new release, then need to change commit to `fix:`

PR #1472, opened 2026-05-13
fix(YouTube - PlayerOverlayButton): Video heading margin not applied when no bottom buttons present

MarcaDian:
### Description

The video heading end margin was only updated inside the OnPreDrawListener of lower overlay buttons (speed, quality, etc.). When none of those buttons were added, the listener never fired, so player_video_heading never received its margin and overlapped the upper custom buttons.

### Test results

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/videoplayer/LegacyPlayerControlButton.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/videoplayer/LegacyPlayerControlButton.java
@@ -83,8 +83,6 @@ public class LegacyPlayerControlButton {
                                      PlayerControlButtonStatus enabledStatus,
                                      View.OnClickListener onClickListener,
                                      @Nullable View.OnLongClickListener longClickListener) {
-        PlayerOverlayButton.initializeHeadingFromUpperButton(controlsViewGroup);
-        
         View containerView = Utils.getChildViewByResourceName(controlsViewGroup, viewToHide);
         containerView.setVisibility(View.GONE);
         containerRef = new WeakReference<>(containerView);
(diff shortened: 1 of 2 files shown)


MarcaDian:
```
PlayerOverlayButton: Could not find button overlay: player_video_heading
```

0xrxL:
> ```
> PlayerOverlayButton: Could not find button overlay: player_video_heading
> ```

If the fix works, and the problem is this, you can check if the object is null?

MarcaDian:
I tried 2 times and didn't fix the error, I don't have time to do it yet

0xrxL:
Now it works fully, although it return only one error after opening the player from the cold runned app.

0xrxL:
@LisoUseInAIKyrios what do you think? This one time error can be ignored or should be fixed?

LisoUseInAIKyrios:
It shows an error toast on cold launch? We should fix, or if the error is not actually an error then can ignore the first time it happens and not show a toast.

0xrxL:
> It shows an error toast on cold launch? We should fix, or if the error is not actually an error then can ignore the first time it happens and not show a toast.

It show the error one time on cold launch, during video opening. But idk how to fix it honestly.

LisoUseInAIKyrios:
Can the error be ignored?

If so just add a static boolean field and allow a one time ignoring of the error.

0xrxL:
> Can the error be ignored?
> 
> If so just add a static boolean field and allow a one time ignoring of the error.

Yes, the error can be ignored, and it mainly depends on the player loading. After the first error, no further errors are displayed. I'll try to see if it can be fixed by applying some null checks to the bytecode at least.

MarcaDian:
> Now it works fully, although it return only one error after opening the player from the cold runned app.

Basically, this is what I tried to fix about 8 times in different ways.

0xrxL:
> > Now it works fully, although it return only one error after opening the player from the cold runned app.
> 
> Basically, this is what I tried to fix about 8 times in different ways.

Different class methods have different behaviour. That's why it cannot be reproduced on the bottom bar.

PR #1475, opened 2026-05-14
chore: Improve release notes generation

brosssh:


.releaserc
```diff
@@ -8,13 +8,32 @@
...
+    ],
+    [
+        "@semantic-release/release-notes-generator", {
+        "preset": "conventionalcommits",
+        "presetConfig": {
+            "types": [
+                { "type": "fix", "section": "🐛 Bug Fixes", "hidden": false },
+                { "type": "feat", "section": "✨ New Features", "hidden": false },
+                { "type": "bump", "section": "🚀 Updated App Support", "hidden": false },
+                { "type": "ci", "section": "🔁 CI", "hidden": false },
```
LisoUseInAIKyrios: This will generate a release for `ci:` commits.

Is this desirable? Normally ci: does not generate a release, and continuous integration isn't anything that affects the patches.

brosssh: Depends on the type of ci change I guess. Let me know if you want me to remove that. Even if you want some other custom keywords.

LisoUseInAIKyrios: Lets remove it because I don't think users will care, and the default for 'ci' is no build.

I can't think of any other types.

PR #1366, opened 2026-05-04
feat: Simplify settings summary text

krvstek:
### Description

This PR cleans up how we handle settings descriptions across all apps. Instead of maintaining separate `_on` and `_off` strings for every single toggle, we're moving to a single, unified `_summary` that just describes what the feature does. 

Closes #1097. 

### Additional context

This cuts down the string payload and reduces the translation work almost in half.

### Test results

- [x] Tested on both the **minimum** and **maximum** supported versions
- [x] Tested on **experimental** supported versions (YT 21.18.163, YTM 9.17.51, Reddit 2026.18.0)

PR #1485, opened 2026-05-15
chore(Settings): Additional strings improvements

kitadai31:
@krvstek

Additional follow-up string improvements to #1366.

These are all just suggestions. Please feel free to share your opinions.

Notable changes:

- `morphe_hide_creator_store_shelf_title`: Moved to the correct patch section
- `morphe_restore_old_settings_menus_title`: Because it was unclear whether this was a Morphe setting or a YouTube setting, I have clarified that it is YouTube. I have also emphasized that it restores the old "screen." I think the word "menu" can be confused with any menus such as old quality setting menu.
- `morphe_hide_surveys_summary`: Added a example of surveys. This is needed because in my language, "survey" was being confused with polls in a community post by a creator.
- `morphe_show_settings_button_type_summary` and `morphe_show_toolbar_settings_button_type_summary`: I used a new expression here. The meanings of ON and OFF are now listed separately.
- `morphe_sb_guidelines_popup_content`: This is reverted to the previous string. Since reading the guidelines is an obligation for everyone submitting a segment, I felt that "Read" sounds stronger and is more appropriate than "View."
- `morphe_sb_general_min_duration_summary`: The information that the segment was not being shown was lost, so I reverted to the previous string.

I'm creating this pull request as a draft because I might add more changes later.

### DIFF
--- a/patches/src/main/kotlin/app/morphe/patches/youtube/interaction/downloads/DownloadsPatch.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/youtube/interaction/downloads/DownloadsPatch.kt
@@ -32,7 +32,7 @@ private val downloadsResourcePatch = resourcePatch {
                 key = "morphe_external_downloader_screen",
                 sorting = Sorting.UNSORTED,
                 preferences = setOf(
-                    SwitchPreference("morphe_external_downloader", summaryKey = null),
+                    SwitchPreference("morphe_external_downloader"),
                     SwitchPreference("morphe_external_downloader_action_button"),
                     TextPreference(
                         "morphe_external_downloader_name",
--- a/patches/src/main/resources/addresources/values/youtube/strings.xml
+++ b/patches/src/main/resources/addresources/values/youtube/strings.xml
@@ -61,7 +61,6 @@ Changes include:
 • Account switch button may not appear on the Library tab."</string>
 
     <!-- layout.hide.general.hideLayoutComponentsPatch -->
-    <string name="morphe_hide_creator_store_shelf_title">Hide creator store shelf</string>
     <string name="morphe_hide_feed_flyout_menu_title">Enable feed flyout menu filter</string>
     <string name="morphe_hide_feed_flyout_menu_filter_strings_title">Feed flyout menu filter</string>
     <string name="morphe_hide_feed_flyout_menu_filter_strings_summary">Enter the flyout menu names to filter by, separated by new lines</string>
@@ -268,6 +267,7 @@ Limitations:
     <string name="morphe_hide_keyword_toast_invalid_broad">Keyword will hide all videos: %s</string>
 
     <!-- ad.general.hideAdsPatch -->
+    <string name="morphe_hide_creator_store_shelf_title">Hide creator store shelf</string>
     <string name="morphe_hide_end_screen_store_banner_title">Hide end screen store banner</string>
     <string name="morphe_hide_general_ads_title">Hide general ads</string>
     <string name="morphe_hide_merchandise_banners_title">Hide merchandise banners</string>
--- a/patches/src/main/resources/addresources/values/youtube/strings.xml
+++ b/patches/src/main/resources/addresources/values/youtube/strings.xml
@@ -315,13 +315,13 @@ Limitations:
 
     <!-- interaction.downloads.downloadsPatch -->
     <string name="morphe_external_downloader_screen_title">External downloads</string>
-    <string name="morphe_external_downloader_screen_summary">Configure an external downloader app for downloading videos</string>
+    <string name="morphe_external_downloader_screen_summary">Download videos with third-party downloading apps</string>
     <string name="morphe_external_downloader_title">Show external download button</string>
-    <string name="morphe_external_downloader_summary">Shows a download button in video player that opens your external downloader</string>
+    <string name="morphe_external_downloader_summary">Shows an overlay button in video player that opens your external downloader</string>
     <!-- 'Download action button' should match the language used in 'morphe_hide_download_button_title'. -->
     <string name="morphe_external_downloader_action_button_title">Override Download action button</string>
     <string name="morphe_external_downloader_action_button_summary">Replaces YouTube\'s built-in download function with your external downloader</string>
-    <string name="morphe_external_downloader_name_title">External downloader package name</string>
+    <string name="morphe_external_downloader_name_title">Downloader package name</string>
     <string name="morphe_external_downloader_name_summary">Package name of your installed external downloader app</string>
     <string name="morphe_external_downloader_other_item_hint">Enter the package name</string>
     <string name="morphe_external_downloader_other_item">Other</string>
@@ -1202,7 +1202,7 @@ Tap here to learn more about DeArrow"</string>
 
     <!-- misc.dimensions.spoof.spoofDeviceDimensionsPatch -->
     <string name="morphe_spoof_device_dimensions_title">Spoof device dimensions</string>
-    <string name="morphe_spoof_device_dimensions_summary">May unlock higher video qualities</string>
+    <string name="morphe_spoof_device_dimensions_summary">Enabling this can unlock higher video qualities</string>
     <string name="morphe_spoof_device_dimensions_user_dialog_message">Enabling this can cause video playback stuttering, worse battery life, and unknown side effects.</string>
 
     <!-- misc.hapticfeedback.disableHapticFeedbackPatch -->
--- a/patches/src/main/resources/addresources/values/youtube/strings.xml
+++ b/patches/src/main/resources/addresources/values/youtube/strings.xml
@@ -30,7 +30,7 @@ Second \"item\" text"</string>
     <string name="morphe_settings_screen_09_return_youtube_dislike_title" translatable="false">Return YouTube Dislike</string>
     <string name="morphe_settings_screen_11_misc_title">Miscellaneous</string>
     <string name="morphe_settings_screen_12_video_title">Video</string>
-    <string name="morphe_restore_old_settings_menus_title">Restore old settings menus</string>
+    <string name="morphe_restore_old_settings_menus_title">Restore old YouTube settings screen</string>
 
     <!-- misc.backgroundplayback.backgroundPlaybackPatch -->
     <string name="morphe_remove_background_playback_restrictions_title">Remove background play restrictions</string>
@@ -61,6 +61,7 @@ Changes include:
 • Account switch button may not appear on the Library tab."</string>
 
     <!-- layout.hide.general.hideLayoutComponentsPatch -->
+    <string name="morphe_hide_creator_store_shelf_title">Hide creator store shelf</string>
     <string name="morphe_hide_feed_flyout_menu_title">Enable feed flyout menu filter</string>
     <string name="morphe_hide_feed_flyout_menu_filter_strings_title">Feed flyout menu filter</string>
     <string name="morphe_hide_feed_flyout_menu_filter_strings_summary">Enter the flyout menu names to filter by, separated by new lines</string>
@@ -81,7 +82,7 @@ Changes include:
     <string name="morphe_hide_floating_microphone_button_title">Hide floating microphone button</string>
     <string name="morphe_hide_floating_microphone_button_summary">Hides floating microphone button in the lower-right corner of search suggestions screen</string>
     <string name="morphe_hide_horizontal_shelves_title">Hide horizontal shelves</string>
-    <string name="morphe_hide_horizontal_shelves_summary">Hides horizontal shelves such as Breaking news, Community posts you may like, Continue watching, Explore more channels, Explore more topics, For You, Latest videos, Most relevant, Posts from Creator\'s Community, Shopping and Watch it again</string>
+    <string name="morphe_hide_horizontal_shelves_summary">Hides horizontal shelves such as Breaking news, Community posts you may like, Continue watching, Explore more channels, Explore more topics, For You, Latest videos, Most relevant, Posts from Creator\'s Community, Shopping, and Watch it again</string>
     <string name="morphe_hide_hyped_label_title">Hide Hyped label</string>
     <string name="morphe_hide_image_shelf_title">Hide image shelf</string>
     <string name="morphe_hide_image_shelf_summary">Hides image shelves in search results</string>
@@ -94,7 +95,7 @@ Changes include:
     <!-- 'Notify me' should be translated using the same localized wording YouTube displays for the button.
           This button usually appears in the Subscriptions feed for future livestreams or unreleased videos. -->
     <string name="morphe_hide_notify_me_button_title">Hide \'Notify me\' button</string>
-    <string name="morphe_hide_notify_me_button_summary">Hides the reminder button for upcoming livestreams and premieres</string>
+    <string name="morphe_hide_notify_me_button_summary">Hides the reminder button for scheduled livestreams and premieres</string>
     <string name="morphe_hide_playables_title">Hide Playables</string>
     <string name="morphe_hide_playables_summary">Hides the feature that allows you to play games</string>
     <string name="morphe_hide_search_term_thumbnails_title">Hide search term thumbnails</string>
@@ -105,11 +106,13 @@ Changes include:
     <string name="morphe_hide_show_more_button_summary">Hides the expand button under videos in the search results</string>
     <string name="morphe_hide_subscribed_channels_bar_title">Hide subscribed channels bar</string>
     <string name="morphe_hide_surveys_title">Hide surveys</string>
+    <string name="morphe_hide_surveys_summary">Hides surveys such as \'Is the above video a good recommendation for you?\'</string>
     <string name="morphe_hide_ticket_shelf_title">Hide ticket shelf</string>
     <!-- 'People also watched' and 'You might also like' should be translated using the same localized wording YouTube displays for the label. -->
     <string name="morphe_hide_video_recommendation_labels_title">Hide video recommendation labels</string>
     <string name="morphe_hide_video_recommendation_labels_summary">Hides \'People also watched\' and \'You might also like\' labels in search results</string>
     <string name="morphe_hide_web_search_results_title">Hide web search results</string>
+    <string name="morphe_hide_web_search_results_summary">Hides \'Result from the web\' section from search results</string>
     <string name="morphe_hide_you_may_like_section_title">Hide \'You may like\' section</string>
     <!-- https://logos.fandom.com/wiki/YouTube/Yoodles -->
     <string name="morphe_hide_youtube_doodles_title">Hide YouTube Doodles</string>
@@ -242,7 +245,11 @@ Showing YouTube Doodles may prevent custom headers from displaying."</string>
     <string name="morphe_hide_keyword_content_phrases_title">Keywords to hide</string>
     <!-- For localization, it is preferred, but not required, if 'LeBlanc' is replaced with a localized name or a familiar word that has upper case letters in the middle of the word.
          This is because keywords can be in any language, and showing an example in the localized script helps convey this. -->
-    <string name="morphe_hide_keyword_content_phrases_summary">Enter keywords or channel names, separated by new lines. Words with uppercase letters in the middle must match exactly (ie: iPhone, TikTok, LeBlanc)</string>
+    <string name="morphe_hide_keyword_content_phrases_summary">"Enter keywords and phrases to hide, separated by new lines
+
+Keywords can be any text shown in video titles or channel names
+
+Words with uppercase letters in the middle must match exactly (ie: iPhone, TikTok, LeBlanc)"</string>
     <string name="morphe_hide_keyword_content_about_title">About keyword filtering</string>
     <string name="morphe_hide_keyword_content_about_summary">"Home / Subscriptions / Search results are filtered to hide content that matches keyword phrases
 
@@ -261,11 +268,11 @@ Limitations:
     <string name="morphe_hide_keyword_toast_invalid_broad">Keyword will hide all videos: %s</string>
 
     <!-- ad.general.hideAdsPatch -->
-    <string name="morphe_hide_creator_store_shelf_title">Hide creator store shelf</string>
     <string name="morphe_hide_end_screen_store_banner_title">Hide end screen store banner</string>
     <string name="morphe_hide_general_ads_title">Hide general ads</string>
     <string name="morphe_hide_merchandise_banners_title">Hide merchandise banners</string>
-    <string name="morphe_hide_paid_promotion_label_title">Hide paid promotion label</string>
+    <!-- 'Includes paid promotion' should be translated using the same localized wording YouTube displays for the label. -->
+    <string name="morphe_hide_paid_promotion_label_title">Hide \'Includes paid promotion\' label</string>
     <string name="morphe_hide_player_popup_ads_title">Hide player popup ads</string>
     <string name="morphe_hide_self_sponsor_ads_title">Hide self sponsored cards</string>
     <string name="morphe_hide_shopping_links_title">Hide shopping links</string>
@@ -335,8 +342,8 @@ Verify the package name is correct and the app is installed"</string>
 
     <!-- interaction.playall.playAllButtonPatch -->
     <string name="morphe_play_all_button_title">Show Play all button</string>
-    <string name="morphe_play_all_button_summary">"Tap to generate a playlist, tap and hold to undo
-
+    <string name="morphe_play_all_button_summary">"Tap to generate a playlist of all videos from channel
+Tap and hold to undo
 Info: May not work on live streams"</string>
     <string name="morphe_play_all_button_type_title">Generate playlist mode</string>
     <string name="morphe_play_all_button_type_entry_0">All contents (Sort by time, Ascending)</string>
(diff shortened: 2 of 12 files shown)


kitadai31:
p.s. I checked about 80% of the text, and I was very impressed by how many features were explained in detail.

Another good point is that some of the new summaries are consistent with its patch description.
I've thought the same thing. In many patches, the patch description has detailed explanations, but the summary in the Morphe settings lacks sufficient explanation.
This allows users to understand the functionality without having to read the patch description.
(In this PR, more summaries have been taken from the patch description.)

Once again, thank you for your hard work @krvstek !

krvstek:
Thanks! Really glad we're on the same page with improving the UX.

Just let me know when it's ready, and I'll be happy to drop my thoughts in the review ;P

ILoveOpenSourceApplications:
> - `morphe_hide_creator_store_shelf_title`: Moved to the correct patch section

This change isn't needed as the setting itself is in "Ads".

kitadai31:
> > * `morphe_hide_creator_store_shelf_title`: Moved to the correct patch section
> 
> This change isn't needed as the setting itself is in "Ads".

Don't strings belong to patches?
That string is from `layout.hide.general.hideLayoutComponentsPatch`
The hideLayoutComponentsPatch adds that settings to the `ADS` screen

ILoveOpenSourceApplications:
That can be moved into `HideAdsPatch` I think without any issues.

patches/src/main/kotlin/app/morphe/patches/youtube/interaction/downloads/DownloadsPatch.kt
```diff
@@ -32,7 +32,7 @@ private val downloadsResourcePatch = resourcePatch {
                 key = "morphe_external_downloader_screen",
                 sorting = Sorting.UNSORTED,
                 preferences = setOf(
-                    SwitchPreference("morphe_external_downloader", summaryKey = null),
+                    SwitchPreference("morphe_external_downloader"),
```
LisoUseInAIKyrios, line 35: Unrelated to this PR, but I think this we can cleanup this `summaryKey = null` usage.

`SwitchPreference("key", summaryKey = null)` is used 253 times
no summary key parameter (meaning there is a summary) is used 101 times.

I think the default behavior of `SwitchPreference("key")` can be changed to 'no summary',
and an overloaded constructor can be for the 1/3 of switches that have a summary. Something like `SwitchPreference("key", summary = true)`

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -61,7 +61,6 @@ Changes include:
 • Account switch button may not appear on the Library tab."</string>
 
     <!-- layout.hide.general.hideLayoutComponentsPatch -->
-    <string name="morphe_hide_creator_store_shelf_title">Hide creator store shelf</string>
     <string name="morphe_hide_feed_flyout_menu_title">Enable feed flyout menu filter</string>
     <string name="morphe_hide_feed_flyout_menu_filter_strings_title">Feed flyout menu filter</string>
     <string name="morphe_hide_feed_flyout_menu_filter_strings_summary">Enter the flyout menu names to filter by, separated by new lines</string>
```
LisoUseInAIKyrios, line 66:
```
    <string name="morphe_exit_fullscreen_title">Exit fullscreen mode on video end</string>
    <string name="morphe_exit_fullscreen_entry_1">Disabled</string>
    <string name="morphe_exit_fullscreen_entry_2">Portrait only</string>
    <string name="morphe_exit_fullscreen_entry_3">Landscape only</string>
    <string name="morphe_exit_fullscreen_entry_4">Both</string>
```

Not part of the changes here, but 'both' shows in the screen as:

```
Exit fullscreen mode on video end
Both
```

Does both mean it's current set to both disabled and portrait only?
Or is it exit fullscreen mode, and exit video end?

I think `morphe_exit_fullscreen_entry_4` reads better as `Portrait and landscape`

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -308,14 +315,14 @@ Limitations:
...
-
+    <string name="morphe_external_downloader_summary">Shows a download button in video player that opens your external downloader</string>
     <!-- 'Download action button' should match the language used in 'morphe_hide_download_button_title'. -->
     <string name="morphe_external_downloader_action_button_title">Override Download action button</string>
-    <string name="morphe_external_downloader_action_button_summary">Replaces YouTube\'s built-in download function with your chosen third-party downloading app</string>
-    <string name="morphe_external_downloader_name_title">Downloader package name</string>
-    <string name="morphe_external_downloader_name_summary">Package name of your installed external downloader app</string>
+    <string name="morphe_external_downloader_action_button_summary">Replaces YouTube\'s built-in download function with your external downloader</string>
+    <string name="morphe_external_downloader_name_title">External downloader package name</string>
+    <string name="morphe_external_downloader_name_summary">Package name of your installed third-party downloading app</string>
```
krvstek: Since we are switching the vocabulary to `external` anyway, it's better to keep it consistent and use `external downloader` in both summaries.

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -335,8 +342,8 @@ Verify the package name is correct and the app is installed"</string>
 
     <!-- interaction.playall.playAllButtonPatch -->
     <string name="morphe_play_all_button_title">Show Play all button</string>
-    <string name="morphe_play_all_button_summary">"Tap to generate a playlist, tap and hold to undo
-
+    <string name="morphe_play_all_button_summary">"Tap to generate a playlist of all videos from channel
```
krvstek: Better to keep it as a single continuous paragraph.

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -365,12 +372,12 @@ Info: May not work on live streams"</string>
     <string name="morphe_save_to_watch_later_error_toast">Cannot save this video in watch later playlist</string>
 
     <!-- interaction.livering.openChannelOfLiveAvatarPatch -->
-    <string name="morphe_open_channel_of_live_avatar_title">Live avatar tap action</string>
-    <string name="morphe_open_channel_of_live_avatar_summary">Opens the channel page when enabled, or the live stream when disabled</string>
+    <string name="morphe_open_channel_of_live_avatar_title">Open channel of live avatar</string>
```
krvstek: The new title sounds grammatically clunky. Something like `Open channel on live avatar tap` or `Live avatar opens channel` would sound much more natural and clear.

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -523,8 +534,9 @@ Info: May not work on live streams"</string>
     <string name="morphe_show_settings_button_title">Show Settings button</string>
     <string name="morphe_show_settings_button_index_title">Settings button position</string>
     <string name="morphe_show_settings_button_type_title">Settings button action</string>
-    <string name="morphe_show_settings_button_type_summary">Opens Morphe settings when enabled, or YouTube settings when disabled</string>
-    <string name="morphe_swap_create_with_notifications_button_title">Swap Create with Notifications</string>
+    <string name="morphe_show_settings_button_type_summary">"OFF: Open YouTube settings
```
krvstek: The old conversational summary was much cleaner and more natural. Hardcoding ON/OFF states into a multi-line summary just bloats the UI unnecessarily.

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -557,7 +569,8 @@ To access Settings, enable the Settings button via General → Toolbar → Show
     <string name="morphe_show_toolbar_settings_button_title">Show Settings button</string>
     <string name="morphe_show_toolbar_settings_button_index_title">Settings button position</string>
     <string name="morphe_show_toolbar_settings_button_type_title">Settings button action</string>
-    <string name="morphe_show_toolbar_settings_button_type_summary">When enabled, tap opens Morphe settings and long press opens YouTube settings, disable to swap</string>
+    <string name="morphe_show_toolbar_settings_button_type_summary">"OFF: Tap opens YouTube settings and long press opens Morphe settings
```
krvstek: Same as [R537 line](https://github.com/MorpheApp/morphe-patches/pull/1485/changes#r3253037739).

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -1189,7 +1202,7 @@ Tap here to learn more about DeArrow"</string>
 
     <!-- misc.dimensions.spoof.spoofDeviceDimensionsPatch -->
     <string name="morphe_spoof_device_dimensions_title">Spoof device dimensions</string>
-    <string name="morphe_spoof_device_dimensions_summary">Tricks the server into providing higher quality options by spoofing device dimensions</string>
+    <string name="morphe_spoof_device_dimensions_summary">May unlock higher video qualities</string>
```
krvstek: The new summary is a bit too vague. While the old one was slightly redundant, it actually explained the mechanism. A middle ground like `Tricks the server into unlocking higher video qualities` would be much better.

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -1214,19 +1227,19 @@ Tap here to learn more about DeArrow"</string>
 
     <!-- misc.links.bypassURLRedirectsPatch -->
     <string name="morphe_bypass_url_redirects_title">Bypass URL redirects</string>
-    <string name="morphe_bypass_url_redirects_summary">Opens destination URLs directly, bypassing redirect tracking links</string>
+    <string name="morphe_bypass_url_redirects_summary">Opens the original URL directly, bypassing the tracking link (youtube.com/redirect)</string>
 
     <!-- misc.links.openLinksExternallyPatch -->
     <string name="morphe_external_browser_title">Open links in browser</string>
-    <string name="morphe_external_browser_summary">Opens links in your browser instead of the in-app browser</string>
+    <string name="morphe_external_browser_summary">Disables the in-app browser (Custom Tabs)</string>
```
krvstek: The new summary only focuses on what it disables. It's better to explicitly state that it opens links in an external browser, just like the old version did.

patches/src/main/resources/addresources/values/youtube/strings.xml
kitadai31: It's the name of the patch tho

https://github.com/MorpheApp/morphe-patches/blob/d239de3f2491826b94a466ad0a33e9d41b57bd91/patches/src/main/kotlin/app/morphe/patches/youtube/layout/livering/OpenChannelOfLiveAvatarPatch.kt#L59-L62

How about `Open channel from live avatar` then?

krvstek: UI strings don't necessarily have to match the internal patch names perfectly, but `Open channel from live avatar` sounds good.

LisoUseInAIKyrios: The patch could be renamed to the same. I think "Open channel from live avatar" reads better.

patches/src/main/resources/addresources/values/youtube/strings.xml
kitadai31: Indeed, this was too vague.

However, I think that sentence is too complicated. I don't think it's necessary to mention servers here.
The description for this feature can be more concise.

I'd like to suggest these four strings

1. `Higher video qualities might be unlocked`
2. `Enabling this can unlock higher video qualities`
3. `Unlocks higher video qualities`
4. `This may unlock higher video qualities`

1 is from old summary_on text
2 is from old summary_off text
3 and 4 is something I just came up with.

In 3, the expression indicating possibility that was present in the other two is lost, but the word "higher" is used, implying that the best quality is not guaranteed to be unlocked.
However, I can already imagine that users enable this feature easily and then complain that "8K resolution doesn't appear".

4 is almost the same as my initial suggestion, but with the addition of the noun "This". I think this makes the grammar more natural.

Which one do you think is best? Also, do you have any other concise description ideas?

krvstek: I think `Enabling this can unlock higher video qualities` hits the sweet spot. It's concise, sounds like an active switch description, and doesn't promise 100% success.

patches/src/main/resources/addresources/values/youtube/strings.xml
kitadai31: Done.
However, I liked how you rephrased the terms in the summary.

e.g.
> Title: Haptic feedback
Summary: Choose which interactions trigger **device vibration**

Therefore, I also liked that you rephrased "external downloader" as "third-party downloading app".
This should help beginners understand the meaning of "external downloader".

The problem was the place where you rephrased it.
You were only using this paraphrase in the summary of "Override action button", so it felt like it was referring to something different from the external downloader.
In the settings that configure the external downloader itself, I believe it would be acceptable to use the rephrased explanation “third-party downloading app”.

Therefore, I actually want to revert this back to "third-party downloading app".

krvstek: Initially, I just wanted to keep the terms consistent, but your reasoning makes sense. Feel free to handle it your way, I'll leave the final call to you.

patches/src/main/resources/addresources/values/youtube/strings.xml
kitadai31: Settings renaming is done.

> UI strings don't necessarily have to match the internal patch names perfectly, but Open channel from live avatar sounds good.

Ultimately, Morphe Settings is for dynamically enabling/disabling patches without having to rebuild the app. Therefore, in my opnion, the setting name should be as close as possible to the patch name. (For the patches which have only one simple boolean toggle switch setting)

> The patch could be renamed to the same. I think "Open channel from live avatar" reads better.

Since this involves renaming variabls, it's numerous changes, may I leave it out of this PR and address it later?

patches/src/main/resources/addresources/values/youtube/strings.xml
krvstek, line 1026: How about this: `Enabling this can restore legacy UI features`?

krvstek: We should also probably look at `morphe_spoof_video_streams_summary` too, in `shared-youtube/strings.xml`.

How about making it a single, clean paragraph? Something like: `Fixes playback issues for most users by requesting a compatible video stream. May not be required for YouTube Premium users`

patches/src/main/resources/addresources/values/youtube/strings.xml
LisoUseInAIKyrios: > Since this involves renaming variabls, it's numerous changes, may I leave it out of this PR and address it later?

We could change just the patch name declaration and add a TODO to update the internal code names later.

Or we do it all later. Either way.

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -242,7 +244,11 @@ Showing YouTube Doodles may prevent custom headers from displaying."</string>
     <string name="morphe_hide_keyword_content_phrases_title">Keywords to hide</string>
     <!-- For localization, it is preferred, but not required, if 'LeBlanc' is replaced with a localized name or a familiar word that has upper case letters in the middle of the word.
          This is because keywords can be in any language, and showing an example in the localized script helps convey this. -->
-    <string name="morphe_hide_keyword_content_phrases_summary">Enter keywords or channel names, separated by new lines. Words with uppercase letters in the middle must match exactly (ie: iPhone, TikTok, LeBlanc)</string>
+    <string name="morphe_hide_keyword_content_phrases_summary">"Enter keywords and phrases to hide, separated by new lines
```
kitadai31, line 247: I don't think this setting needs to be made compact.
This string will appear in the PreferenceScreen dedicated to keyword filtering, and this setting is the main focus of that screen.
Therefore, I don't think there's a problem if that summary is long.
This setting should be explained in detail, even if it means using up some UI space.

Long texts are easier to read when they are divided into paragraphs.

patches/src/main/resources/addresources/values/youtube/strings.xml
kitadai31: Why does that make the UI bloated?
Your text will wrap to two lines anyway, so it's no different than adding line breaks beforehand.
The number of characters has also been reduced.

Let me explain the intention behind this new notation.

The reason I introduced this new expression is because this type of setting should ideally be a `ListPreference` in the first place.
For example, in this case, this setting should ideally be a ListPreference like this.

ListPreference

This is because choosing an action should involve selecting between two options, rather than turning a switch ON or OFF.

However, since ListPreference makes the code more complicated, this kind of settings are implemented abusing SwitchPreference in Morphe.

So, I figured that by listing the two available options on separate lines in the summary, we could replicate a ListPreference.
This allows the user to choose from two options, similar to the experience of ListPreference.
Therefore, I am confident that this is superior to a conversational summary.

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -72,16 +72,16 @@ Changes include:
     <string name="morphe_hide_community_posts_title">Hide community posts</string>
     <string name="morphe_hide_compact_banner_title">Hide compact banners</string>
     <string name="morphe_hide_compact_banner_summary">Hides compact banners and spacing between some components</string>
-    <string name="morphe_hide_expandable_card_title">Expandable cards under videos</string>
-    <string name="morphe_hide_expandable_card_entry_1">Show all</string>
+    <string name="morphe_hide_expandable_card_title">Hide expandable cards under videos</string>
+    <string name="morphe_hide_expandable_card_entry_1">Disabled</string>
```
ILoveOpenSourceApplications:
```suggestion
    <string name="morphe_hide_expandable_card_entry_1">Show chapters, product, and summary cards</string>
```

patches/src/main/resources/addresources/values/youtube/strings.xml
krvstek: If it's on a dedicated screen, then taking up vertical space isn't really an issue. Fair enough, let's drop the topic then!

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -72,16 +72,16 @@ Changes include:
...
     <string name="morphe_hide_expandable_card_entry_2">Hide product card only</string>
     <string name="morphe_hide_expandable_card_entry_3">Hide summary card only</string>
     <string name="morphe_hide_expandable_card_entry_4">Hide product and summary cards</string>
     <string name="morphe_hide_expandable_card_entry_5">Hide all</string>
```
ILoveOpenSourceApplications, line 80:
```suggestion
    <string name="morphe_hide_expandable_card_entry_5">Hide chapters, product,  and summary cards</string>
```

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -105,11 +105,13 @@ Changes include:
     <string name="morphe_hide_show_more_button_summary">Hides the expand button under videos in the search results</string>
     <string name="morphe_hide_subscribed_channels_bar_title">Hide subscribed channels bar</string>
     <string name="morphe_hide_surveys_title">Hide surveys</string>
+    <string name="morphe_hide_surveys_summary">Hides surveys such as \'Is the above video a good recommendation for you?\'</string>
```
ILoveOpenSourceApplications: Considering this is only one type of survey mentioned, can a general summary be used instead? 

Otherwise we'll have to find the names of all the different surveys being hidden and mention them as well, which I feel is a bit too much.

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -704,7 +720,7 @@ Settings → Playback (or Autoplay) → Autoplay next video"</string>
 
     <!-- layout.hide.relatedvideooverlay.hideRelatedVideoOverlayPatch -->
     <string name="morphe_hide_player_related_videos_overlay_title">Hide related videos overlay</string>
```
ILoveOpenSourceApplications:
```suggestion
    <string name="morphe_hide_player_related_videos_overlay_title">Hide \'More videos\' overlay</string>
```

patches/src/main/resources/addresources/values/youtube/strings.xml
krvstek: I get the logic behind wanting to simulate a `ListPreference`.

However, visually it's still a Switch, and native Android settings usually use conversational sentences for Switch summaries rather than `ON/OFF` lists, which can look a bit like a debug menu.

patches/src/main/resources/addresources/values/youtube/strings.xml
krvstek: I think the old approach was actually more intuitive. Changing `Show all` to `Disabled` under a `Hide` title forces the user to process a double negative (Disable hiding = Show).

Also, if we change the title to `Hide expandable cards...`, then repeating `Hide...` in every single dropdown entry makes it redundant. It might be better to keep the noun title (`Expandable cards under videos`) and keep the explicit actions (`Show all`, `Hide all`, etc.) in the list.

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -637,14 +651,14 @@ To show the \'Audio Track\' menu, change \'Spoof video streams\' to \'Android Re
...
-    <!-- Seekbar shown inside video thumbnails found on the Home / Subscriptions / Search / History.  The seekbar shows the prior watch progress from when the video was last opened. -->
     <string name="morphe_hide_seekbar_thumbnail_title">Hide video thumbnails seekbar</string>
-    <string name="morphe_hide_seekbar_thumbnail_summary">Hides the watch progress bar shown on video thumbnails in feeds and search results</string>
+    <string name="morphe_hide_seekbar_thumbnail_summary">Hides the watch progress bar shown on thumbnails of videos you have already watched</string>
     <string name="morphe_fullscreen_large_seekbar_title">Enable fullscreen large seekbar</string>
 
     <!-- layout.hide.shorts.hideShortsComponentsPatch -->
     <string name="morphe_shorts_player_screen_title">Shorts player</string>
     <string name="morphe_shorts_player_screen_summary">Choose which components appear in the Shorts player</string>
+    <string name="morphe_preference_category_hide_shorts">Hide Shorts</string>
```
krvstek: I don't think we use `preference_category` in string names anywhere else in the project. It breaks the naming consistency. 

Could we rename this key to something cleaner like `morphe_hide_shorts_category_title` to match the rest of the file?

patches/src/main/resources/addresources/values/youtube/strings.xml
kitadai31: Thank you for your understanding!

patches/src/main/resources/addresources/values/youtube/strings.xml
kitadai31: Thanks for a good catch
I missed other usage of a preference category

patches/src/main/resources/addresources/values/youtube/strings.xml
kitadai31: I'll remove the added summary and instead add a comment for translators.

The reason I gave a specific example here was to distinguish it from surveys that creators conduct in community posts.
This was particularly noticeable in my language. Since the title alone didn't provide enough detail about the functionality, it was translated using vague words in my language.
I thought one example would be sufficient for that purpose.

However, thinking about it again now, this could be solved by adding a comment for translators.
If the translator can choose the appropriate words, the issue can be resolved without displaying a summary to the user.

Furthermore, since the surveys conducted by creators are just community posts, it's obvious that they will disappear by "Hide community posts", not "Hide surveys".
Perhaps I was the only Japanese speaker who had this misunderstanding. 😅
Sorry for my unfounded worry.

patches/src/main/resources/addresources/values/youtube/strings.xml
kitadai31: Looking at the implementation, the "Hide all" option hides all expandable cards unconditionally.
Considering the possibility of adding different types of expandable cards in the future, I think it's more appropriate to call this option `Hide all`.
Listing all the names would create future problems.

If you want to let users know that "Hide all" also hides the chapters card, how about something like `Hide all (includes chapters)`?

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -1212,19 +1228,19 @@ Tap here to learn more about DeArrow"</string>
 
     <!-- misc.links.bypassURLRedirectsPatch -->
     <string name="morphe_bypass_url_redirects_title">Bypass URL redirects</string>
-    <string name="morphe_bypass_url_redirects_summary">Opens destination URLs directly, bypassing redirect tracking links</string>
+    <string name="morphe_bypass_url_redirects_summary">Opens the original URL directly, bypassing the tracking link (youtube.com/redirect)</string>
 
     <!-- misc.links.openLinksExternallyPatch -->
     <string name="morphe_external_browser_title">Open links in browser</string>
-    <string name="morphe_external_browser_summary">Opens links in your browser instead of the in-app browser</string>
+    <string name="morphe_external_browser_summary">Opens links in external browser instead of the in-app browser</string>
```
krvstek:
```suggestion
    <string name="morphe_external_browser_summary">Opens links in an external browser instead of the in-app browser</string>
```

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -808,13 +824,13 @@ Settings → Playback (or Autoplay) → Autoplay next video"</string>
     <string name="morphe_sb_create_segment_category">Creating new segments</string>
     <string name="morphe_sb_enable_create_segment">Show \'Create new segment\' button</string>
     <string name="morphe_sb_enable_create_segment_summary">Adds a button to the video player allowing you to submit new SponsorBlock segments</string>
-    <string name="morphe_sb_general_adjusting">Adjust new segment step</string>
-    <string name="morphe_sb_general_adjusting_summary">Set the number of milliseconds for the time adjustment buttons when creating new segments</string>
+    <string name="morphe_sb_general_adjusting">Time adjustment buttons step</string>
+    <string name="morphe_sb_general_adjusting_summary">Number of milliseconds the time adjustment buttons move</string>
```
krvstek: This feels a bit clunky grammatically:

- The title `Time adjustment buttons step` is a bit of a mouthful. How about just `Time adjustment step`?
- The summary ends a bit abruptly. It would sound more natural with `by` at the end, or phrased as an action: `Sets how many milliseconds the time adjustment buttons skip by`.

krvstek:
Overall, great work on these improvements!

Aside from the inline comments and tweaks suggested by me and others, the rest of the PR looks solid. Once those minor things are addressed, it's LGTM.

patches/src/main/resources/addresources/values/youtube/strings.xml
kitadai31: I changed it to a general summary

kitadai31:
I'm ready on my end.
I've reverted some of the strings that were problematic.

LisoUseInAIKyrios:
Will merge tomorrow if there are no new concerns.

PR #1497, opened 2026-05-17
fix(YouTube - Open channel of live avatar): Fixed null exception after tap on some shorts player buttons

0xrxL:


error-reporting:
the original exception that would need the fix is still better than an potentially ineffective open channel from live avatar patch that the fix could cause

0xrxL:
> the original exception that would need the fix is still better than an potentially ineffective open channel from live avatar patch that the fix could cause

???

error-reporting:
just want to say that the fixes especially null checks that return false and remove of android.text.TextUtils import seen in this pr may in certain conditions make the patch ineffective

LisoUseInAIKyrios:
The null check means there was no value.

The removal of android.text.TextUtils import is because the import was not used.

There is no problem.

0xrxL:
> just want to say that the fixes especially null checks that return false and remove of android.text.TextUtils import seen in this pr may in certain conditions make the patch ineffective

Don't worry. I removed the import because no istructions in the class use it. And the null check help to prevent the buttons, that are not a live avatar, to being recognized as it.

error-reporting:
thank you guys for answering my concerns

PR #1511, opened 2026-05-19
chore(YouTube - Open channel of live avatar): deprecate match with translations

### LINKED ISSUE #1458: bug(YouTube): 'Open channel of live avatar' may not work when app is using non English language
### Bug description

'Open channel of live avatar' may not work when app is using non English language

### App version patched

20.47.62

### Error logs

```shell

```

### Acknowledgements

0xrxL:


0xrxL:
@LisoUseInAIKyrios Maybe I found a stable way to make it works:

Every live avatar element have a ComponentHost parent (the red circle with the transparent center, that overlap the channel avatar). This is now the new match.

And, as a further check, I also verified the width and the height of the element, that are both same and not higher of a certain pixel density compared to thumbnails.

PR #1536, opened 2026-05-23
fix(YouTube - Sanitize sharing links): Sanitize system share sheet

### LINKED ISSUE #1454: bug(YouTube): "Sanitize sharing links" does not work when "Open system share sheet" is enabled
## Bug description

When both `Open system share sheet` and `Sanitize sharing links` patches are enabled, the share URL is **not sanitized** and is delivered to the system share sheet in YouTube's new `sharing/invite/TOKEN?w=VIDEO_ID` invite-link format. This happens regardless of the `Sanitize sharing links` configuration, including with the new "Change Share links to shortened URLs" option (#1346) enabled.

Interestingly, the sanitization **does work** when accessed through a different route: tap Share → use the internal YouTube share panel → scroll right → tap "More" → system share sheet. In this case the URL is correctly sanitized to `youtube.com/watch?v=VIDEO_ID&is=...`.

So it looks like the two share routes behave differently — only the indirect one (via the in-app panel) gets sanitized. The direct system share sheet route delivered by `Open system share sheet` does not.

Context: this is related to YouTube's new "Video sharing and messaging" feature (https://support.google.com/youtube/answer/16650958), now rolling out beyond the announced Ireland/Poland regions. The invite-link URL is unwanted when sharing to external apps (WhatsApp, browsers, third-party services) since those don't use YouTube's internal messaging.

## Steps to reproduce

1. Patch YouTube with Morphe Patches v1.27.0 with these patches enabled:
   - `Sanitize sharing links` (with "Change Share links to shortened URLs" sub-option enabled)
   - `Open system share sheet`
2. Open any YouTube video
3. Tap Share — system share sheet opens directly
4. Check the shared text — it is: Let's share videos and message on YouTube! https://youtube.com/sharing/invite/TOKEN?w=VIDEO_ID instead of a sanitized `youtu.be/VIDEO_ID` or `youtube.com/watch?v=VIDEO_ID`.

## Expected behavior

The URL delivered to the system share sheet should be sanitized when `Sanitize sharing links` is enabled, regardless of which share route is used.

## Actual behavior

The `sharing/invite/TOKEN?w=VIDEO_ID` URL is delivered unsanitized when `Open system share sheet` is enabled. Sanitization only fires on the indirect "More" route via the in-app share panel.

## Affected versions

- Morphe Patches: **1.27.0**
- YouTube Morphe: **20.47.62**
- Android: **16**
- Device: **Samsung Galaxy S25 Ultra**

## Configuration tested

- ✅ YouTube account setting "Channel visibility for shared links" — **OFF**
- ✅ `Sanitize sharing links` — enabled
- ✅ Sub-option "Change Share links to shortened URLs" — enabled
- ✅ `Open system share sheet` — enabled
- ❌ Result: invite URL still delivered to system share sheet, unmodified

Workaround that works: disable `Open system share sheet` and use the internal share panel → More → system share sheet. URL is sanitized correctly. But this adds friction to every share.

## Attachments

See screenshots:
- Screenshot 1: system share sheet via `Open system share sheet` ON — invite URL is delivered
- Screenshot 2: internal YouTube share panel (alternative route shown)
- Screenshot 3: system share sheet reached via "More" — URL correctly in `watch?v=` format

## Checklist

- [x] This issue does not reproduce on unpatched YouTube (unpatched YouTube generates the same invite URL, but the issue here is that the patch silently fails to sanitize it)
- [x] This issue is not a duplicate of an existing bug report (searched open and closed issues)
- [x] I have chosen an appropriate title
- [x] All requested information has been provided properly
- [x] I have written the title and contents in English

Image

Image

Image

### App version patched

- Morphe Patches: **1.27.0** - YouTube Morphe: **20.47.62**

### Error logs

```shell

```

### Acknowledgements

0xrxL:


patches/src/main/resources/addresources/values/shared-youtube/strings.xml
```diff
@@ -125,5 +123,7 @@ To sign in, press continue below"</string>
     <string name="morphe_sanitize_sharing_links_summary">Removes tracking query parameters from shared links</string>
     <string name="morphe_replace_music_with_youtube_title">Change share links to youtube.com</string>
     <string name="morphe_replace_links_with_shortener_title">Change share links to youtu.be</string>
-    <string name="morphe_replace_links_with_shortener_summary">Forces live streams and Shorts to use the shortened youtu.be URL format</string>
+    <string name="morphe_replace_links_with_shortener_summary">Forces videos to use the shortened youtu.be URL format</string>
+    <string name="morphe_replace_invites_with_video_shortener_title">Change invite links to video youtu.be</string>
+    <string name="morphe_replace_invites_with_video_shortener_summary">Convert invites to a video URL and forced shortened youtu.be format</string>
```
LisoUseInAIKyrios: Is there a benefit to shortening one type of URL but not the other?

I think this could be part of the existing `morphe_replace_links_with_shortener` and not an additional setting.

0xrxL: > Is there a benefit to shortening one type of URL but not the other?
> 
> I think this could be part of the existing `morphe_replace_links_with_shortener` and not an additional setting.

Sorry for previous wrong answer.

This is the URL for invites. From what I understood, the user asked for a way to convert this link to an only video link. But this require a separate setting, to prevent to lose the invites feature.

LisoUseInAIKyrios: The invite feature is user data, so to sanitize the link requires removing the invite.

PR #1544, opened 2026-05-23
 fix(YouTube - Open channel of live avatar): Improved check to exclude the live avatar in the channel header

0xrxL:


0xrxL:
@LisoUseInAIKyrios This is a fix for foldables, but it require to get displayMetrics properties at every code execution, to know if the foldable is currently used with external or internal display. Check if you can add a method that respect this requirement in Dim.

0xrxL:
P.S: I removed the patch to detect if the channel avatar is currently opened, because it not much reliable (a specific flag can break it).

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/OpenChannelOfLiveAvatarPatch.java
```diff
@@ -67,41 +65,43 @@ public static boolean openChannel(Map<Object, Object> playbackStartDescriptorMap
...
+            // height and width, compared to thumbnails or the header channel avatar.
             // This check will avoid any false positives.
             final int width = componentHost.getWidth();
             final int height = componentHost.getHeight();
-            if (width == 0 || width != height || width > Dim.SCREEN_WIDTH / 4) {
-                    return false;
+            // The getDisplayMetrics() properties must be retrieved dynamically to avoid false positives when
+            // switching between the inner and outer screens (or vice versa) on foldable devices.
+            DisplayMetrics currentMetrics = componentHost.getContext().getResources().getDisplayMetrics();
+            boolean isLandscapeOrTablet = currentMetrics.widthPixels > currentMetrics.heightPixels;
```
LisoUseInAIKyrios, line 76: Can use `Dim` static field

https://github.com/MorpheApp/morphe-patches-library/blob/34345ce85aec8bb8283ab156697ddefc54e22aec/extension-library/src/main/java/app/morphe/extension/shared/ui/Dim.java#L14-L15

LisoUseInAIKyrios (approved):

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/OpenChannelOfLiveAvatarPatch.java
0xrxL: Does Dim static field constantly update or just when the app starts? Because in this case it needs to be constantly updated.

LisoUseInAIKyrios: Ok I see the issue.

No it's not constantly updated.

Can fix Dim later.

Issue #1555, opened 2026-05-25
bug(YouTube): UI not responsive when "skip" button or "undo skip" button is shown


gwolf2u:
### Bug description

When sponsorblock toast notification is visible, can't click anything in UI until disappears (times out).
This happens on YT 21.21.80 with patches 1.30.0-dev.2
Happens with some older YouTube versions also.

### App version patched

YouTube 21.21.80

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

gwolf2u:
https://github.com/user-attachments/assets/05c2f77c-da9f-4231-bed0-7ea386e02b22

MarcaDian:
This should fix the problem.
<img width="995" height="271" alt="Image" src="https://github.com/user-attachments/assets/a3ff0f27-a0bf-4c5f-a4a1-ee1d924fb1a1" />

0xrxL:
@gwolf2u Try to patch with this local source, and check if you can reproduce the issue again:

https://github.com/MorpheApp/morphe-patches/actions/runs/26403694071/artifacts/7199406614

gwolf2u:
> [@gwolf2u](https://github.com/gwolf2u) Try to patch with this local source, and check if you can reproduce the issue again:
> 
> https://github.com/MorpheApp/morphe-patches/actions/runs/26403694071/artifacts/7199406614

Nope, didn't work
Patched locally and same result
Running this on Android 15 if matters and using attached settings (if needed) with below video
https://youtu.be/haJ6d5DCrDU?is=MuQxHbsgz5DF245K

[YouTube_Morphe_Settings_2026-05-25.txt](https://github.com/user-attachments/files/28232430/YouTube_Morphe_Settings_2026-05-25.txt)

0xrxL:
@gwolf2u Follow this and see if you can fix your issue:

https://www.reddit.com/r/MorpheApp/comments/1tls6df/youtube_issues_and_google_ab_tests_flags/

gwolf2u:
> [@gwolf2u](https://github.com/gwolf2u) Follow this and see if you can fix your issue:
> 
> https://www.reddit.com/r/MorpheApp/comments/1tls6df/youtube_issues_and_google_ab_tests_flags/

Disabled all flags with 453, 456 and even 457
Restarted 
Does exactly the same

[...]

PR #1556, opened 2026-05-25
fix(YouTube - SponsorBlock): Prevent undo skip toast from stealing window focus

MarcaDian:
Closes https://github.com/MorpheApp/morphe-patches/issues/1555

### Test results


MarcaDian:
I can't reproduce the problem, so I can't test if this will work, but it's highly likely.

0xrxL:
@MarcaDian This PR doesn't work, based on the user feedback.

PR #1557, opened 2026-05-25
feat(YouTube - Loop video): Add loop range via long-press on loop button

### LINKED ISSUE #1281: feat(YouTube): Playback loop range
### Feature description

Implement a selective timeframe loop for the current warching video, with a dedicated view where to enable the feature and two boxes where to put start and end minute.

### Motivation

When I watched some videos, especially music videos, I realized that not every has a beginning and an end without pauses or additions related to channel sponsorship (where sponsorblock doesn't work for a reason or another), or that there are simply some videos of movie clips related to specific scenes that are not properly cut.

### Acknowledgements

### LINKED ISSUE #1568: bug(YouTube): Video keeps looping at the end
### Bug description

https://github.com/user-attachments/assets/41b41753-6de5-4514-a03d-e72657619202

### App version patched

version 21.21.80

### Error logs

```shell

```

### Acknowledgements

MarcaDian:
### Description

Long-press the loop button to set a start and end time for looping.
A new icon indicates when a range is active. Short-press toggles loop on/off and clears any active range when turning off.

Closes https://github.com/MorpheApp/morphe-patches/issues/1281

### Test results

___

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/videoplayer/LoopVideoButton.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/videoplayer/LoopVideoButton.java
@@ -125,7 +125,7 @@ public class LoopVideoButton {
             final String startPrefill = rangeActive
                     ? formatTime(LoopVideoPatch.rangeStartMs)
                     : (currentTime > 0 ? formatTime(currentTime) : "");
-            final String endPrefill = (rangeActive && !LoopVideoPatch.rangeEndIsVideoEnd)
+            final String endPrefill = rangeActive
                     ? formatTime(LoopVideoPatch.rangeEndMs)
                     : "";
 
@@ -175,7 +175,7 @@ public class LoopVideoButton {
         }
 
         final long endMs;
-        final boolean endIsVideoEnd = endInput.trim().isEmpty();
+        boolean endIsVideoEnd = endInput.trim().isEmpty();
         if (endIsVideoEnd) {
             if (videoLengthMs <= 0) {
                 Utils.showToastShort(str("morphe_loop_video_range_invalid_time"));
@@ -183,10 +183,16 @@ public class LoopVideoButton {
             }
             endMs = videoLengthMs;
         } else {
-            endMs = parseTime(endInput);
-            if (endMs < 0) {
+            final long parsed = parseTime(endInput);
+            if (parsed < 0) {
                 Utils.showToastShort(str("morphe_loop_video_range_invalid_time"));
                 return;
+            }
+            if (videoLengthMs > 0 && parsed >= videoLengthMs) {
+                endMs = videoLengthMs;
+                endIsVideoEnd = true;
+            } else {
+                endMs = parsed;
             }
         }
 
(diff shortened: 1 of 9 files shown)


0xrxL:
@MarcaDian which is the issue with this?

MarcaDian:
Іf set the end time of the looping to the end of the video, then when it switch to the beginning of segment, the final video layout remains, that is, either recommended videos or the player state at the end.

0xrxL:
> Іf set the end time of the looping to the end of the video, then when it switch to the beginning of segment, the final video layout remains, that is, either recommended videos or the player state at the end.

So avoid it reaching the end of the video, by blocking it 1 second before. This will solve also the possible conflicts with stock/patched video loop routine.

0xrxL:
And also add a check for the minimum setable loop, and the PR will be ready imo.

MarcaDian:
> So avoid it reaching the end of the video, by blocking it 1 second before. This will solve also the possible conflicts with stock/patched video loop routine.

I already did that, but still the final layout is showing (10 seconds before the end of the video?)

0xrxL:
> > So avoid it reaching the end of the video, by blocking it 1 second before. This will solve also the possible conflicts with stock/patched video loop routine.
> 
> I already did that, but still the final layout is showing (10 seconds before the end of the video?)

Yeah...you're right. Final layout is visible more than few seconds early. In this case you can force the final layout visibility to off (there's already a patch for this), if the loop segment is enabled.

MarcaDian:
> you can force the final layout visibility to off

I already explained that I am not friends with bytecode)

MarcaDian:
Not sure, but it looks like it was fixed in the last commit.

MarcaDian:
It seems that all the problems are fixed. Can you test it? I will probably never use it so I am not interested in testing.
It is very long for me to wait 5-6 minutes every time just for the build.

0xrxL:
@ILoveOpenSourceApplications Test it please. We needs of more feedbacks.

ILoveOpenSourceApplications:
Looks good on my end, did a basic loop test and seems to work without any issues.

MarcaDian:
I've also tried it several times in different scenarios and haven't found any problems yet.

___
> and haven't found any problems

This is even strange to me 😅

MarcaDian:
Why did you decide that this fixes it? This is an unrelated PR.

Ah, I saw the user's response. That's strange.

extensions/youtube/src/main/java/app/morphe/extension/youtube/videoplayer/LoopVideoButton.java
```diff
@@ -97,72 +248,171 @@ private static void animateButtonTransition(View view, boolean newState) {
...
         }
+        return sb.toString();
+    }
+
+    private static LinearLayout buildFieldsLayout(Context context, EditText startField, EditText endField) {
+        LinearLayout container = new LinearLayout(context);
+        container.setOrientation(LinearLayout.VERTICAL);
+        container.setLayoutParams(new LinearLayout.LayoutParams(
+                ViewGroup.LayoutParams.MATCH_PARENT,
+                ViewGroup.LayoutParams.WRAP_CONTENT));
```
LisoUseInAIKyrios, line 374: The time editor dialog has no way to type `:` to specify a minute or hour.

MarcaDian: `:` is automatically inserted when the user enters numbers, numeric keypad only

Screenshot_2026-05-27-16-51-16-179_com google android youtube test-edit

PR #1558, opened 2026-05-25
feat: ifunny patches

rotatordisk92:
### Description

Adds three bytecode patches for iFunny (mobi.ifunny):

  - Hide ads — patches AdsDisableManagerImpl to always return true, suppressing all ad display checks
  - Unlock premium — patches User.isUserPremium() to always return true, enabling premium features
  - Save videos — patches SaveContentCriterion to always return true, unlocking video saving

All fingerprints target stable DEX class names sourced from Kotlin @Metadata annotations rather than jadx-renamed package names. Obfuscated methods are identified using structural filters (field access patterns and string literals) rather than method names, making them resilient to R8/ProGuard reruns across versions.

### Additional context

iFunny uses partial obfuscation — package/class names under mobi.ifunny.* are stable, but method names inside classes are minified to single letters. The fingerprint for each patch accounts for this: isUserPremium is unobfuscated and matched by name; the other two use structural bytecode filters unique to their target methods. See DEVELOPMENT.md in the patch directory for full rationale.

_(Tested on iFunny 10.38.11 and authored against 10.39.11)_

### DIFF
--- a/patches/src/main/kotlin/app/morphe/patches/ifunny/DEVELOPMENT.md
+++ b/patches/src/main/kotlin/app/morphe/patches/ifunny/DEVELOPMENT.md
@@ -1,97 +1,182 @@
 # iFunny Patches — Development Guide
 
+## Tested against
+
+**iFunny 10.39.11** (`mobi.ifunny`, minSdk 26). Fingerprints were authored and verified against
+this version. Use `--force` with morphe-cli to test on adjacent versions; re-verify fingerprints
+before bumping `Constants.kt`.
+
+---
+
 ## Patches
 
-| Patch | File | DEX class | Target method | Strategy |
-|-------|------|-----------|---------------|----------|
-| Hide ads | `ad/` | `Lmobi/ifunny/ads/criterions/AdsDisableManagerImpl;` | `g()` — public, no-arg boolean | `returnEarly(true)` |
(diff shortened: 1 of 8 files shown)


LisoUseInAIKyrios:
Thanks for the interest, but at this time we only accept patches for YouTube, YT Music, and Reddit.

You can release these patches under your own self published repo.

Clone or use this template, add your patches here, and generate your first release:

https://github.com/morpheapp/morphe-patches-template

rotatordisk92:
That's honestly terrible.

Morphe DMCA'd my self-published repository that was based on ReVanced, and the only way to bring it back was to rebase to morphe-patches, which I did, only to now find you don't support other apps. Why not add that in bold at the top of your README.md to discourage contributions and wasted efforts from the community?

LisoUseInAIKyrios:
The problem was RV copying our code in non compliant ways. You could add your repo back again and still use ReVanced if you wish, assuming you didn't pull in any of the offending code RV added in the month or two before GitHub permanently removed the RV patches repo.

We will update the contributing guide here that is still listed as a TODO.

https://github.com/MorpheApp/morphe-patches/blob/main/CONTRIBUTING.md


We help third party Morphe devs if they wish. Send us a mod mail on our subreddit if you have any development questions.

PR #1604, opened 2026-05-30
chore(Reddit): Simplify settings UI

krvstek:
### Description

N/A

### Additional context

N/A

### Test results

### DIFF
--- a/extensions/reddit/src/main/java/app/morphe/extension/reddit/settings/preference/BooleanSettingPreference.java
+++ b/extensions/reddit/src/main/java/app/morphe/extension/reddit/settings/preference/BooleanSettingPreference.java
@@ -11,6 +11,7 @@ import static app.morphe.extension.shared.StringRef.str;
 import android.content.Context;
 import android.preference.SwitchPreference;
 
+import app.morphe.extension.shared.ResourceUtils;
 import app.morphe.extension.shared.settings.BooleanSetting;
 
 @SuppressWarnings("deprecation")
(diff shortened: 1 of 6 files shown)


extensions/reddit/src/main/java/app/morphe/extension/reddit/settings/preference/BooleanSettingPreference.java
```diff
@@ -19,7 +19,11 @@ public class BooleanSettingPreference extends SwitchPreference {
     public BooleanSettingPreference(Context context, BooleanSetting setting) {
         super(context);
         this.setTitle(str(setting.key + "_title"));
-        this.setSummary(str(setting.key + "_summary"));
+        String summaryKey = setting.key + "_summary";
+        int summaryResId = context.getResources().getIdentifier(summaryKey, "string", context.getPackageName());
```
LisoUseInAIKyrios: Can use resource utils 

https://github.com/MorpheApp/morphe-patches-library/blob/34345ce85aec8bb8283ab156697ddefc54e22aec/extension-library/src/main/java/app/morphe/extension/shared/ResourceUtils.java#L153

kitadai31:
In my opinion, the information about where the item that the setting hides is necessary.

Omitting of the description should only be done if the title is truly self-explanatory.
Users should be able to learn about the details of each feature without having to search the wiki, ask questions on r/MorpheApp, or check the actual implementation in the source code.

For example, `Auto-dismiss notification popups` is the most unclear.
What is "notification popup?"
An explanation is needed regarding what a “notification popup” is and where it appears.

That being said, I don’t think there’s nothing to improve about the current Reddit settings.
An idea I had been thinking about recently is to create new "Navigation bar" category and "Sidebar" category, and separate them from the existing "Layout" category.
If we do this, we can remove all summaries from the settings related to navigation bar and sidebar.

Other than those two categories, we can remove summaries from the Ads settings and the "Disable modern home" setting.
I think all other summaries other besides those I mentioned above should be restored.

PR #1615, opened 2026-05-31
fix(YouTube - Theme): Apply dark text color to "Material You" notification dot count

ILoveOpenSourceApplications:
### Description

N/A

### Additional context

N/A

### Test results

### DIFF
--- a/patches/src/main/kotlin/app/morphe/patches/shared/layout/theme/BaseThemePatch.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/shared/layout/theme/BaseThemePatch.kt
@@ -3,14 +3,11 @@ package app.morphe.patches.shared.layout.theme
 import app.morphe.patcher.patch.BytecodePatchBuilder
 import app.morphe.patcher.patch.BytecodePatchContext
 import app.morphe.patcher.patch.PatchException
-import app.morphe.patcher.patch.ResourcePatchContext
 import app.morphe.patcher.patch.bytecodePatch
 import app.morphe.patcher.patch.resourcePatch
 import app.morphe.patcher.patch.stringOption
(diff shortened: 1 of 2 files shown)


ILoveOpenSourceApplications:
@EDM115, can you download this [artifact](https://github.com/MorpheApp/morphe-patches/actions/runs/26720582409/artifacts/7318980633) and let me know if that fixes your issue?

EDM115:
fixed in the actual notifications bell position but not with the setting that allows it to be in the navbar :

ILoveOpenSourceApplications:
The bottom navigation button dot should also be following Material You. @MarcaDian, how to tackle this one?

MarcaDian:
Is it because you deleted the resource files? I never get notifications on YouTube.

ILoveOpenSourceApplications:
@EDM115, was the theme for the notification dot on the nav bar fine before you tested the artifact?

EDM115:
it seems to have self corrected ? didn't apply any new patches.

ILoveOpenSourceApplications:
I noticed this earlier, but how did you have two notification buttons, at the top and bottom at the same time.

EDM115:
it's this combination of settings :

ILoveOpenSourceApplications:
Yes, but still you should only have one, not two though.

MarcaDian:
So the problem is solved? I can't check this because I don't have notifications on YouTube at all

EDM115:
i'd consider it solved yes !

ILoveOpenSourceApplications:
Seems that way. For some reason the user had two notification buttons, which shouldn't be possible and may have caused the Material You theme not getting applied onto one of them.

MarcaDian (approved):

Issue #1626, opened 2026-06-01
feat(YouTube): Retry native PiP when background playback stays audio-only


Steven13799:
## Problem

On the affected setup, YouTube keeps background playback active after HOME, but the app does not transition into a visible native PiP window. The result is audio-only playback even though regular video playback is still active.

## Current direction

After maintainer feedback in #1628, this should be handled as a native PiP/background playback fix rather than by maintaining a custom overlay window.

The PR now targets the existing `Remove background playback restrictions` patch and retries native Android PiP at the end of `MainActivity.onUserLeaveHint()` only when the normal YouTube path did not already enter PiP.

## Scope

The native PiP retry is guarded so it only runs when:

- `Remove background playback restrictions` is enabled.
- Android supports `PictureInPictureParams`.
- The activity is not already in PiP.
- Shorts is not open.
- The current player is a regular player, not inline feed playback.

## Safety / privacy constraints

No APKs, logs, screenshots, keystores, API keys, account data, cookies, or tokens are included. The current PR does not add overlay permissions, services, network calls, or account access.

## Related

PR: #1628
Related public issues: #994, #444, #1196.

Steven13799:
Opened draft implementation PR: #1628. It ports the local proof of concept into a Morphe patch with an optional setting, Shorts guards, non-focusable overlay service, and local compile validation for :patches:compileKotlin and :extensions:youtube:compileDebugJavaWithJavac.

Steven13799:
Updated #1628 for review: the branch is now a single implementation commit, removes the draft docs from the final diff, declares the media playback foreground service type, and passes local validation for :patches:compileKotlin and :extensions:youtube:compileDebugJavaWithJavac. The PR is marked ready for review and is currently mergeable.

PR #1628, opened 2026-06-01
fix(YouTube - Remove background playback restrictions): Background playback is audio only

Steven13799:
### Problem

YouTube background/Shorts PiP is now controlled by additional YouTube-side code. The previous overlay/native-PiP-retry approaches were dropped after maintainer feedback because the preferred fix is to force the relevant YouTube control-flow flags, matching the existing background playback patch style.

### Description

Adds a new background playback feature flag override:

- Adds `PipAutoEnterFeatureFlagFingerprint` for literal `45685828L`.
- Forces that flag off through the existing `addBackgroundPlaybackFeatureFlagHook(false)` path.
- Keeps the change inside `Remove background playback restrictions`.
- Builds on #1629 / current `dev`, including the corrected Shorts background playback hook.

This PR no longer adds overlay permissions, services, UI controls, or a direct `enterPictureInPictureMode` retry.

### Test results

- [x] Rebased on `dev` after #1629.
- [x] `:patches:compileKotlin`.
- [x] Static diff/privacy check: only two Kotlin background playback files changed.

### On-device notes

On the problem device, previous local experiments showed that forcing `45685828` on/off alone did not restore native PiP, and forcing `45685828` + `45408908` off together also did not restore PiP. The reliable signal on that device was that system PiP never became active (`mLastReportedPictureInPictureMode=false`).

Keeping this PR as draft until the flag direction and combined behavior are verified on a device where native PiP works, per maintainer guidance.

### DIFF
--- a/patches/src/main/kotlin/app/morphe/patches/youtube/misc/backgroundplayback/BackgroundPlaybackPatch.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/youtube/misc/backgroundplayback/BackgroundPlaybackPatch.kt
@@ -133,5 +133,8 @@ val backgroundPlaybackPatch = bytecodePatch(
             // must also be disabled, otherwise the player is a black screen with no buttons and no playback.
             NewPlayerOverlaysFeatureFlagFingerprint.addBackgroundPlaybackFeatureFlagHook(false)
         }
+
+        // New YouTube PiP auto-enter flag in the same controller as the PiP input consumer flag above.
+        PipAutoEnterFeatureFlagFingerprint.addBackgroundPlaybackFeatureFlagHook(false)
     }
 }
--- a/patches/src/main/kotlin/app/morphe/patches/youtube/misc/backgroundplayback/Fingerprints.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/youtube/misc/backgroundplayback/Fingerprints.kt
@@ -111,3 +111,9 @@ internal object NewPlayerOverlaysFeatureFlagFingerprint : Fingerprint(
         literal(45752335L)
     )
 )
+
+internal object PipAutoEnterFeatureFlagFingerprint : Fingerprint(
+    filters = listOf(
+        literal(45685828L)
+    )
+)


LisoUseInAIKyrios:
What problem is this trying to solve?

YouTube already supports background playback PiP.

Steven13799:
This is meant as an optional fallback, not as a replacement for native YouTube PiP.

The specific failure case I am trying to cover is: on some devices, after leaving YouTube, background playback/media session remains active and audio continues, but the expected PiP/floating video window is not created. The user ends up with audio-only playback even though the video is still active.

The local proof of concept was tested against that case:
- regular video opened from the YouTube UI -> HOME -> overlay appears
- regular video opened from an external URL -> HOME -> overlay appears
- real Shorts from the home Shorts shelf -> HOME -> overlay is suppressed to avoid the black/stale surface issue

So the patch is intentionally behind `morphe_floating_video_overlay` and only starts when:
- the setting is enabled
- a regular player is active
- Shorts is not open
- Android overlay permission is already granted

If native PiP works correctly on a device, users do not need to enable this.

LisoUseInAIKyrios:
If background playback PiP doesn't work for some situations, then we should patch whatever is needed so it does work.

I don't think writing and maintainig our own PiP code is needed.

Steven13799:
That makes sense. I agree that a native PiP fix is preferable if we can identify the failing gate.

My local testing started from the symptom rather than from a known native-PiP root cause: playback stayed active after HOME, but the app did not transition to a visible PiP window on the affected device. I tried permission/manifest-side checks first and did not find the blocker there, so this PR implemented the overlay as a fallback.

From the current code, the likely native route is still the existing `backgroundPlaybackPatch` area:
- `BackgroundPlaybackManagerFingerprint` / `isBackgroundPlaybackAllowed(...)`
- the PiP input consumer flag override
- `NewPlayerTypeEnumFeatureFlagFingerprint`
- `NewPlayerOverlaysFeatureFlagFingerprint`
- `PlayerType.WATCH_WHILE_PICTURE_IN_PICTURE` state tracking

I can rework/close this PR if the project direction is to avoid maintaining a separate overlay. The next useful step would be to reproduce the affected-device case while tracing which of those native PiP/background gates differs between "audio keeps playing but no PiP" and "PiP appears". If you know a specific fingerprint/flag that usually controls the PiP transition itself, I can pivot the PR toward that instead of this fallback overlay.

Steven13799:
Reworked the PR in the direction you suggested.

It no longer adds or maintains a custom overlay/PiP implementation. The diff is now limited to the existing background playback patch:
- adds an `onUserLeaveHint()` hook,
- lets YouTube's own PiP path run first,
- retries Android native `enterPictureInPictureMode(...)` at the end only if the activity is not already in PiP, background playback restrictions are removed, Shorts is not open, and the current player is a regular player.

The final diff is 3 files / 41 lines and does not add overlay permissions, a service, or UI controls.

Validation still passes:
- `:patches:compileKotlin`
- `:extensions:youtube:compileDebugJavaWithJavac`

LisoUseInAIKyrios:
It would be much more preferable to figure out the changes needed to remove background playback restrictions.

We would greatly prefer not maintaining our own PiP code.

It's already known that Shorts background PiP is now controlled by some new YT code that is not patched.

Steven13799:
Thanks for the feedback — agreed, maintaining our own PiP code isn't the right approach.

I'll refocus this on the actual goal: forcing the YouTube feature flag(s) that gate background/Shorts PiP, the same way the patch already handles `45415425` (Shorts background), `45638483` (PiP input consumer), `45698813` (new player type) and `45752335` (new player overlays) — rather than calling `enterPictureInPictureMode` ourselves. I'll drop the `enterPictureInPictureFallback` / `onUserLeaveHint` hook.

You mentioned it's already known that Shorts background PiP is now controlled by some new, unpatched YT code. If you can point me to the flag id / class (or even a rough area), I'll put together a proper `…FeatureFlagFingerprint` + `addBackgroundPlaybackFeatureFlagHook` for it.

Marking this as draft while I investigate the gating flag.

Steven13799:
Update — I dug into the decompiled target to locate the gating code.

The native PiP auto-enter is configured in class `izx` (it builds `PictureInPictureParams` and calls `setAutoEnterEnabled`; `MainActivity.onUserLeaveHint()` routes into it via `jah.g(View)`). That class already contains the patched `45638483` (PiP input consumer) flag, so the existing patch is already targeting this area.

Within the same controller there are two experiment flags that are not currently forced by the patch (both read through the usual `afgi.r/s(J…)Z` accessor):

- `45685828` (field `g`) — strongest candidate; it gates PiP code paths (runs an extra Runnable when on, and blocks another path).
- `45408908` (field `K`).

Neither appears in `backgroundplayback/Fingerprints.kt`. My plan is to drop the `enterPictureInPictureFallback` / `onUserLeaveHint` hook and instead add a `…FeatureFlagFingerprint(literal(45685828L))` + `addBackgroundPlaybackFeatureFlagHook(...)`, matching the existing flag-override approach.

Before I commit that: does this line up with the "new YT code" you mentioned, and should the flag be forced on or off? Happy to test on-device either way.

LisoUseInAIKyrios:
The new YT code was the additional Shorts PiP method (identified it the other recently made PR).

Most likely the flags should be forced off, but need to verify.

Try forcing those flags on and verify if PiP doesn't behave right. Flags can be forced on by adding a line of:
`if (flag == 45685828) return true;` to:
https://github.com/MorpheApp/morphe-patches/blob/a8a566774ec8b99c02ce60fdc81ff5258c3b0caf/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/patches/EnableDebuggingPatch.java#L46

LisoUseInAIKyrios:
I have both of those flags forced on (using the Shorts background PiP fix in your other PR), and I don't see any issues.

Maybe it's a combination of those flags and other flags, or it was the existing Shorts background PiP wrong changes that broke it.

Steven13799:
On-device check on the affected device from #1626 (Android 15, with "Remove background playback restrictions" enabled).

I built test APKs that hard-force the candidate flags at their read sites in `izx` (equivalent to the EnableDebuggingPatch override, since each flag is read exactly once):

- `45685828` (field `g`) — forced both `true` and `false`
- `45408908` (field `K`) — forced `true`

Across all of these, for both normal videos and Shorts, pressing HOME never produced a native PiP window: `dumpsys activity` stays at `mLastReportedPictureInPictureMode=false` (no pinned task) and playback goes to `STOPPED`. So on this affected device, forcing these `izx` flags alone does not restore native PiP.

It looks like this device simply never transitions YouTube into system PiP (it only backgrounds/stops audio), so the "force on and watch PiP misbehave" check can't be reproduced here — there's no working PiP to break. The exact flag + direction would need verifying on a device where PiP already works.

For reference: `45685828` appears exactly once, in `izx` (the auto-enter PiP controller that already carries the patched `45638483`). I've kept this PR's draft pointed at `addBackgroundPlaybackFeatureFlagHook(false)` as the best-evidence default per your "most likely off", pending that verification.

LisoUseInAIKyrios:
What happens if those flags are forced off for the problem device? Is background playback fixed?

forcing flags is much preferred since it's simply forcing a specific control flow that YT already has and uses.

Steven13799:
On the problem device, forcing `45685828` **off** did not fix it — native PiP still never appears (`mLastReportedPictureInPictureMode=false`, no pinned task), same as forcing it on. So that flag off alone isn't the fix here.

I haven't yet tried `45408908` off, or **both of them off together** — since you mentioned it might be a combination, I'll force both off on the problem device and report back.

One caveat from my runs: background *audio* behaviour was inconsistent — sometimes it kept playing after HOME, sometimes it went to `STOPPED` — so I want a clean both-off run before drawing conclusions.

Agreed that forcing flags is the right approach over our own PiP code. Given you have both forced on with no issues, your other theory — that the existing Shorts background PiP changes are what regressed it — seems worth checking too; I can bisect that on the problem device if useful.

Steven13799:
Tested both `45685828` and `45408908` forced **off together** on the problem device (with "Remove background playback restrictions" enabled):

- Normal video, confirmed `PLAYING` right before HOME → after HOME: no native PiP (`mLastReportedPictureInPictureMode=false`, no pinned task), and playback goes to `STOPPED` (stable at +3s and +7s).

So forcing both off does **not** restore PiP here, and in this config playback actually stops on HOME. Across everything I've tried (`45685828` on/off, `45408908` on, both off), `PictureInPictureMode` never becomes true on this device.

Caveat: I had toggled the "Remove background restrictions" setting during testing and background-*audio* was sensitive to that, so treat the audio-stop with caution — the reliable signal is PiP mode staying `false`.

Given you have both forced **on** with no issues, and off doesn't help (and stops playback) here, this looks less like these two flags and more like a combination with other flags, or the existing Shorts background PiP changes — as you suggested. I'm happy to bisect the recent Shorts background PiP changes on the problem device if that would help narrow it down.

LisoUseInAIKyrios:
Yes try again with the changes from the other PR. I'll merge that now so you can rebase off dev to include the changes here.

Steven13799:
Sounds good. Once #1629 lands in `dev` I'll rebase this branch on top of it and re-test on the problem device with the corrected Shorts background hook in place (with and without forcing the `izx` flags), then report what actually restores PiP/background there.

Steven13799:
Rebased this PR on current `dev` after #1629 and removed the direct native-PiP retry / `onUserLeaveHint` approach.

The diff is now only the flag override approach you requested:
- `PipAutoEnterFeatureFlagFingerprint` for `45685828L`
- `addBackgroundPlaybackFeatureFlagHook(false)` inside `backgroundPlaybackPatch`

Current diff: 2 Kotlin files in `youtube/misc/backgroundplayback`, no overlay code, no service, no manifest changes, no direct `enterPictureInPictureMode` call.

Local validation:
- `:patches:compileKotlin` passes after updating local `morphe-patches-library` to current `dev`.

Leaving as draft until the flag direction / combined behavior is verified on a working-PiP device.

Steven13799:
Retested on the affected device after rebasing on current `dev` with #1629 included.

Build/test setup:
- Built a local MPP from this branch on top of `dev` (`:patches:build` passes).
- Patched YouTube 20.47.62 with the local MPP and installed it over the existing Morphe YouTube install.
- Confirmed Android PiP app-op is allowed.
- Test flow was a normal feed video, not Shorts: launch app, open a normal video from the home feed, wait until the YouTube media session reports `PLAYING`, then press HOME and inspect activity/window state.

Results:
- With this PR as-is (`45685828` forced off): playback is active before HOME, but after HOME the activity becomes `STOPPED`, `mLastReportedPictureInPictureMode=false`, and no pinned/PiP task appears.
- I also temporarily tested the opposite direction (`45685828` forced on) on the same `dev + #1629` base. Same result: media session `PLAYING` before HOME, then activity `STOPPED`, no PiP.

So on this affected device, toggling `45685828` alone does not restore native auto-PiP/background PiP. #1629 is included and the Shorts hook fix no longer changes this result. This probably means there is another PiP decision gate involved, or a device/form-factor-specific condition before YouTube calls into Android PiP.

Keeping this PR as draft for now because the current 2-file flag override does not actually fix the affected-device behavior.

Steven13799:
Rebased once more on current \\dev\\ after the latest sanitize-link cleanup commit. The PR diff is still only the 2 background playback files for \\45685828\\, and local \\:patches:compileKotlin\\ passes.

LisoUseInAIKyrios:
So the fix is just this flag, correct?

`45685828L`

patches/src/main/kotlin/app/morphe/patches/youtube/misc/backgroundplayback/BackgroundPlaybackPatch.kt
```diff
@@ -133,5 +133,8 @@ val backgroundPlaybackPatch = bytecodePatch(
             // must also be disabled, otherwise the player is a black screen with no buttons and no playback.
             NewPlayerOverlaysFeatureFlagFingerprint.addBackgroundPlaybackFeatureFlagHook(false)
         }
+
+        // New YouTube PiP auto-enter flag in the same controller as the PiP input consumer flag above.
+        PipAutoEnterFeatureFlagFingerprint.addBackgroundPlaybackFeatureFlagHook(false)
```
LisoUseInAIKyrios, line 138: Please cleanup the commit message. It has a nameless non-existent co-author of "Morphe Contributor"

Steven13799:
Yes, this PR is only `45685828L`.

Specifically:
- add a `PipAutoEnterFeatureFlagFingerprint` for `45685828L`
- force it through the existing background playback feature-flag hook (`addBackgroundPlaybackFeatureFlagHook(false)`)

No other PiP code is included here now: no overlay/service/manifest changes and no direct `enterPictureInPictureMode` call.

One caveat from my affected-device retest: toggling only this flag did not restore PiP on that device, so I cannot personally confirm it fixes that device. But the current PR diff itself is just this flag as requested.

LisoUseInAIKyrios:
> One caveat from my affected-device retest: toggling only this flag did not restore PiP on that device, so I cannot personally confirm it fixes that device. But the current PR diff itself is just this flag as requested.

Wait, so what do this change actually do?

If this doesn't fix anything we should not make this change.

Steven13799:
You're right. Based on the affected-device retest, this change does not have evidence of fixing the reported issue.

What it does mechanically is only force YouTube's `45685828L` PiP auto-enter feature flag off, but on the problem device that still left the activity `STOPPED` after HOME with `mLastReportedPictureInPictureMode=false` and no pinned/PiP task. Since it does not restore PiP there, this should not be merged as a fix.

Closing this PR. If we find a different native YouTube gate that actually restores PiP/background behavior on the affected device, I'll open a new focused PR with that evidence.

Steven13799:
Closing because the affected-device retest did not show this flag-only change fixing the issue.

PR #1629, opened 2026-06-01
fix(YouTube - Remove background playback restrictions): Fix Shorts background playback not working

### LINKED ISSUE #289: bug(YouTube): Shorts background playback does not work
### Bug description

Shorts no longer play in the background irrespective of "Disable Shorts background play" setting being off.

### Error logs

```shell

```

### Acknowledgements

Scalee:
The Shorts-background-allowed check was split into two public static boolean(L) methods that both contain feature-flag literal 151635310: the original boolean-returning method and a new one that reads an int field. BackgroundPlaybackManagerShortsFingerprint only filtered on the literal, so it matched both and the patcher resolved to the wrong one. Added an IGET_BOOLEAN opcode filter so the fingerprint uniquely resolves to the correct method. Verified against 21.21.80.

### Description

Closes #289

### Additional context

N/A

### Test results

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -2,7 +2,7 @@
 
 [versions]
 morphe-patcher = "1.5.1"
-morphe-patches-library = "1.4.0-dev.6"
+morphe-patches-library = "1.4.0-dev.8"
 # Tracking https://github.com/google/smali/issues/100.
 smali = "d92701d947"
 agp = "9.1.0"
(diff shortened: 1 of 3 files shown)


0xrxL:
@Scalee You officially fixed the last non-working feature on Morphe xD

👍

LisoUseInAIKyrios:
In later YT versions the literal value was added to another almost identical method in the same class.

So the shorts PiP override was overriding the wrong value.

Could this be the cause of the PiP sometimes using audio only?

PR #1649, opened 2026-06-03
feat(YouTube - Swipe controls): Disable swipe to enter/exit fullscreen

0xrxL:
This patch is incomplete at the moment, because I don't know how to implement it in the new gesture controls to change volume/brightness (I guess that once will be completed, this patch will disable the exit from fullscreen gesture only when a tap is performed on the part of the display reserved for gestures to change volume/brightness).

### DIFF
--- a/patches/src/main/kotlin/app/morphe/patches/reddit/shared/Constants.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/reddit/shared/Constants.kt
@@ -15,17 +15,17 @@ internal object Constants {
         ),
         targets = listOf(
             AppTarget(
+                version = "2026.22.0",
+                minSdk = 28,
+                isExperimental = true,
+            ),
+            AppTarget(
                 version = "2026.21.0",
                 minSdk = 28,
                 isExperimental = true,
             ),
             AppTarget(
                 version = "2026.20.0",
-                minSdk = 28,
-                isExperimental = true,
-            ),
-            AppTarget(
-                version = "2026.19.0",
                 minSdk = 28,
                 isExperimental = true,
             ),


MarcaDian:
I don't know if we need to hide the Native section in the swipe graph if these gestures are disabled.

0xrxL:
> I don't know if we need to hide the Native section in the swipe graph if these gestures are disabled.

I think it's not needed. Should be implicit that the fullscreen gesture is disabled where brightness/volume gestures works.

AmazingKo:
@0xrxL Is that the feature that allows disabling the swipe-down gesture below the player?
<p>

Picsart_26-05-05_15-09-39-305

</p>
</details> 

I've really been waiting for this specific feature just so I can comfortably make the switch to YT Morphe

0xrxL:
> @0xrxL Is that the feature that allows disabling the swipe-down gesture below the player? Picsart_26-05-05_15-09-39-305 I've really been waiting for this specific feature just so I can comfortably make the switch to YT Morphe

That's another thing, always based on what you're looking for, but not the same. This is the patch to disable the entire code to drag the player view (included the swipe to minimize it), but limited to the fullscreen, to ensure brightness/volume gestures works flawlessly.

I'll look to the selective gesture to blocking fullscreen gesture asap anyway.

0xrxL:
> @0xrxL Is that the feature that allows disabling the swipe-down gesture below the player? Picsart_26-05-05_15-09-39-305 I've really been waiting for this specific feature just so I can comfortably make the switch to YT Morphe

Correction. After discussing about this PR, I decided that it will become the patch you requested.

AmazingKo:
> Correction. After discussing about this PR, I decided that it will become the patch you requested.

Oh man, I'm so grateful to you for considering such a major feature. Much respect!

AmazingKo:
@0xrxL Wait is it cancelled?

0xrxL:
> @0xrxL Wait is it cancelled?

Nope. I only reset the PR because there was some issues with dev branch. I'll push new stuff once I'll finish to port it to Morphe (it's currently done from the bytecode side, but I need to write it as patch).

PR #1653, opened 2026-06-04
chore(Settings): Merge some restart dialog messages into summary

kitadai31:
### Description

Thanks to #1366, we now have room in the summary to include a more detailed explanation of that feature.

In this PR, I moved explanations that didn't warrant a warning dialog to the user, such as a simple list of limitations, from the restart dialog to the summary.

I've kept the restart dialog for settings that should "warn" the user.

Here are some of the changes that need to be explained:

- `Settings.EXPAND_LIVESTREAM_DVR_DURATION` has been moved to a different line because it was in the wrong block.
- `<!-- interaction.seekbar.expandLivestreamDvrDurationPatch -->` section has been moved to `livestreamDvrPatch` because it had been merged into one patch.
- `morphe_bypass_ambient_mode_restrictions_user_dialog_message` has been removed. This restart dialog was necessary because the previous summary didn't clearly explained that this setting affected battery saver, but I think it is unnecessary now in the new summary because it's much clearer.
- `morphe_miniplayer_disable_resuming_summary`: The message "This setting cannot force on this Premium feature" has been removed, but it's not necessary now. Previously, its summary_off text was "'Continue watching' can show on app startup," which led users to expect the premium features to be unlocked, so the dialog was necessary.

### Additional context

Some strings may need renaming on Crowdin before merging.
There are several strings where only the punctuation has been changed.

### Test results


LisoUseInAIKyrios:
Anything left Todo before merging?

kitadai31:
> Anything left Todo before merging?

No, it's ready.

Please remember to rename on Crowdin.

LisoUseInAIKyrios:
Oops, nobody renamed on Crowdin before merging.

Any translations made by users should be kept from the "translation memory", but any automated translations will be recreated.

Issue #24, opened 2025-12-31
feat(YouTube): Add a function to determine if the video is playable on YouTube Music


inotia00:
### Feature description

Add a function to determine if this video is playable on YouTube Music.

This function can be implemented with minimal API calls, doesn't require login, and doesn't place a burden on the server since it only checks the playabilityStatus.

With this implementation, the following two functions can be implemented:

1. If the video is playable on YouTube Music, change the playback speed to 1.0.
: This is a useful feature for me, as I listen to a lot of music on YouTube.

2. If the video is playable on YouTube Music, show a `View lyrics` button in the player.
: Fetch lyrics using [LRCLIB](https://lrclib.net/docs) (open source), a crowdsourced lyrics database.

### Motivation

The strengthened YouTube Music paywall policy has led to more users listening to music on YouTube instead of YouTube Music.

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] This is a feature request for YouTube or YT Music

Soler3737:
I'm not used to github so apologies for my question, is this being worked on, if so, is there an expectation on release of this patch?

smi123ler:
pls implement this feature..... we are waiting for this .. i was waiting for next update and got the update but feature was not there ..

LisoUseInAIKyrios:
It'll be released when someone (anyone) decides to spend their personal time and add the feature.

We invite anyone with modest dev skills to help speed this up.

smi123ler:
besides hiding the comments and removing peoples request saying its duplicate, pls implement that feature,, simple

PR #1661, opened 2026-06-04
feat(YouTube - SponsorBlock): Add channel whitelist

MarcaDian:
### Description

Adds a channel whitelist to SponsorBlock. Channels added to the whitelist are fully skipped - no segments are fetched or applied.

 - Long-press the `Create segment button` during playback to open the whitelist dialog
 - The same dialog is accessible from `Settings → SponsorBlock → General → Channel whitelist`
 - Add or remove channels directly from the dialog; the list updates without closing
 - Optional toast notification when opening a video on a whitelisted channel (disabled by default)

### Test results

___


fromhgbwithluv:
Imo, we should have Whitelist for playback speed too :D

MarcaDian:
I was thinking of adding it after this feature, but @LisoUseInAIKyrios suggests implementing what @inotia00 suggested in https://github.com/MorpheApp/morphe-patches/issues/24

PR #1685, opened 2026-06-07
feat(YouTube): Add `Voice over translation` patch

### LINKED ISSUE #844: feat(YouTube): Add 'Voice over translation'
### Feature description

This translation feature will be helpful to understand the video topics in native language or will help to understand the  details..

Its implemented in Anddea Rvx patches...
 Consider adding this.

Image

### Motivation

Will help to Break the language barrier and to help understanding the topic as like native language

### Acknowledgements

MarcaDian:
### Description

Adds a voiceover translation feature.

Feature is enabled in:
`YouTube settings > Morphe > Video > Voice over translation`

Voiceover is handled by translating the video closed captions using a translation service:
- Google Translate (Free)
- MyMemory (Free)
- OpenRouter (Almost free)

Then voiceover audio is generated using Edge TTS (OpenAI) or the native Android TTS system.

Google Translate and MyMemory are free but can be rate limited. If you reach a rate limit it is recommended to use OpenRouter provider because it has no rate limits and is nearly free in practice.

The cheapest built-in OpenRouter model designed for translations (Mistral Nemo) cost about $0.02 per 100 hours of video. A 1 hour video that is watched halfway still uses 1 hour of OpenRouter costs because the entire video is translated in the background.

The "Max speech rate" slider (1.0x - 2.5x, default 1.5x) caps how fast the voiceover can be sped up to fit a longer translation into its caption time slot.

- Lower values keep speech closer to natural pace. If a translation does not fit even at the cap, the segment may be skipped rather than rushed.
- Higher values let the voiceover speed up more aggressively to fit longer translations, trading some clarity for completeness.

Because some languages are wordier than others (translations into Ukrainian, German, etc. tend to be longer than the English source), lowering the video playback speed to 0.8x or 0.9x gives the voiceover more room to fit at a natural pace - TTS automatically follows the video playback speed.

### Test results

___


MarcaDian:
The best way to implement this is to get native translated subtitles, but I couldn't do it so I give up.

For now it works by translating the original English subtitles in the runtime. If anyone knows a way to get them that would be the best solution.

MarcaDian:
The subtitle parser needs to be configured, currently it breaks it into phrases as they are available on the endpoint, that is, if a sentence is inserted into both phrases, TTS does not understand this and makes pauses where they are not needed.

MarcaDian:
I added some heuristics to very old subtitles that are just a stream of text, and it seems to have worked for me.

imPratyoosh:
Is this has been adopted from anddea patches or made from scratch?

LisoUseInAIKyrios:
We wrote everything from scratch.

Jav1x, who is the author of the RVX patch, uses a Yandex browser spoof to play a Yandex provided audio track synced to the video.

We are using Google translate to translate, and native Android and Microsoft Edge for TTS audio voiceover.

There is no Yandex here, no Jav1x code, and no Anddea code. This is entirely Marca and my code, just as we did with all prior features we added to ReVanced and Morphe.

l205306:
Close https://github.com/MorpheApp/morphe-patches/issues/844

0xrxL:
Still needs a way to keep the button visible but always set on the last user's preference (also after the next app opening).

LisoUseInAIKyrios:
Yes will fix soon.

MarcaDian:
I think the initialization and display of the voiceover button needs a refactoring.

MarcaDian:
If we don't have any ideas or fixes, we can merge it. I bet users will find a million more problems.

LisoUseInAIKyrios:
Will merge after the next stable release (tomorrow?)

0xrxL:
@MarcaDian Set the initial value of overlay button to disabled, in a way to avoid ppl a scare jump when they'll hear two voices. 😂

MarcaDian:
I hope users aren't so stupid as to not understand what they're doing when they enable this feature in the settings.
Especially when they even show a dialog when turned on.

0xrxL:
> I hope users aren't so stupid

You already know the answer in your heart. 😂

0xrxL:
@LisoUseInAIKyrios I think worth to merge this in the imminent stable, in order to receive more possible feedbacks from users and prepare the next big PR to fix all them.

LisoUseInAIKyrios:
I think it needs to be in pre-release first in case there's a large bug or glitch that went unnoticed. Any immediate fixes or improvements can be on pre-release so the stable release is more polished

0xrxL:
Last time I was able to reproduce a delayed VoT starts, that caused a faster talking until it reached the current playback time. Maybe it would be good to syncronize VoT with the current video time, instead to let VoT starting from the beginning.

MarcaDian:
It's easy to say, but not to implement.

LisoUseInAIKyrios:
We previously tried adding a setting of "Wait for TTS" where it blocks playback until the first few TTS samples are received. It completely fixed fast speaking TTS caused by the delay of running TTS pipeline, but the hook didn't work for all situations.

Need to find another suitable hook that is off the main thread and allows blocking playback (hook must also have the video id, or video id is already available from VideoInformation).

0xrxL:
Go go go...we're near to 300. 😂

MarcaDian:
We are ready to merge it

Metadorius:
> There is no Yandex here

I wonder, how's the quality in comparison? Of the voiceover translation I mean.

MarcaDian:
The quality is as good as it gets translated by translation providers. All we can do is use free APIs. Yes, there are paid ones that produce almost human-quality voice acting, but this is not our variant.

DarkHorse313:
Can you make caption translation option in my preferred language

LisoUseInAIKyrios:
In the voiceover dialog menu and the Morphe settings menu, you can set the language to translate into. By default it's the app language.

KerimDemirkaynak:
It would be great to have additional TTS providers, such as Azure and Gemini.

Also, the synchronization seems to have some issues. For example, if I start playing a video from the middle, the TTS begins reading from the start of the transcript instead of the current playback position. It should stay synchronized with the video's current timestamp.

For reference, you can take a look at the synchronization in the YouTube Dubbing – Translate & Voice Over extension. It handles playback synchronization very well:
https://chromewebstore.google.com/detail/youtube-dubbing-%E2%80%93-transla/oglffgiaiekgeicdgkdlnlkhliajdlja

MarcaDian:
If you find any free TTS, we would be happy to add them. Do not compare web extensions and app mods; they are different things.

KerimDemirkaynak:
Can the Gemini API or Nvidia endpoint API be added for translation?

LisoUseInAIKyrios:
OpenRouter offers numerous Gemini models
https://openrouter.ai/models?q=gemini&order=most-popular

I'm not sure what nvidia end point you're referring to, but OpenRouter has Nvidia models as well including some rate limited 100% free models
https://openrouter.ai/models?order=most-popular&q=nvidia

PR #1687, opened 2026-06-07
refactor(YouTube - Open system share sheet): Move share sheet logic to `OpenSystemShareSheetPatch`

ILoveOpenSourceApplications:
### Description

N/A

### Additional context

N/A

### Test results

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/OpenSystemShareSheetPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/OpenSystemShareSheetPatch.java
@@ -7,10 +7,21 @@
 
 package app.morphe.extension.youtube.patches;
 
+import static app.morphe.extension.shared.Utils.getContext;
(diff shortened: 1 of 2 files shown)


extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/OpenSystemShareSheetPatch.java
```diff
@@ -27,4 +38,68 @@ public final class OpenSystemShareSheetPatch {
     public static void onFlyoutMenuCreate(final RecyclerView recyclerView) {
         flyoutMenuRecyclerView = new WeakReference<>(recyclerView);
     }
-}
+
+    /**
+     * Injection point.
+     */
+    public static boolean openSystemShareSheet(String asciiBuffer) {
```
LisoUseInAIKyrios: If this is called from patch code then need to update that. If this isn't called from patch code then injection comment should be removed.

Issue #1666, opened 2026-06-06
bug(YouTube): "Save to watch later/Queue manager" patch sometimes targets personal channel instead of brand channel


98xe:
### Bug description

This is sort of a continuation of https://github.com/MorpheApp/morphe-patches/issues/1432.

When on a brand channel, using the patch would sometimes target/use the playlists of your personal channel, rather than those of the active brand channel. The bug persists in the `Save to watch later` patch, and now the "Queue manager" part of it is experiencing similar behavior.

I don't know what exactly causes it to happen, but here's how I've been able to both trigger the bug and get it working (though not really practical).

---

To reproduce bug using "Save to watch later"

1. Open YouTube
1. Switch to a brand channel
2. Force close and reopen the YouTube app
3. Load a video and hit the `Save to watch later` player button
4. Check the "Watch Later" playlist of the brand channel, the video is not listed there
5. Switch over to the personal channel
6. The video appears in the personal channel's "Watch Later" playlist instead

When it works properly

1. Open YouTube
1. Switch to your personal channel
2. Switch back to the brand channel
3. Load a video and hit the `Save to watch later` player button
4. Check the "Watch Later" playlist of the brand channel, the video is there

---

To reproduce bug using "Queue manager"

1. Open YouTube
1. Switch to a brand channel
2. Force close and reopen the YouTube app
3. Load a video and trigger the `Queue manager` menu to open
4. Select `Add to queue`
5. Reopen the menu and select `Save queue to playlist`
6. The playlists that appear belong to the personal channel, not the brand channel

When it works properly

1. Open YouTube
2. Switch to your personal channel
3. Switch back to the brand channel
4. Load a video and trigger the `Queue manager `menu to open
5. Select `Add to queue`
6. Open the menu and select `Save queue to playlist`
7. The playlists that appear belong to the brand channel

### App version patched

20.51.39

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

0xrxL:
@98xe mhh...try this

...

98xe:
@0xrxL 

Couldn't get it to patch.

<details>
<summary>Log</summary>

```
app.morphe.patcher.patch.PatchException: The patch "Save to watch later" depends on "BytecodePatch@217486701", which raised an exception:
[...]

Issue #876, opened 2026-03-16
feat(YouTube): Whitelist functionality to exclude channels from the default playback speed


tk-333:
### Feature description

Other applications (such as anddea) offer features like adding playback speed to a whitelist via an action button (to exclude specific channels from the default playback speed) or excluding videos in the music category from the default playback speed.

### Motivation

I recently discovered Morphe and have found it very useful, thank you. I frequently watch talk-based YouTube channels, so I have set the default playback speed to be faster. However, I often revert the playback speed when watching music channels or channels where the speaker's pace is already very fast.

I apologize for this minor request, but I would appreciate it if you could consider adding the aforementioned features if possible. I would like to make Morphe my primary application.

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] The feature request is only related to Morphe (Manager).

KenzKD:
I think instead of whitelisting the channels we should automatically save the last speed of the channel  if it is different from the default and if we open any video from that channel, it can just refer to speed.

For an example, My Default speed is set to 2x, some channels I know has a fast talker as host or it might be a music channel in which case I slow down the video to 1.5x or 1x. If this speed gets saved to some list, it can then be later referred to when playing a video from those channels.

I had recently made a custom script that works with the YT Enhancer plugin to get this to work on my Browser and it has been amazing to use so far. Here is my browser implementation that can be used as reference: https://github.com/KenzKD/YouTube_Channel_Speed_Controller

dvvmvp:
[Anddea RVX YouTube](https://github.com/anddea/revanced-patches) already implemented this.

<img width="1080" height="321" alt="Image" src="https://github.com/user-attachments/assets/7b29b23e-8411-4420-85d1-41c56921ba66" />

It would be amazing if Morphe could port the patch. 🙏

LisoUseInAIKyrios:
@dvvmvp we have plans to whitelist music speed using a completely different strategy. See https://www.github.com/MorpheApp/morphe-patches/issues/24

dvvmvp:
@tk-333 , this is live starting from release v1.31.0. Feel free to close this issue.

KenzKD:
I think this should be reopened as the feature added in 1.31.0 is only for music videos. Where as this requested feature is to save the speed of any channel

MarcaDian:
[...]

PR #1691, opened 2026-06-08
feat(YouTube - Playback speed): Add "Disable playback speed for music" setting

### LINKED ISSUE #24: feat(YouTube): Add a function to determine if the video is playable on YouTube Music

inotia00:
### Description

Closes #24.

### Additional context

N/A

### Test results

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/playback/speed/RememberPlaybackSpeedPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/playback/speed/RememberPlaybackSpeedPatch.java
@@ -82,7 +82,13 @@ public final class RememberPlaybackSpeedPatch {
         if (newVideoStarted) {
             newVideoStarted = false;
 
+            final float defaultSpeed = Settings.PLAYBACK_SPEED_DEFAULT.get();
+
             if (DISABLE_PLAYBACK_SPEED_MUSIC) {
+                if (defaultSpeed == 1.0f) {
+                    return 1.0f;
+                }
+
                 String videoId = VideoInformation.getVideoId();
                 GetMixPlaylistRequest request = GetMixPlaylistRequest.getRequestForVideoId(videoId);
                 boolean isMusic = request != null && Boolean.TRUE.equals(request.getResult());
@@ -91,7 +97,6 @@ public final class RememberPlaybackSpeedPatch {
                 }
             }
 
-            final float defaultSpeed = Settings.PLAYBACK_SPEED_DEFAULT.get();
             if (defaultSpeed > 0) {
                 return defaultSpeed;
             }
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/playback/speed/RememberPlaybackSpeedPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/playback/speed/RememberPlaybackSpeedPatch.java
@@ -2,15 +2,22 @@ package app.morphe.extension.youtube.patches.playback.speed;
 
 import static app.morphe.extension.shared.StringRef.str;
 
+import java.util.Collections;
+
 import app.morphe.extension.shared.Logger;
 import app.morphe.extension.shared.Utils;
 import app.morphe.extension.youtube.patches.VideoInformation;
+import app.morphe.extension.youtube.patches.utils.requests.GetMixPlaylistRequest;
 import app.morphe.extension.youtube.settings.Settings;
 
 @SuppressWarnings("unused")
 public final class RememberPlaybackSpeedPatch {
 
+    private static final boolean DISABLE_PLAYBACK_SPEED_MUSIC = Settings.DISABLE_PLAYBACK_SPEED_MUSIC.get();
+
     private static final long TOAST_DELAY_MILLISECONDS = 750;
+
+    private static volatile String lastFetchedVideoId = "";
 
     private static volatile boolean newVideoStarted;
 
@@ -75,6 +82,15 @@ public final class RememberPlaybackSpeedPatch {
         if (newVideoStarted) {
             newVideoStarted = false;
 
+            if (DISABLE_PLAYBACK_SPEED_MUSIC) {
+                String videoId = VideoInformation.getVideoId();
+                GetMixPlaylistRequest request = GetMixPlaylistRequest.getRequestForVideoId(videoId);
+                boolean isMusic = request != null && Boolean.TRUE.equals(request.getResult());
+                if (isMusic) {
+                    return 1.0f;
+                }
+            }
+
             final float defaultSpeed = Settings.PLAYBACK_SPEED_DEFAULT.get();
             if (defaultSpeed > 0) {
                 return defaultSpeed;
@@ -84,4 +100,11 @@ public final class RememberPlaybackSpeedPatch {
         return -2.0f;
     }
 
+    public static void newVideoStarted(String videoId, boolean isShortAndOpeningOrPlaying) {
+        if (DISABLE_PLAYBACK_SPEED_MUSIC && !VideoInformation.lastPlayerResponseIsShort() &&
+                !lastFetchedVideoId.equals(videoId)) {
+            lastFetchedVideoId = videoId;
+            GetMixPlaylistRequest.fetchRequestIfNeeded(videoId, Collections.emptyMap());
+        }
+    }
 }
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/utils/requests/GetMixPlaylistRequest.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/utils/requests/GetMixPlaylistRequest.java
@@ -1,0 +1,142 @@
+/*
+ * Copyright 2026 Morphe.
+ * https://github.com/MorpheApp/morphe-patches
+ *
+ * See the included NOTICE file for GPLv3 §7(b) and §7(c) terms that apply to this code.
+ */
+
+package app.morphe.extension.youtube.patches.utils.requests;
+
+import androidx.annotation.Nullable;
+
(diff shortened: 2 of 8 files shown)


patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -1348,6 +1348,8 @@ Tap here to learn more about DeArrow"</string>
     <string name="morphe_remember_playback_speed_last_selected_toast_summary">Shows a notification when a new default playback speed is saved</string>
     <string name="morphe_playback_speed_default_title">Default playback speed</string>
     <string name="morphe_remember_playback_speed_toast">Changed default playback speed to: %s</string>
+    <string name="morphe_disable_playback_speed_music_title">Disable playback speed for music</string>
+    <string name="morphe_disable_playback_speed_music_summary">Change the default playback speed to 1.0 when playing music</string>
```
inotia00: Is the summary of this preference appropriate?

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/playback/speed/RememberPlaybackSpeedPatch.java
```diff
@@ -84,4 +100,11 @@ public static float getPlaybackSpeedOverride() {
         return -2.0f;
     }
 
+    public static void newVideoStarted(String videoId, boolean isShortAndOpeningOrPlaying) {
+        if (DISABLE_PLAYBACK_SPEED_MUSIC && !VideoInformation.lastPlayerResponseIsShort() &&
+                !lastFetchedVideoId.equals(videoId)) {
+            lastFetchedVideoId = videoId;
+            GetMixPlaylistRequest.fetchRequestIfNeeded(videoId, Collections.emptyMap());
```
inotia00, line 112: Since this endpoint request is not something like playing a video or creating a new playlist, sign-in is not required. That is why an empty map was used  (This is not a major issue as alternative clients like NewPipe always work with sign-out).

If necessary, add `authHookPatch` as a dependency and replace `Collections.emptyMap()` with `AuthUtils.getRequestHeader()`.

patches/src/main/kotlin/app/morphe/patches/youtube/misc/auth/Fingerprints.kt
```diff
@@ -8,25 +8,27 @@
...
-    accessFlags = listOf(AccessFlags.PUBLIC, AccessFlags.CONSTRUCTOR),
+internal const val GET_PAGE_ID_STRING = ", getPageId="
+internal const val IS_INCOGNITO_STRING = ", isIncognito="
+
+internal object AccountIdentityToStringFingerprint : Fingerprint(
+    name = "toString",
+    accessFlags = listOf(AccessFlags.PUBLIC, AccessFlags.FINAL),
+    returnType = "Ljava/lang/String;",
+    parameters = listOf(),
     strings = listOf(
```
inotia00, line 29: Fingerprint filters were not used because the order in which strings are positioned in the `toString()` method could change at any time.

patches/src/main/kotlin/app/morphe/patches/youtube/misc/auth/AuthHookPatch.kt
```diff
@@ -7,118 +7,55 @@
...
-        ).matchAll().forEach { fingerprint ->
-            val method = fingerprint.method
-            val instructionIndex = fingerprint.instructionMatches.first().index
+        AccountIdentityConstructorFingerprint.method.apply {
+            val pageIdIndex = indexOfFirstInstructionOrThrow {
+                opcode == Opcode.IPUT_OBJECT && getReference<FieldReference>() == pageIdField
+            }
+            val incognitoIndex = indexOfFirstInstructionOrThrow {
+                opcode == Opcode.IPUT_BOOLEAN && getReference<FieldReference>() == incognitoField
+            }
```
inotia00, line 56: Fingerprint filters were not used because the order in which fields are positioned in the `<init>` method could change at any time.

0xrxL:
@inotia00 Please, restore the fingerprint and the patch that I pushed to get pageId value from the alternative method, because it fix this issue

https://github.com/MorpheApp/morphe-patches/issues/1666

inotia00:
Since fields are initialized only in the `<init>` method (this class inherits from Parcelable and Serializable), these appear to be unnecessary hookings.

I tested this PR on YouTube on 20.21.37, 20.51.39, and 21.23.480, but the #1666 was not reproduced on my side.

patches/src/main/kotlin/app/morphe/patches/youtube/misc/auth/Fingerprints.kt
0xrxL: This fingerprint is 100% used, and it map a method inside the class that return the pageId value from one of both available methods in 21.xx:

The first return a boolean, and hooking it allow to get the empty pageId.

The second return a string, and hooking it allow to get a filled pageId value when you switch to the brand account.

0xrxL:
> Since fields are initialized only in the `<init>` method (this class inherits from Parcelable and Serializable), these appear to be unnecessary hookings.
> 
> I tested this PR on YouTube on 20.21.37, 20.51.39, and 21.23.480, but the #1666 was not reproduced on my side.

The problem is not when the variable is initialized, but the fact that the init is called at the wrong times and with wrong values.

Try to tap on change account button, without changing the account, and you will notice that if you are in a brand account at that time, the current value of the pageId will be set empty and will no longer be set with the right value again.

That at least happens to me (and to the user that reported the issue).

For this reason I changed the hook, because the two alternative methods are called in the right moments (and in any case with the right value).

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/utils/requests/GetMixPlaylistRequest.java
```diff
@@ -0,0 +1,142 @@
...
+            }
+            JSONObject watchEndpoint = navigationEndpoint.getJSONObject("coWatchWatchEndpointWrapperCommand")
+                    .getJSONObject("watchEndpoint")
+                    .getJSONObject("watchEndpoint");
+
+            if (!watchEndpoint.has("playerParams")) {
+                return false;
+            }
+
+            return watchEndpoint.getString("playerParams").startsWith("8AUB");
```
fromhgbwithluv, line 127: Is it possible to use `Hook.ProtoBufferParameterBeforeVideoId` to check instead of sending a request, similar to how we check if a video is a short ?

inotia00: If the current video is opened in a mix playlist, an endpoint request is not required; otherwise, an endpoint request is necessary.

patches/src/main/kotlin/app/morphe/patches/youtube/misc/auth/Fingerprints.kt
inotia00: I reverted to the state prior to refactoring as much as possible, and as a result, there are now only minor changes to improve code visibility.

patches/src/main/kotlin/app/morphe/patches/youtube/misc/auth/Fingerprints.kt
```diff
@@ -8,25 +8,85 @@
...
-        opcode(Opcode.IF_EQZ, location = InstructionLocation.MatchAfterAnywhere()),
-        opcode(Opcode.IPUT_OBJECT, location = InstructionLocation.MatchAfterImmediately()),
-        opcode(Opcode.IPUT_BOOLEAN, location = InstructionLocation.MatchAfterImmediately()),
-    ),
+        GET_PAGE_ID_STRING,
+        IS_INCOGNITO_STRING
+    )
 )
+
+internal fun getIncognitoStatusFingerprint(incognitoField: FieldReference) = object : Fingerprint(
```
inotia00, line 35: If the fingerprint is declared as a variable, the fingerprint name is not shown if the fingerprint resolve fails. This is why it is declared as a function (object).

0xrxL:
@inotia00 Anyway...if you needs of any information about my PRs, feel free to contact me as the usual address (I also sent you some mails).

patches/src/main/resources/addresources/values/youtube/strings.xml
LisoUseInAIKyrios: Maybe for the title something like `Use normal playback speed for music`?

LisoUseInAIKyrios (approved):

LisoUseInAIKyrios (approved):
GitHub wants another review for whatever glitchy reason.

MarcaDian:
Are podcasts also positioned as music videos?

LisoUseInAIKyrios:
Doesn't seem to. Only videos labeled as 'music' work.

MarcaDian:
So is it still necessary to add a whitelist to exclude channels from the default playback speed? because of

> "watch talk-based YouTube channels" or "channels where the speaker's pace is already very fast"

https://github.com/MorpheApp/morphe-patches/issues/876

LisoUseInAIKyrios:
We could if enough people want it, and we want to do it.

Issue #1085, opened 2026-04-02
bug(YouTube): Dislike and likes are not shown, cannot hide action buttons, restore old player layout


lucai9000:
&nbsp;
## SOLUTION

Re-patch using the latest patches. You can confirm you have the latest patches by checking: `YouTube Settings > Morphe patches > About`

Then enable: `YouTube settings > Morphe > Player > Restore old video action bar`

&nbsp;

### Bug description

Suddenly today the like and dislike numbers are gone. Am I the only one? A/b test or incompatibility with the latest experimental only?

### App version patched

21.13.163

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

LisoUseInAIKyrios:
It's a new a/b layout.

Try spoof app version and see if it's fixed.

ILoveOpenSourceApplications:
I think this is just one of the bugs introduced by  YouTube and not an A/B layout. YouTube being themselves.

lightningHoneypot:
> I think this is just one of the bugs introduced by  YouTube and not an A/B layout. YouTube being themselves. 

Oh wtf? Now we can't see which videos are a scam in mobile, ffs

LisoUseInAIKyrios:
I don't think it's a bug. It seems they are experimenting with hiding the likes count entirely.

adityamall3181:
> It's a new a/b layout.
> 
> Try spoof app version and see if it's fixed.

but if i switch to spoof app version then my version is,changed to 20.13 which is i not want if any other solution you have then please tell me

LisoUseInAIKyrios:
Your only other solutions are:

- Disable layout updates (has side effects)
- Patch an old version of YouTube like 20.21 or 20.31
- Try a different YouTube account (maybe works maybe not)
- Clear YouTube app data and maybe they won't give new layout

Quicksilver151:
Just happened to me as well

Disable layout updates works for now

DeHaos:
> It's a new a/b layout.
> 
> Try spoof app version and see if it's fixed.

It fixed it for me. Disabling layout update didn't work.

inotia00:
Using the `Enable tablet layout in player` setting added in #1695 can be a temporary workaround.

Since this layout has now been widely rolled out, we can parse the API response to write a patch for the fix.

error-reporting:
[...]

PR #1695, opened 2026-06-08
feat(YouTube - Change form factor): Add "Enable tablet layout in player" setting

inotia00:
### Description

You can change to **Tablet layout** via the **Device layout** setting, but this applies everywhere, including `search results`, `YouTube settings`, `playlists`, `channels`, and `feeds`, not just the `player`.

Some people use the tablet layout simply for the **non-collapsed video action bar**, and they may not want the tablet layout to be used in `search results`, `YouTube settings`, `playlists`, `channels`, `feeds`, etc.

This PR is for these people.

The **Enable tablet layout in player** setting enables the tablet layout `only in the player`.

### Additional context

N/A

### Test results

### DIFF
--- a/patches/src/main/resources/addresources/values/youtube/strings.xml
+++ b/patches/src/main/resources/addresources/values/youtube/strings.xml
@@ -1094,6 +1094,8 @@ Automotive layout
 • Remix button and Sound button are not available in Shorts
 • Shorts open in the regular player
 • Video action bar is not collapsed"</string>
+    <string name="morphe_tablet_layout_in_player_title">Enable tablet layout in player</string>
+    <string name="morphe_tablet_layout_in_player_summary">This might be useful for those who want a non-collapsed video action bar</string>
 
     <!-- layout.spoofappversion.spoofAppVersionPatch -->
     <string name="morphe_spoof_app_version_title">Spoof app version</string>
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/ChangeFormFactorPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/ChangeFormFactorPatch.java
@@ -10,12 +10,15 @@
 
 package app.morphe.extension.youtube.patches;
 
+import static app.morphe.extension.youtube.patches.ChangeFormFactorPatch.FormFactor.LARGE;
 import static app.morphe.extension.youtube.shared.NavigationBar.NavigationButton;
 
 import androidx.annotation.Nullable;
 
+import java.util.List;
 import java.util.Optional;
 
+import app.morphe.extension.shared.settings.Setting;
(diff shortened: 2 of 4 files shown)


patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -1094,6 +1094,8 @@ Automotive layout
 • Remix button and Sound button are not available in Shorts
 • Shorts open in the regular player
 • Video action bar is not collapsed"</string>
+    <string name="morphe_tablet_layout_in_player_title">Enable tablet layout in player</string>
+    <string name="morphe_tablet_layout_in_player_summary">This might be useful for those who want a non-collapsed video action bar</string>
```
inotia00, line 1098: Since this code was used in a personal build, the preference summary may not be appropriate.

LisoUseInAIKyrios: Maybe this should be a user dialog message? Is the action bar area the only noticeable change?

kitadai31: Most users who want the non-collapsed video action bar probably simply want to display the channel name.
I think it would be better to mention about "channel name".
Also, the term "non-collapsed video action bar" is unclear if user is seeing it for the first time.

Therefore, I suggest the following ideas.

- This might be useful for those who want a non-collapsed video action bar (includes full channel name)
- This might be useful for those who want to show channel name above the video action bar
- This might be useful for those who want to show channel name below the video title
- This might be useful for those who want to restore old non-collapsed video action bar layout (includes full channel name)
- This might be useful for those who want to restore channel name above video action bar
- This might be useful for those who want to restore channel name below video title

Of these, I prefer the variant with "restore" in it. For phone users, the "non-collapsed video action bar" is an "old" layout, so I think the phrase "restore" is more appropriate.

By the way, in a recent pull request, I changed the description for version spoofing 20.13.41 to "Restore channel name below video player".
https://github.com/MorpheApp/morphe-patches/blob/a8a566774ec8b99c02ce60fdc81ff5258c3b0caf/patches/src/main/resources/addresources/values/youtube/strings.xml#L1064
You can update this summary to match this PR as needed to keep consistency

---

> Maybe this should be a user dialog message?

User dialog messages cannot be viewed until tapped.
Furthermore, it doesn't show up in search results. Even if a user searches for "non-collapsed," that setting will not be found.
IMO, it should only be used to explain negative effects or very long explanations that cannot be contained in a summary.
(In my recent PR, I re-evaluated the user dialog messages based on this criterion.)

LisoUseInAIKyrios: I think the "this might be useful" could be omitted, and instead just mention the primary thing it will do (Restore old player layout?)

inotia00:
This PR can be used as a temporary workaround for #1085.

PR #1699, opened 2026-06-08
feat(YouTube - Hide ads): Hide 'Includes paid promotion' label in miniplayer

ILoveOpenSourceApplications:
### Description

Screenshot_20260608_171554_YouTube
Screenshot_20260608_222527_YouTube

### Additional context

N/A

### Test results


0xrxL:
Finally. ✌️

LisoUseInAIKyrios:
This should be already hidden with "hide subtexts" is enabled.

But current subtext can be many other things besides paid promotional labels.

ILoveOpenSourceApplications:
What other things do show up in miniplayer? Certainly not subtitles/closed captions.

LisoUseInAIKyrios:
It's been a while since I added the hide function, but I remember it showing a "resume playback" subtitle for YT Premium users when reopening the app, and the channel name could briefly show when playback started.

ILoveOpenSourceApplications:
Oh, I haven't used premium in a while and I guess the "Continue watching" part shows up there too right?

ILoveOpenSourceApplications:
Selecting other text views doesn't show anything. So how do we distinguish, the paid promotion label from others?

Screenshot_20260609_130840_YouTube

LisoUseInAIKyrios:
Possibly we don't distinguish? And we leave this as part of the hide subtitle?

ILoveOpenSourceApplications:
Does "modern_miniplayer_continue_watching" do anything?

LisoUseInAIKyrios:
I think that's the Premium "continue watching" view.

Previously I remember that was hidden with the "hide subtexts" setting, but maybe the layout has changed for it. I think hiding that is less important now because now there is a miniplayer setting to turn off the premium resume function, and the resume text is hidden if the user chooses to resume playback.

ILoveOpenSourceApplications:
If continue watching uses its own separate text view, then subtitle text could only be for the paid promotion label. Someone needs to check this out though.

ILoveOpenSourceApplications:
@inotia00 or @0xrxL, any suggestions?

0xrxL:
> @inotia00 or @0xrxL, any suggestions?

Can't see it (continue watching). But move it to another setting.

OrdinaryPerson0:
> Selecting other text views doesn't show anything. So how do we distinguish, the paid promotion label from others?
> 
> Screenshot_20260609_130840_YouTube

@ILoveOpenSourceApplications  This is off-topic, but what's the name of the app in the screenshot?

ILoveOpenSourceApplications:
https://play.google.com/store/apps/details?id=com.drhowdydoo.layoutinspector

0xrxL:
> Possibly we don't distinguish? And we leave this as part of the hide subtitle?

If the premium badge is always visible for non-premium/premium users, you could use that as check to distinguish one subtitle from another, by checking childViews.

ILoveOpenSourceApplications:
@LisoUseInAIKyrios, you have yt premium right? Can you capture the component for "Continue watching"?

LisoUseInAIKyrios:
`modern_miniplayer_continue_watching_premium` is a LinearLayout

ILoveOpenSourceApplications:
So, "Continue watching" uses its own separate component?

LisoUseInAIKyrios:
Yes.

I don't remember what the extension method `hideMiniplayerSubTexts()` hides but it may be the parent view of both the subtitle and the continue watching view.

ILoveOpenSourceApplications:
I have never seen the video title showing up in miniplayer though. Is that really a thing?

LisoUseInAIKyrios:
The subtitle could show the channel the video is playing on.

It's been a while since I saw that, and I always use miniplayer 1 which doesn't have a subtext, so maybe things change changed. But the layout subtext view is still there.

ILoveOpenSourceApplications:
Does hide "Hide subtext" setting hide the above mentioned "Continue watching" part?

LisoUseInAIKyrios:
I checked now and hiding subtext does not hide premium continue watching label

ILoveOpenSourceApplications:
Well then, shall we move it to hide ads as it seemingly only shows the paid promotion label and nothing else?

LisoUseInAIKyrios:
Ok sure.

LisoUseInAIKyrios (approved):

LisoUseInAIKyrios:
The hide subtext/paid promotion extension logic or patch could be adjusted.

When the setting is on the Premium resumable miniplayer has no buttons and no overlays, so it's impossible to resume the video and the only way to close it is to swipe it away. I'm not sure how long it has been like this.

ILoveOpenSourceApplications:
Yeah, I just tested, and it's not hiding the label for me. Need to adjust the code by the looks of it, to be 1:1 of what it was inside the miniplayer patch.

LisoUseInAIKyrios:
the simplest may be to move the code back to MiniplayerPatch, but remove the UI setting for the subtext and change extensions MiniPlayerPatch to use the Ads setting for the subtext hiding.

ILoveOpenSourceApplications:
I think I fixed it in the dev. Can you test it now?

Issue #1300, opened 2026-04-26
bug(YouTube): "Target music app package name" doesn't work


moonfight1:
### Bug description

i have the target music app package name set to `com.metrolist.music` and metrolist is installed, but it doesnt open metrolist and instead opens this url: http://market.android.com/details

https://github.com/user-attachments/assets/248a7305-36cc-4102-b7ee-eb8faa7f9f36

### App version patched

20.47.62 (1.24.0)

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

moonfight1:
what does that have to do with my issue?

mysterious-myself:
Noting i just test a before/after compsrison comment template

ILoveOpenSourceApplications:
Did you restart after setting the package name?

moonfight1:
yes

ILoveOpenSourceApplications:
It's working perfectly fine for me.
<details><summary>Details</summary>
<p>

https://github.com/user-attachments/assets/416da480-ceed-4a5f-a3c7-5cc3839106e3

</p>
</details>

moonfight1:
well what relevant info can i provide?

- android 9 (one ui 1)

is that relevant?

moonfight1:
wait why are your settings different than mine?
<details><summary>Details</summary>
<p>

<img width="810" height="3177" alt="Image" src="https://github.com/user-attachments/assets/1f748b92-3846-4f4a-a618-469020089bf2" />

</p>
</details>

ILoveOpenSourceApplications:
I use expert mode when patching, and only include patches that are needed for my use case, that's why it looks different. 

Where did you download metrolist from?

moonfight1:
> Where did you download metrolist from?

IzzyOnDroid (through Droid-ify)

jpa102:
what if you get metrolist from this site? they have a repo here

`com.metrolist.music` from github redirects just fine for me, maybe yours has a different package name if i'm not wrong

moonfight1:
it does not have a different package name, `pm list packages | grep metrolist` shows `com.metrolist.music`, and i retyped the package name in morphe just to make sure

jpa102:
odd, i'll try reproducing your error with the exact app version and patches you stated to verify

mine is at 21.17.480 and 1.25.0-dev.4 patches

jpa102:
[...]

PR #1707, opened 2026-06-09
fix(YouTube - Override YouTube Music buttons): Target app opens when clicking on 'YouTube Music' button inside explore menu

ILoveOpenSourceApplications:
### Description

N/A

### Additional context

N/A

### Test results


ILoveOpenSourceApplications:
I’m hitting a wall with Android 8 (Oreo) and 9 (Pie) compatibility. There seems to be a strict difference in how these specific OS versions parse or sanitize incoming Intents compared to newer versions.

With my current patch, the target music app successfully opens, but the actual song playback fails to trigger. I've tried several variations of Intent flags and URI formatting, but I can't get the track to play. I've out of ideas here and if anyone has deeper experience with Oreo/Pie Intent routing quirks, I'd they take a look at #1300.

PR #1762, opened 2026-06-15
feat(YouTube Music): Add `SponsorBlock` patch

yaldar:
## Summary

Adds **SponsorBlock for YouTube Music**:

- Automatically skips enabled segment categories (sponsor, non-music section, intro/outro, self-promo, interaction, etc.) during playback.
- Dedicated top-level **SponsorBlock** settings section with its own icon, per-category enable switches, and per-category **seekbar color pickers**.
- Draws colored segment markers on the now-playing seekbar.

### Supporting changes
- New **music VideoInformation** patch (video id / time / length hooks + a `seekTo` bridge) that SponsorBlock depends on.
- Cross-version fix to the miniplayer previous/next patch and a new **9.23.52** target, so the feature works on current YouTube Music.

## How it works
- Segments are fetched from the SponsorBlock API on each new track and skipped via the player's `broadcastCurrentProgress` time hook + a `seekTo` bridge added to the player class.
- Enabled categories and colors are plain settings, force-registered through the music `Settings` holder so they resolve in the preferences screen.

## Testing
Verified by patching a real **YouTube Music 9.23.52** APK with `morphe-cli` and running on-device:
- Full default patch set + SponsorBlock apply together (no fingerprint failures).
- Skipping confirmed on tracks with sponsor and non-music segments.
- Settings section, color pickers, and seekbar markers render.
- Existing miniplayer patch still applies on 9.23.52.

## Credits / inspiration
- [inotia00/revanced-patches](https://github.com/inotia00/revanced-patches) (ReVanced Extended) — music SponsorBlock & VideoInformation patches.
- [ReVanced/revanced-patches](https://github.com/ReVanced/revanced-patches) — original SponsorBlock & video-information patch patterns.
- [SponsorBlock](https://sponsor.ajay.app) ([ajayyy/SponsorBlock](https://github.com/ajayyy/SponsorBlock)) by Ajay Ramachandran and contributors — segment API/data and the logo used for the section icon.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


extensions/music/src/main/java/app/morphe/extension/music/sponsorblock/SegmentPlaybackController.java
```diff
@@ -0,0 +1,237 @@
...
+            if (!MusicSponsorBlockSettings.SB_ENABLED.get()
+                    || segments == null || segments.length == 0) return;
+
+            final float speed = VideoInformation.getPlaybackSpeed();
+            final long lookAhead = (long) (speed * 1200);
+            final long startLookAhead = millis + lookAhead;
+
+            SponsorSegment upcomingFound = null;
+            SponsorSegment currentFound = null;
+
```
LisoUseInAIKyrios, line 112: This entire PR is a huge copy paste and it duplicates all code. We really need to minimize copying and instead move the SB code into the shared YT/Music extension

yaldar:
@LisoUseInAIKyrios Agreed on the duplication, thanks for the catch. The Music SB re-implements YouTube's `SponsorSegment` / `CategoryBehaviour` / `SBRoutes` / `SBRequester` / `SegmentCategory` rather than sharing them, and that should be fixed before merge.

Before I rewrite, I want to align on the boundary. Proposed layering, with the shared code living in the `shared-youtube/library` extension (package `app.morphe.extension.shared.sponsorblock`, since both YouTube and Music already depend on it):

**Move to shared:**
- `SponsorSegment` (pure data)
- `CategoryBehaviour` (enum)
- `SBRoutes` and `SBRequester` core (parameterized by API URL, enabled categories, and the toast/connection settings)
- a `SegmentCategory` base (category key, color, behaviour resolution)

**Stays app-specific (genuinely differs):**
- the `SegmentPlaybackController` glue: video time/length/seek hooks differ per app, and YouTube additionally has the segment UI, voting, stats, channel whitelist
- the settings keys/holders

Two questions so I scope it right:

1. Do you want me to also migrate the existing YouTube SB onto the shared core in this PR (true dedup, but it touches shipped code and I would re-test YouTube), or extract the shared core and rebase Music on it now and migrate YouTube in a follow-up?
2. Any preferred package name or location other than `app.morphe.extension.shared.sponsorblock` in `shared-youtube/library`?

Happy to go whichever way you prefer.

extensions/music/src/main/java/app/morphe/extension/music/sponsorblock/SegmentPlaybackController.java
yaldar: @LisoUseInAIKyrios Fair point, I agree. The music version copies too much from the youtube one.

I think these can move into the shared-youtube library so both apps share them:
SponsorSegment, CategoryBehaviour, SBRoutes, the requester, and a base
SegmentCategory. The playback controller and the settings would stay per-app
since they work differently.

Before I redo it: do you want the existing youtube SB switched over to the
shared code in this same PR, or should I just do the shared part plus music now
and move youtube in a follow-up? Also, is shared-youtube/library the right place
for it, or somewhere else?

yaldar: @LisoUseInAIKyrios checking in with you again. Which do you want?

LisoUseInAIKyrios: We're still deciding how to do this with SponsorBlock for YT Music that will minimize future maintenance.

I think nearly all the code can be in youtube-shared, and add various set/configure methods that YT and YT Music specific patches can call into to configure the code how it should work.

ILoveOpenSourceApplications:
This PR is now superseded by #1842. Thanks for the contribution though.

PR #1773, opened 2026-06-17
fix(YouTube Music - Track crossfade): Dismissing the queue does not stop playback, fix songs incorrectly skipped

### LINKED ISSUE #1671: bug(YouTube  Music): Track crossfade fails when using dismiss queue or swipe to dismiss miniplayer
### Bug description

Today I tried an artifact file of a new patch called `swipe to dismiss miniplayer` from here [here](https://github.com/MorpheApp/morphe-patches/pull/1420#issuecomment-4637957759). I tested on version 9.22.53 and 8.51.51. It was playing music in background even after swiped to dismiss the miniplayer & the crossfade was conflicting with it. After a restart of the player when swipe again to close the miniplayer, it’s showing the same issue again & again like this. 👇👇

https://github.com/user-attachments/assets/64df146b-6558-4069-a07f-85d985925866

In the video I already closed the miniplayer but still music is playing in background but after disable the crossfade no issues found like this. 

# There are TWO methods to reproduce this issue

### App version patched

9.22.53 & 8.51.51

### Error logs

```shell

```

### Acknowledgements

### LINKED ISSUE #1754: bug(YouTube Music): Crossfade on auto-advance skips a song
### Bug description

Step 1: Enable Crossfade and Crossfade on auto-advance
Step 2: Play a song and wait for it to end
Result: Song 1 will crossfade into song 2 but as soon as the fade is over the player will skip to song 3. This means every other track is skipped.

Tested on 8.51.51 and 8.47.56 with 0.31.0 patches. 0.30.0 didn't have this issue.

### App version patched

8.51.51

### Error logs

```shell

```

### Acknowledgements

VazerOG:
> [!IMPORTANT]
> **Track crossfade now requires YouTube Music 9.x or newer.** On 8.x it's a no-op. The 8.x ExoPlayer listener setup never worked properly with the dual-player swap, and the auto-advance regression below is 8.x-only — so crossfade is 9.x+ from here on.

### Description

Several crossfade fixes bundled together:

- **Restrict crossfade to 9.x+.** See the note above. The about text now states the 9.x+ requirement, and the patch no-ops on older versions.
- **Dismiss queue / swipe-to-dismiss no longer break playback (#1671).** Both come through as a `stopVideo(5)`, same as a manual skip, so the patch was starting a crossfade onto a player that never gets a track — the old song kept playing in the background and the player stayed wedged until a force-close. It now hooks the dismiss paths (the `DismissWatchEvent` handler for the menu, and the swipe-dismiss handler) and suppresses/tears down the crossfade instead. A timeout-based recovery stays as a fallback.
- **Repeat-one now crossfades the song into itself.** Auto-advance always fired `MEDIA_NEXT`, which ignores repeat-one and jumps to the next track. It now reads the loop mode and, when it's repeat-one, reloads the same track from 0 on a fresh player so the loop actually crossfades.
- **Crossfade works on Android Auto again.** The cast/mirror guard was treating AA's `REMOTE_SUBMIX` audio capture as casting and disabling crossfade, even though AA decodes locally. `REMOTE_SUBMIX` is no longer in that list (real Chromecast is still caught through MediaRouter).
- **One-time toast when crossfade is disabled for casting/mirroring**, so it isn't a silent mystery.

Closes #1671
Closes #1754

### Additional context

N/A

### Test results

Patching verified on 9.10.52, 9.11.54, 9.12.51, 9.14.51 and 9.15.50 (all fingerprints resolve). Runtime tested on 9.15.50, including over Android Auto.


LisoUseInAIKyrios:
9.15 seems to work well. We can change YT Music recommended stable target to that.

sr-shishir:
Can I get the artifact file for testing?

LisoUseInAIKyrios:
https://github.com/MorpheApp/morphe-patches/actions/runs/27676937868/artifacts/7690388158

sr-shishir:
https://github.com/user-attachments/assets/0e63077d-9a00-4f57-af01-ab08fce09988

Okay but in this update when swipe to dismiss miniplayer is enabled, crossfade sometimes works and sometimes doesn’t. If the phone is locked it doesn’t work at all. The same old issue is happening again for me. I tested on versions 9.22.53 and 9.15.51 — same issue.

Need more testing from others.

VazerOG:
> lv_0_20260617202940.mp4 
> Okay but in this update when swipe to dismiss miniplayer is enabled, crossfade sometimes works and sometimes doesn’t. If the phone is locked it doesn’t work at all. The same old issue is happening again for me. I tested on versions 9.22.53 and 9.15.51 — same issue.
> 
> Need more testing from others.

@sr-shishir can you provide a log export, please?

VazerOG:
Not a patch log... I'm looking for debugging logs that are obtainable from:
* YTM >> Settings >> Morphe >> Miscellaneous >> Debugging

Make sure debug logging is enabled, reproduce the issues you're seeing, and post the log here.

sr-shishir:
Strange... After enabling debug logging crossfade working perfectly no issues even with the S to DM on. Still I'm testing..

sr-shishir:
Nah bro. no issues face like that video I dropped  after enabling debug logging. I don’t know what to say now. 😑

VazerOG:
Interesting... I'm going to test with it turned off now. Because that's interesting.

VazerOG:
Pushed two follow-up fixes for issues that surfaced after the dismiss work, both on 9.x.

### 1. Audio gap when skipping after a queue dismiss

A few of you (and I) noticed that once you swipe-to-dismiss the player and start playing again, every manual skip had a ~half-second pause before the next track came in — and a force-close made it disappear until the next dismiss.

The dismiss fires `onPauseVideo`, which flips an internal "is playing" flag to `false`. Normally `onPlayVideo` flips it back, but crossfade loads the next track on its own player and never goes through that path — so the flag got stuck on `false` for the rest of the session. The next skip read it, assumed the outgoing track was paused, and muted it while the incoming track was still buffering → silence for the swap window, then the new song. Force-closing only "fixed" it by resetting the process.

Fix: set the flag back to `true` at the point the crossfade actually starts the incoming track playing, which is exactly what `onPlayVideo` would have done. Genuine pauses still behave correctly.

### 2. Crossfade dying on the lock screen (the one @sr-shishir hit)

This is the intermittent "crossfade doesn't work when the screen is locked" report. I couldn't reproduce it on my Samsung, and the device split turned out to be the whole story — it reproduces on Xiaomi/MIUI but not on lighter ROMs.

When a song auto-advances, the new player sits briefly in `IDLE` while YouTube Music kicks off the load, then moves to `BUFFERING`. The dismiss work added a safety net that tears the crossfade down if the player stays `IDLE` too long (the assumption being a dismissed/ended queue never loads anything). The problem: that `IDLE` dwell stretches when the device throttles background work with the screen off. On my Samsung it was ~210ms unlocked vs ~400ms locked — already most of the way to the 600ms cutoff — and MIUI's heavier throttling pushes it over, so the safety net was killing perfectly good locked auto-advances.

Fix: only treat a stuck-`IDLE` player as a dead load when no `loadVideo` was ever issued for it (a real dismiss / end-of-queue). If a load *was* issued, the content is on its way no matter how slow the device is, so we wait for it instead of bailing. I also bumped the cutoff from 600ms to 2s as a second layer for the case where the device is throttled enough to delay the load call itself. Real dismisses are unaffected — they're handled directly by the dismiss hook.

---

Both are on `fix/crossfade-followups`. The first is verified on my device; the second I can't reproduce locally since it's specific to aggressive background throttling. @sr-shishir — if you get a chance to build this branch and test auto-advance with the screen locked on your Xiaomi, I'd love to confirm it's actually fixed before this goes in.

sr-shishir:
I think it’s ok now. Even on lock screen didn’t face the crossfade dying issue. You can go ahead.

VazerOG:
@sr-shishir 
Re-patch with [this artifact](https://github.com/MorpheApp/morphe-patches/actions/runs/27711447929/artifacts/7704651571), with debug logging **_turned off_** and let me know how that goes.

sr-shishir:
Hmm tested with debug logging _turned off_. Everything works perfect. Even on Lock Screen also works well. 😊

VazerOG:
@LisoUseInAIKyrios Ready. Still need to bump recommended version of YTM... I personally am recommending 9.15.51, but that's your call.

LisoUseInAIKyrios:
YT Music recommended is now [9.15.51](https://github.com/MorpheApp/morphe-patches/commit/4ef13c7a0475c1b9f3aeac3613711df81806b461)

Will merge tomorrow if no new concerns

MrGapi:
I found a bug. 
-Youtube version: 9.15.51
-Using the latest artifact

Description:
If the mini-player is dissmised, while the cross-fade is in progress, song will keep playing. 

See the video.

https://github.com/user-attachments/assets/29ffee05-fc73-4649-97e4-aa89c4e55b80

VazerOG:
@MrGapi, try now with latest artifact.

MrGapi:
Now it works as intended.

MrGapi:
@VazerOG I repatched to YouTube music version: 9.23.52 and the crossfade is completly broken. It spits out a bunch of errors like: "CrossfadeManager: Could not find coordinator from atad". In patcher logs can be found this:
Screenshot_20260618_003407_Morphe

9.15.xx version seems to work.

VazerOG:
Actually, nevermind... fix incoming. 

Previous: 
> @MrGapi YTM has refactored code that my fingerprint doesn't look for on that version it seems. Fix will be included in a later release.

VazerOG:
@MrGapi give that a shot.

VazerOG:
#1754 resolution confirmed.

Ready for merge.

MrGapi:
@VazerOG Tried on 9.23.xx and it looks good. Crossfade is working and patcher logs are as they should be.

MrGapi:
@VazerOG One more issue I found in version 9.23.52: after dismissing the player and fully closing the app, the next time I launch it, a bar appears at the bottom showing 'nothing is playing.' It can't be swiped away or dismissed. Playing a song makes it return to normal.

Screenshot_20260618_065700_YT Music Morphe

sr-shishir:
> @VazerOG Tried on 9.23.xx and it looks good. Crossfade is working and patcher logs are as they should be.

Thanks bro. for trying on 9.23.xx. I'm on 9.22.xx & don't want to move from here. So far so good to me. 😄

VazerOG:
@MrGapi it almost feels like normal behavior to me. I didn't even think of that as an issue until you pointed it out. 

 How much do you hate it?

MrGapi:
@VazerOG I can get used to it. Just visual clutter, if you know what I mean.

VazerOG:
@MrGapi I do.

krvstek:
Great feature! Minor UI nitpick though: wouldn't it be better to move the VazerOG credit to the main About/Credits section alongside current/prior development and licenses?

Having an author shoutout inside a specific feature's menu breaks consistency and clutters the view.

MarcaDian:
But SB and RYD also have similar information

krvstek:
> But SB and RYD also have similar information

SB and RYD are external 3rd-party services and APIs. They often have specific attribution requirements in their terms, and their names act as proper nouns for the services themselves.

Crossfade is a standard open-source code contribution. Typically, individual code contributions don't get dedicated UI shoutouts per feature, that's exactly what GitHub's commit history and eventually the general `Credits/About` section are for. If we follow this logic, we'd have to put a contributor's name under every single toggle they added.

VazerOG:
> > But SB and RYD also have similar information
> 
> SB and RYD are external 3rd-party services and APIs. They often have specific attribution requirements in their terms, and their names act as proper nouns for the services themselves.
> 
> Crossfade is a standard open-source code contribution. Typically, individual code contributions don't get dedicated UI shoutouts per feature, that's exactly what GitHub's commit history and eventually the general `Credits/About` section are for. If we follow this logic, we'd have to put a contributor's name under every single toggle they added.

Partially. I believe that about section is reserved for full Morphe Team members and I have not been invited to join the team, plus I'm the only one so far who's working on this patch.

I'll leave it up to the team to decide how to handle this. If anything I would rather remove my credit strings within the feature patch menu to maintain consistency.

krvstek:
> I'll leave it up to the team to decide how to handle this. If anything I would rather remove my credit strings within the feature patch menu to maintain consistency.

I'd be down for adding a dedicated `Contributors` section in the credits that fetches the list straight from GitHub, though honestly, the commit history and GitHub's Contributors tab are probably enough.

Either way, whatever the team decides, amazing job on the patch, it's a highly needed feature for YT Music (:

PR #1823, opened 2026-06-24
feat: Add `Network proxy` patch

### LINKED ISSUE #1444: feat(YouTube  Music): Bypass regional restrictions
### Feature description

Is your feature request related to a problem? Please describe.**
In South Korea, YouTube Music does not offer a "Free Tier" at all. Because of this, even when using patched versions, the app detects the local IP and displays an error message: **"YouTube Music is not available in your area"** or prevents playback entirely. This makes the patch unusable for Korean users without a separate VPN.
**Describe the solution you'd like**
I would like to request a feature that can bypass this regional restriction. Specifically:
 * An option to **spoof the app's location** or region (e.g., to the US or other free-tier supported countries).
 * A fix to prevent the "Unavailable in your country" popup from appearing in restricted regions like South Korea.
**Describe alternatives you've considered**
Currently, the only workaround is using a VPN, but it is inconvenient, slows down the connection, and often gets disconnected.
**Additional context**
Since the South Korean market is strictly "Premium only" for YouTube Music, a regional bypass is essential for the patch to function properly here.

### Motivation

Currently, South Korea is one of the very few regions where Google strictly prohibits the YouTube Music free tier. This means that without a Premium subscription, the app is completely locked, and even with the current Morphe patches, users are still blocked by the "Country Unavailable" wall.
Adding a regional bypass or location spoofing feature would:
 1. **Expand Accessibility:** Make Morphe functional for a massive user base in South Korea who are currently unable to use it.
 2. **Enhance Utility:** Provide a seamless experience without the need for external tools like VPNness, which are often unstable.
 3. **Complete the Patch Experience:** Truly unlock the potential of the app by overcoming the toughest regional restriction enforced by Google.

### Acknowledgements

naijun0403:
### Description

Adds a YouTube Music proxy patch that lets users route supported app traffic through a user-configured HTTP or HTTPS proxy.

In some countries, including South Korea, access to the YouTube Music free tier may be unavailable due to local regulatory or licensing restrictions. This patch provides an in-app proxy option so users can route supported requests through a region where the service is available.

### Additional context

- Supports both current `ProxyOptions.fromProxyList(List, int)` and legacy `ProxyOptions.fromProxyList(List)` Cronet proxy APIs.
- Some older supported versions, such as `7.29.52`, do not include the Cronet proxy APIs required by this patch. On those versions, patching still succeeds, but the proxy hook is intentionally left as a no-op to avoid crashes.
- Proxy authentication passwords are excluded from settings import/export.

### Test results


patches/src/main/kotlin/app/morphe/patches/music/misc/proxy/ProxyPatch.kt
```diff
@@ -0,0 +1,59 @@
...
+    compatibleWith(COMPATIBILITY_YOUTUBE_MUSIC)
+
+    execute {
+        if (SetProxyOptionsFingerprint.methodOrNull == null ||
+            CreateHttpProxyFingerprint.methodOrNull == null ||
+            (FromProxyListFingerprint.methodOrNull == null &&
+                    FromProxyListWithFallbackBehaviorFingerprint.methodOrNull == null)
+        ) {
+            return@execute
+        }
```
LisoUseInAIKyrios: Do all of these fingerprints resolve with YT Music 7.29.52, 9.15.51, and 9.24.51?

If all resolve then no orNull logic is needed and it can throw an exception if it stops resolving in future app targets (someone needs to update the patch).

If some of these fingerprints don't resolve on certain versions, then version checks can be done like [this](https://github.com/MorpheApp/morphe-patches/blob/adc371ba9b3e221d091bbd4bbb8e9213a82c55a3/patches/src/main/kotlin/app/morphe/patches/music/layout/miniplayer/EnableSwipeToDismissMiniplayerPatch.kt#L166).

extensions/music/src/main/java/app/morphe/extension/music/patches/ProxyPatch.java
```diff
@@ -0,0 +1,209 @@
...
+        try {
+            return (ProxyOptions) ProxyOptions.class
+                    .getMethod("fromProxyList", List.class, Integer.TYPE)
+                    .invoke(
+                            null,
+                            proxies,
+                            allowDirectFallback
+                                    ? ALL_PROXIES_FAILED_BEHAVIOR_ALLOW_DIRECT
+                                    : ALL_PROXIES_FAILED_BEHAVIOR_DISALLOW_DIRECT
+                    );
```
LisoUseInAIKyrios: The stubbed class is declared. Can this work without reflections and call the method directly?

naijun0403: Yes, the stub only helps compilation. The actual available methods still depend on the Cronet version bundled in each target APK.

A direct-call fallback could be implemented with `try/catch (NoSuchMethodError)`, so reflection is not the only possible approach. I used reflection here to avoid emitting direct `invoke-static` references to overloads that are absent on part of the supported version range.

The supported APKs do not expose the same overloads: newer 9.x builds only have `fromProxyList(List, int)`, `8.51.51` only has `fromProxyList(List)`, and `7.29.52` has neither. With direct calls, the extension DEX would contain hard references to both overloads and rely on failed method resolution for fallback.
Reflection makes the version-dependent API selection explicit by resolving the available overload first, then invoking it.

patches/src/main/kotlin/app/morphe/patches/music/misc/proxy/ProxyPatch.kt
naijun0403: Good point. This has been [updated](https://github.com/MorpheApp/morphe-patches/pull/1823/changes/2d00ab5d08224ac80898d30c0e4fa82720c56d44) to use explicit version checks instead of a silent `methodOrNull` guard.

Versions older than `8.50` now return early with a warning because their bundled Cronet does not include the required proxy APIs. For versions where the proxy API should exist, the patch now resolves the expected fingerprints through `.method`, so future fingerprint breakage fails visibly instead of silently no-oping.

The selected `fromProxyList` fingerprint is version-specific:
- `9.20+`: `fromProxyList(List, int)`
- `8.50` to pre-`9.20`: legacy `fromProxyList(List)`

extensions/music/src/main/java/app/morphe/extension/music/patches/ProxyPatch.java
LisoUseInAIKyrios: It's ok to have calls to methods that don't exist, as long as they are wrapped in an if check (or called from separate extension methods that are conditionally called). Android code does with Build version checks to allow selectively calling classes/methods that only exist in certain versions of Android.

There is an extension version check patch for YT Music (app.morphe.extension.music.patches.VersionCheckPatch), and it could do something like:
```
if (IS_9_00_OR_GREATER) { // Or whatever version YT Music uses the method with an int
    return ProxyOptions.fromProxyList(proxies, allowDirectFallback
            ? ALL_PROXIES_FAILED_BEHAVIOR_ALLOW_DIRECT
            : ALL_PROXIES_FAILED_BEHAVIOR_DISALLOW_DIRECT);
}
return ProxyOptions.fromProxyList(proxies); // 8.x
```

ILoveOpenSourceApplications:
The patch name can be changed as well to `Bypass regional restriction` from the issue opened here or something similar than just `Proxy`.

extensions/music/src/main/java/app/morphe/extension/music/patches/ProxyPatch.java
LisoUseInAIKyrios: Actually, hold off on that just yet. I'm trying some other refactoring and a version check in the extensions makes it more complex.

naijun0403: Got it, I’ll hold off on the extension-side version check for now.

For reference, the boundary between the two `fromProxyList` overloads is around `9.20`, so the change I was about to commit added an `IS_9_20_OR_GREATER` check to select the correct overload.

naijun0403:
> The patch name can be changed as well to `Bypass regional restriction` from the issue opened here or something similar than just `Proxy`.

I’d prefer to keep the patch name focused on what it implements rather than the specific use case.

The patch adds configurable HTTP/HTTPS proxy support. Regional availability is one reason someone may use it, but the patch does not directly patch licensing, entitlement, or region-check logic. Naming it `Bypass regional restriction` could imply a broader behavior than the implementation provides.

I’m open to a slightly more descriptive name like `Network proxy` or `Music proxy`, but I think `Proxy` is more accurate than a region-bypass-focused name.

LisoUseInAIKyrios:
I think `Network proxy` sounds good.

naijun0403:
> I think `Network proxy` sounds good.

`Network proxy` sounds good to me. It describes the implemented behavior more clearly without implying that the patch modifies region or entitlement logic itself.

I updated the patch name and settings screen title here: https://github.com/MorpheApp/morphe-patches/pull/1823/changes/73cc918715175061cc3e7d3a4997887bb6088f32

LisoUseInAIKyrios:
I had to rebase to fix the dev branch that changed.

The patch also works with YouTube.

extensions/music/src/main/java/app/morphe/extension/music/patches/ProxyPatch.java
LisoUseInAIKyrios: I remove the reflections. Please verify everything still works because I don't have a proxy to try this on.

naijun0403: Tested all supported YouTube and YouTube Music versions, and confirmed there are no mismatches or runtime issues with the proxy API version split.

LisoUseInAIKyrios:
Will merge tomorrow if there's no new concerns.

error-reporting:
why Youtube 20.31.42 patched with morphe 1.33.0 patch version didn't have proxy settings even though the network proxy patch is included at patch time 

<details>

Screenshot_20260703_204749
Screenshot_20260703_204518
Screenshot_20260703_204755

</details>

LisoUseInAIKyrios:
Because YouTube is missing the required code that allows proxies.

In expert mode it prints a warning stating the patch isn't available. The patch description could be updated to mention it requires YT 20.51.39+ or Music 8.51+

error-reporting:
> Because YouTube is missing the required code that allows proxies.
> 
> In expert mode it prints a warning stating the patch isn't available. The patch description could be updated to mention it requires YT 20.51.39+ or Music 8.51+

those mentioned newer app versions I think are practically experimental at this point as there are some issues on them that also happen on unpatched app of the same version

LisoUseInAIKyrios:
If we could have easily supported all older app versions, we would have.

Issue #1088, opened 2026-04-03
bug(YouTube): Pressing back exits the app instead of returning to previous scroll position


amdffsharma6543-cpu:
### Bug description

Back button exits app instead of returning to previous position

### App version patched

1.23.0-dev.3

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

amdffsharma6543-cpu:
<details><summary>Details</summary>
<p>

```
LithoFilterPatch: Searching ID: video_lockup_with_attachment.eml-fe|96c46b8c4b1a61f2 Path: video_lockup_with_attachment.eml-fe|96c46b8c4b1a61f2|CellType|video_lockup.eml-fe|70e729ede8a9750|ContainerType|video_lockup_metadata.eml-fe|cd741a693108baf9|ContainerType|ContainerType|lockup_metadata.eml-fe|8aedc9d8ccf76701|ContainerType|ContainerType|
LithoFilterPatch: Searching ID: video_lockup_with_attachment.eml-fe|96c46b8c4b1a61f2 Accessibility: eml.lockup_metadata Path: video_lockup_with_attachment.eml-fe|96c46b8c4b1a61f2|CellType|video_lockup.eml-fe|70e729ede8a9750|ContainerType|video_lockup_metadata.eml-fe|cd741a693108baf9|ContainerType|ContainerType|lockup_metadata.eml-fe|8aedc9d8ccf76701|ContainerType|
LithoFilterPatch: Searching ID: video_lockup_with_attachment.eml-fe|96c46b8c4b1a61f2 Accessibility: eml.overflow_button Path: video_lockup_with_attachment.eml-fe|96c46b8c4b1a61f2|CellType|video_lockup.eml-fe|70e729ede8a9750|ContainerType|video_lockup_metadata.eml-fe|cd741a693108baf9|ContainerType|ContainerType|overflow_button.eml-fe|65bb605b41237eb9|ImageType|
LithoFilterPatch: Searching ID: video_lockup_with_attachment.eml-fe|96c46b8c4b1a61f2 Path: video_lockup_with_attachment.eml-fe|96c46b8c4b1a61f2|CellType|video_lockup_attachments.eml-fe|58bd2553701799f|ContainerType|
[...]

PR #1877, opened 2026-06-29
chore(YouTube): Fix predictive back function for newer targets

0xrxL:


0xrxL:
This PR solved an issue that exists on last versions, which use predictive gestures classes instead the MainActivity's one.

ILoveOpenSourceApplications:
Does this solve #1088?

0xrxL:
> Does this solve #1088?

Nope. This only allows the back to exit fix (the patch that allows to close the app with the back button) to works for newer versions.

patches/src/main/kotlin/app/morphe/patches/youtube/misc/fix/backtoexitgesture/Fingerprints.kt
```diff
@@ -53,3 +55,23 @@ internal object BackToRefreshFeatureFlagFingerprint : Fingerprint(
...
     )
 )
+
+internal object PredictiveGesturesOnBackInvokedFingerprint : Fingerprint(
+    classFingerprint = Fingerprint(
+        name = "onBackCancelled",
+        accessFlags = listOf(AccessFlags.PUBLIC, AccessFlags.FINAL),
+        returnType = "V",
+        parameters = listOf(),
+        filters = listOf(
```
LisoUseInAIKyrios, line 65: This works for 20.51.39+

But 20.21 and 20.31 the code is very different and I cannot find anything that resembles the 20.51 code

0xrxL: I don't think it's needed for so older versions. Did you tried to add a logger to check if the old fingerprint works in that case? In this way we can use a version check.

LisoUseInAIKyrios: What version YT does this problem first appear?

0xrxL: I think after 20.51 more or less.

For this reason I think it's better to check if the new fingerprint is effectively needed for 20.21, or 20.30 for example.

0xrxL:
Checked on older versions. Can be merged.

PR #1885, opened 2026-06-29
fix(YouTube Music - Scrobbling): Parse artist and track from video title

Nai64:
**Problem:** For reuploaded content, YouTube Music's MediaSession provides the channel name as the artist and the full video title (e.g. "Artist - Track") as the track name, causing incorrect scrobbles.

**Solution:** Added a new opt-in setting `SCROBBLING_PARSE_TITLE` that, when enabled, splits the video title on " - " to extract the real artist and track name before scrobbling.

**Changes:**
- `Settings.java`: New `SCROBBLING_PARSE_TITLE` boolean setting (default: off)
- `ScrobbleManager.java`: Parsing logic in `resolveTitleAndArtist()` splits title on first " - ", applies existing cleanup to both parts
- `strings.xml`: Title and summary strings for the new setting
- `ScrobblingPatch.kt`: `SwitchPreference` added to the metadata cleanup category

**How to test:**
1. Enable scrobbling and the new "Parse artist and track from title" switch
2. Play a reuploaded track (where the channel name differs from the artist)
3. Check the scrobble; artist and track should now be correct

### DIFF
--- a/patches/src/main/resources/addresources/values/music/strings.xml
+++ b/patches/src/main/resources/addresources/values/music/strings.xml
@@ -231,6 +231,8 @@ YouTube Music 8.x — crossfade only runs on 9.x and newer"</string>
     <string name="morphe_music_scrobbling_metadata_cleanup_summary">Automatically clean common clutter (like \"Official Video\", \"Remaster\", \"Live\") from titles, artists, and albums before scrobbling</string>
     <string name="morphe_music_scrobbling_custom_regex_title">Custom regex cleanup</string>
     <string name="morphe_music_scrobbling_custom_regex_summary">Enter a custom regular expression to remove additional text from song titles, artists, or albums. Example: (?i)\\s*\\((clean|explicit)\\)</string>
+    <string name="morphe_music_scrobbling_parse_title_title">Parse artist and track from title</string>
+    <string name="morphe_music_scrobbling_parse_title_summary">When enabled, extracts artist and track from the video title by splitting on " - " instead of using YouTube Music\'s metadata. Fixes reuploads where the channel name is used as the artist and the full video title is used as the track name. Combine with custom regex to strip remaining tags.</string>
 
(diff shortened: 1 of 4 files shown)


patches/src/main/resources/addresources/values/music/strings.xml
```diff
@@ -231,6 +231,8 @@ YouTube Music 8.x — crossfade only runs on 9.x and newer"</string>
     <string name="morphe_music_scrobbling_metadata_cleanup_summary">Automatically clean common clutter (like \"Official Video\", \"Remaster\", \"Live\") from titles, artists, and albums before scrobbling</string>
     <string name="morphe_music_scrobbling_custom_regex_title">Custom regex cleanup</string>
     <string name="morphe_music_scrobbling_custom_regex_summary">Enter a custom regular expression to remove additional text from song titles, artists, or albums. Example: (?i)\\s*\\((clean|explicit)\\)</string>
+    <string name="morphe_music_scrobbling_parse_title_title">Parse artist and track from title</string>
+    <string name="morphe_music_scrobbling_parse_title_summary">When enabled, extracts artist and track from the video title by splitting on " - " instead of using YouTube Music\'s metadata. Fixes reuploads where the channel name is used as the artist and the full video title is used as the track name. Combine with custom regex to strip remaining tags.</string>
```
LisoUseInAIKyrios: Ideally this text is a lot shorter because translations may be too long and the text is clipped.

Nai64: Changed the summary. I hope it is not long enough now. A lot of people will get confused if the summary is too short.

LisoUseInAIKyrios:
Nice and simple. Looks good, thanks.

para2ox:
Hi! Is it possible to add custom separators, rather than just " - "? For example, I have tracks named "Artist – Track".

Issue #687, opened 2026-03-02
feat(YouTube Music): Theme adapts to the current song/dynamic theme


FTBSchnitzel:
### Feature description

A bit like the Youtube Music Widget or the Player, but Globally. There is also already an Option for Theming the Miniplayer. This just needs to be applied Globally.

Here are some Examples from the Releases images from https://github.com/pear-devs/pear-desktop/

<img width="1552" height="811" alt="Image" src="https://github.com/user-attachments/assets/0f11acd2-417e-418e-a282-10e5464ab27b" />

### Motivation

Dynamic Colors based on the Song looks really great. It looks so great it is already in the WidgetPlayer, just not Globally. Instead of being limited to Theme you set, your Song sets the Color. Morphe already has this as an option for just the miniplayer, so i would think it is not much harder.

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] This is a feature request for YouTube, YT Music, or Reddit

PR #1908, opened 2026-07-03
feat(YouTube Music - Change miniplayer color): Add "Change navigation bar color" setting

MarcaDian:
### Description

Extends the `Change miniplayer color` patch to propagate the fullscreen player's dominant color to the navigation bar TabLayout, with a child "Match navigation bar color" toggle for opting out. The TabLayout view is captured at draw time so the bar repaints together with the miniplayer.

### Test results

___


MarcaDian:
I wanted to add a toggle to the subscreen `morphe_music_navigation_bar_screen`, but the current code does not allow key duplication (merging), so I added this setting to the `PLAYER`.

ILoveOpenSourceApplications:
I wanted to also check if it was possible if the "Dismiss queue" is used, then the miniplayer will change its color back to Grey / Theme color. Currently, it still keeps the color of the last played album art.

MarcaDian:
It seems to be working now. Bytecode isn't really my thing, so it might need a better implementation.

MarcaDian:
I wonder, is it possible to hook player color, the app background, or the flyout menu background and apply the track's accent color to them? 🤔

https://github.com/MorpheApp/morphe-patches/issues/687

patches/src/main/resources/addresources/values/music/strings.xml
```diff
@@ -184,6 +184,8 @@ YouTube Music 8.x — crossfade only runs on 9.x and newer"</string>
     <!-- layout.miniplayer.changeMiniplayerColorPatch -->
     <string name="morphe_music_change_miniplayer_color_title">Change miniplayer color</string>
     <string name="morphe_music_change_miniplayer_color_summary">Matches the miniplayer color to the fullscreen player</string>
+    <string name="morphe_music_match_navigation_bar_color_title">Change navigation bar color</string>
+    <string name="morphe_music_match_navigation_bar_color_summary">Matches the navigation bar color to the miniplayer</string>
```
MarcaDian:
```suggestion
    <string name="morphe_music_match_navigation_bar_color_summary">Matches the navigation bar color to the fullscreen player</string>
```

PR #1917, opened 2026-07-03
feat(Reddit): Add `Incognito keyboard` patch

eifr:
Adds a new `Incognito keyboard` toggle to the Reddit Layout settings (API 29+).

**Extension side**: New `IncognitoKeyboardPatch.java` registers an `ActivityLifecycleCallbacks` that detects focus changes on `TextView`s. When the toggle is enabled, it calls `TextView.onCreateInputConnection()` to generate the `EditorInfo`, adds `IME_FLAG_NO_PERSONALIZED_LEARNING`, and restarts the IME session via `startInputAsync()` (API 30+) or reflection-based `startInput()` (API 22-29).

**Patcher side**: Hooks into `FrontpageApplication.onCreate` to initialize the lifecycle hooks. Also provides a `modifyEditorInfo(EditorInfo)` injection point for future use with direct `onCreateInputConnection` hooks.

### DIFF
--- a/extensions/reddit/src/main/java/app/morphe/extension/reddit/patches/IncognitoKeyboardPatch.java
+++ b/extensions/reddit/src/main/java/app/morphe/extension/reddit/patches/IncognitoKeyboardPatch.java
@@ -13,36 +13,22 @@ import android.os.Bundle;
 import android.view.View;
 import android.view.ViewTreeObserver;
 import android.view.inputmethod.EditorInfo;
-import android.view.inputmethod.InputConnection;
 import android.view.inputmethod.InputMethodManager;
 import android.widget.TextView;
 
 import androidx.annotation.NonNull;
 
 import java.lang.ref.WeakReference;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
-
 import app.morphe.extension.reddit.settings.Settings;
-import app.morphe.extension.shared.Logger;
 
 @SuppressWarnings("unused")
 public final class IncognitoKeyboardPatch {
     private static boolean initialized;
 
-    /**
-     * @return If this patch was included during patching.
-     */
     public static boolean isPatchIncluded() {
         return false;  // Modified during patching.
     }
 
-    /**
-     * Injection point for direct EditorInfo modification in onCreateInputConnection.
-     * Called from injected bytecode - adds the incognito flag to the EditorInfo.
-     *
-     * @param editorInfo The EditorInfo to modify (may be null).
-     */
     public static void modifyEditorInfo(EditorInfo editorInfo) {
         if (editorInfo == null) return;
         if (!Settings.INCOGNITO_KEYBOARD.get()) return;
@@ -51,10 +37,6 @@ public final class IncognitoKeyboardPatch {
         editorInfo.imeOptions |= EditorInfo.IME_FLAG_NO_PERSONALIZED_LEARNING;
     }
 
-    /**
-     * Initialize the keyboard lifecycle hooks.
-     * Called from application onCreate via injected bytecode.
-     */
     public static void initialize(Application application) {
         if (initialized) return;
         initialized = true;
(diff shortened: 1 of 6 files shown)


LisoUseInAIKyrios:
The purpose of this is so the user's Reddit usage does not add to the device keyboard internal autocomplete suggestions?

eifr:
> The purpose of this is so the user's Reddit usage does not add to the device keyboard internal autocomplete suggestions?

Exactly, basically enabling incognito keyboard when using Reddit like when using incognito in chrome

imPratyoosh:
> > The purpose of this is so the user's Reddit usage does not add to the device keyboard internal autocomplete suggestions?
> 
> Exactly, basically enabling incognito keyboard when using Reddit like when using incognito in chrome

The keyboard also won't learn typed words! 🤨

eifr:
> The keyboard also won't learn typed words! 🤨

Gboard learns and improve the autocomplete based on typed words..

eifr:
@LisoUseInAIKyrios hey, can we merge this?

LisoUseInAIKyrios:
Can you rebase off latest dev branch?

I'm not sure if this adds much value, but it is simple and reddit isn't overloaded with patches yet.

eifr:
> Can you rebase off latest dev branch?
> 
> I'm not sure if this adds much value, but it is simple and reddit isn't overloaded with patches yet.

@LisoUseInAIKyrios :heavy_check_mark:

extensions/reddit/src/main/java/app/morphe/extension/reddit/patches/IncognitoKeyboardPatch.java
```diff
@@ -0,0 +1,144 @@
...
+                Logger.printException(() -> "onCreateInputConnection failed", ex);
+                return;
+            }
+            if (ic == null) return;
+
+            editorInfo.imeOptions |= EditorInfo.IME_FLAG_NO_PERSONALIZED_LEARNING;
+
+            try {
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+                    imm.startInputAsync(ic, editorInfo);
```
LisoUseInAIKyrios: Double check this because it doesn't seem to compile.

LisoUseInAIKyrios: `startInputAsync` appears to be an internal and not public method.

eifr: try now? :pray:

LisoUseInAIKyrios: Please try the changed I pushed.

0xrxL:
Any update about this? Looks like in stale.

MarcaDian:
i think because of this https://github.com/MorpheApp/morphe-patches/pull/1917#discussion_r3608808831

eifr:
Had a couple of issues when I tested it last time 
I'll try to test it and fix it in the upcoming days

kitadai31:
To digress a bit, I wondered if it might be possible to patch Gboard rather than Reddit to achieve this.

This patch allows users to toggle Gboard's Incognito mode when using the keyboard on Reddit.
However, since Gboard itself still does not provide an Incognito mode toggle, you and users have to develop a patch for each app in which they wish to use Incognito mode.

Upon searching for existing solutions for this, I found that a new "Incognito Mode Toggle" patch for Gboard had been added to a community patch bundle just a week ago.
This might be a better approach than adding an incognito mode toggle only to Reddit.

> Incidentally, this patch did not yet exist at the time this pull request was opened.
> There was an "Always-incognito mode" patch in another community patch bundle, but since it didn't add a runtime toggle in the Gboard app, it wasn't sufficient to replace this Reddit-specific patch.
> (That said, there is also a patch that change the package name of the patched Gboard, so one option was to switch between the stock Gboard and the patched Gboard to toggle incognito mode.)

LisoUseInAIKyrios:
That sounds more appropriate.

It's also more certain to work, and there's no future maintenance of Reddit patch code. Reddit is notorious for regularly making broad codebase changes and patches regularly need updating.

PR #1916, opened 2026-07-03
feat(YouTube - Miniplayer): Add "Disable playback with horizontal drag" setting

0xrxL:

PR #1920, opened 2026-07-04
feat(YouTube - Miniplayer): Add "Disable automatic reposition" setting

0xrxL:


0xrxL:
@roniui This is what you're looking for, to fix miniplayer position issue. Try it by enabling the feature from miniplayer settings:

https://github.com/MorpheApp/morphe-patches/actions/runs/28724179003/artifacts/8086683884

roniui:
Working,  thanks 👍

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -1029,6 +1039,8 @@ If later turned off, it is recommended to clear the app data to prevent UI bugs.
     <string name="morphe_miniplayer_disable_horizontal_drag_summary">Prevents closing the miniplayer by swiping it horizontally off-screen</string>
     <string name="morphe_miniplayer_disable_horizontal_drag_playback_title">Disable playback with horizontal drag</string>
     <string name="morphe_miniplayer_disable_horizontal_drag_playback_summary">Playback will pause when the miniplayer is dragged off screen</string>
+    <string name="morphe_miniplayer_disable_horizontal_reposition_title">Prevents horizontal drag automatic reposition</string>
```
0xrxL: Feel free to fix it. 👍

ILoveOpenSourceApplications:
Just to be clear, this setting locks the miniplayer on whatever position the user kept, right?

patches/src/main/resources/addresources/values/youtube/strings.xml
0xrxL: > I'd rather not push a commit (or in my case another PR) and add myself as a contributor to your PR just for a single-word change. Could you just apply the fix on your end?

Do not worry of contributions. Everyone deserves to be a contributor of a PR here, if the change help to improve the PR itself. 👍

0xrxL:
> Just to be clear, this setting locks the miniplayer on whatever position the user kept, right?

Yes. It locks the miniplayer in offscreen position, without leave the app to change its state. Look at these videos posted by @roniui:

https://github.com/MorpheApp/morphe-patches/pull/1916#issuecomment-4881378319

patches/src/main/resources/addresources/values/youtube/strings.xml
MarcaDian: You can create review lines directly from the browser.

krvstek:
```suggestion
    <string name="morphe_miniplayer_disable_horizontal_reposition_title">Prevent horizontal drag automatic reposition</string>
```

ILoveOpenSourceApplications:
The setting title can be made compact?

0xrxL:
> The setting title can be made compact?

🤷‍♂️

Hmdmu98:
But we need an option that remembers the location of the trigger in the top corners when we place it there.

0xrxL:
> But we need an option that remembers the location of the trigger in the top corners when we place it there.

It's already automatic I believe.

0xrxL:
> But we need an option that remembers the location of the trigger in the top corners when we place it there.

Or you want to remember the vertical place too, to lock the trigger in that exact point (also if the keyboard overlap it)?

roniui:
> > But we need an option that remembers the location of the trigger in the top corners when we place it there.
> 
> Or you want to remember the vertical place too, to lock the trigger in that exact point (also if the keyboard overlap it)?

Yes, that is the original yt behavior,  and when keyboard is there, it will repositioned it over keyboard (vertically).

Hmdmu98:
> > But we need an option that remembers the location of the trigger in the top corners when we place it there.
> 
> It's already automatic I believe.

Only when you place it at a certain angle at the top and then press it twice, it remembers the location; however, if you move it, it forgets and falls down, which is annoying. I hope that an option can be developed to force him to do this always. Thank you very much for your efforts.

0xrxL:
I'll add a way to remember the vertical position so. 👍

PR #1924, opened 2026-07-05
fix(YouTube - Return YouTube Dislike): Remove RYD Shorts due to YT removing Shorts dislike and RYD API no longer providing reliable Shorts dislikes data

LisoUseInAIKyrios:
### Description

In June 2026 YouTube removed the dislikes button from the Shorts player.

While a Shorts player dislike button could be restored in a few ways (none of which will require 2,000+ lines of AI generated code), there is a now bigger problem with the RYD Shorts data itself.

As of now, the RYD API is still accepting likes/dislikes for Shorts video id's. But with no way for users to dislike in the Shorts viewer, that means it's been overwhelmingly collecting only likes for Shorts and biasing the data to "Everyone loves this Short".

Checking RYD API for new Shorts with a huge number of views (such as world cup videos with 1.5M or more views), shows tens of thousands of likes but zero or nearly zero dislikes. RYD for Shorts is no longer usable.

Because the Shorts RYD data is no longer useful, opening a Short in the regular player will not show any estimated dislikes even though a dislike button is present because the Shorts data is no longer reliable.

Removing RYD Shorts coincidentally fixes the Shorts player loading delay, because there no longer is any workarounds to prevent audio only video playback. So the Shorts player now open immediately or almost immediately, when before it could take up to 5 seconds to open as it waited for the RYD API to respond. 

### Test results


0xrxL:
Do not forget to remove the option to hide the dislike button.

pikoUsername:
(sorry for bad english) I think that it would be better to left dislike button because its a way faster way to signal to algorithm to not to show shorts in feed, ultimately "Do not show" is worse UX to do the same. i think there no need to load RYD for shorts (hopefully someone will make pr for shorts and it will be accepted for RYD extensions)

LisoUseInAIKyrios:
The RYD Shorts data is now worthless and showing the data is misleading at best and a lie at worse.

Any PR to restore the shorts dislikes will not be accepted for the reasons mentioned above.

PR #1931, opened 2026-07-05
fix(YouTube Music - Downloads): Add "Override feed/search flyout button" setting

### LINKED ISSUE #1928: bug(YouTube Music):  Music starts downloading as soon as a track starts playing
### Bug description

Normally, the download option on the external downloader launches when I press ‘Download’, but in this case it launches the moment I tap on the selected track. 

### App version patched

9.26.51

### Error logs

```shell

```

### Acknowledgements

LisoUseInAIKyrios:
- Fixes download action opening immediately for some users.
- Adds a separate setting to override flyout menu download action

### Test results

### DIFF
--- a/extensions/music/src/main/java/app/morphe/extension/music/patches/DownloadsPatch.java
+++ b/extensions/music/src/main/java/app/morphe/extension/music/patches/DownloadsPatch.java
@@ -68,6 +68,9 @@ public final class DownloadsPatch {
     }
 
     private static void launchExternalDownloader(String videoId) {
+        cachedFlyoutVideoId = "";
+        downloadButtonLabel = "";
+
         ExternalDownloaderPreference.launchExternalDownloader(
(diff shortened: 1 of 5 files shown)


extensions/music/src/main/java/app/morphe/extension/music/patches/DownloadsPatch.java
```diff
@@ -221,7 +228,7 @@ public static boolean commandResolverOnClick(ProtocolBufferFieldInterface p1, Ma
             }
 
             final boolean isDownloadClick = Utils.containsAny(p1String,
-                    "[133724106]", "[144224893]");
+                    "[133724106]", "[443434441]");
```
LisoUseInAIKyrios, line 231: These magic numbers are flaky and may stop working or start misbehaving at anytime.

The long term solution is use the same fingerprints and extension logic from the new YT `Add to queue` patch to override YT Music flyout download.

PR #1957, opened 2026-07-08
fix(YouTube - Playback speed): Support new native playback speed menu

### LINKED ISSUE #1655: bug(YouTube): Playback speed is not saved when using native speed dialog
### Bug description

Playback speed doesn't save. YouTube 21.21.80, 

### App version patched

patches 1.30.0.

### Error logs

```shell

```

### Acknowledgements

0xrxL:
This patch adds the necessary hook to fetch the current playback speed from the new native panel. This patch is not done yet, because I had some problems to resolve all the components needed to keep the compatibility.

The hook is called twice at the video start and everytime the playback is changed through the panel. This will entirely replace the currently patch for the onItemClick() method.


0xrxL:
One final note: I do not know how well this solution works when used alone. But it's worth trying, as if the result were positive, it could greatly simplify the current patch.

0xrxL:
Despite the initial difficult, I was finally able to unravel the patch skein and remove unnecessary ones.

This PR is now ready to be reviewed.

LisoUseInAIKyrios (commented):
Needs a tweak or change to fix 'remember playback speed' setting on video load, otherwise everything seems to work.

PR #1958, opened 2026-07-08
refactor: Merge `Video ads` into `Hide ads`

ILoveOpenSourceApplications:
### Description

N/A


patches/src/main/kotlin/app/morphe/patches/music/ad/video/HideVideoAdsPatch.kt
```diff
@@ -0,0 +1,9 @@
+package app.morphe.patches.music.ad.video
+
+import app.morphe.patcher.patch.bytecodePatch
+import app.morphe.patches.music.ad.hideAdsPatch
+
+@Deprecated("Use 'Hide ads' instead.")
+val hideVideoAdsPatch = bytecodePatch {
+    dependsOn(hideAdsPatch)
+}
```
LisoUseInAIKyrios: We can delete these instead of deprecating. Patches are not published as a library.

PR #1972, opened 2026-07-09
feat(YouTube - Hide layout components): Add "AI channel filter" settings

### LINKED ISSUE #1022: feat(YouTube): AiSList AI channel filtering
### Feature description

https://aisloplist.com/

https://github.com/Override92/AiSList

Copypaste: Community-maintained lists of YouTube channels that primarily use AI-generated content. These lists are used by the AiBlock for Youtube browser extension to help you filter AI-generated content on YouTube.

### Motivation

Less AI slop. Need I say more? 

### Acknowledgements

MarcaDian:
### Description

Adds an "AI channel filter" that hides YouTube feed and search videos from channels flagged as AI-generated content, with per-list toggles for Home/Search and a stats section with 24h and all-time counts. Lists are auto-refreshed every 4 hours from the community-maintained AiSList (https://aisloplist.com) (GitHub (https://github.com/Override92/AiSList)) which provides the blocklist and warnlist data.

### Test results

___


DonDae001:
Is there an easy way to report AI slop channels with this integration? The README on their GitHub says there's an option to report in their extension

MarcaDian:
> Is there an easy way to report AI slop channels with this integration? The README on their GitHub says there's an option to report in their extension

We did not receive an answer yet https://github.com/MorpheApp/morphe-patches/issues/1022#issuecomment-4917977983

LisoUseInAIKyrios:
Adding a "report channel is ai slop" flyout menu item to the home/search feed will not be simple.

Maybe a "report AI channel" share option could be added to the built in share sheet, and then it's two taps to report. But maybe that's not simple either.

0xrxL:
> Adding a "report channel is ai slop" flyout menu item to the home/search feed will not be simple.
> 
> Maybe a "report AI channel" share option could be added to the built in share sheet, and then it's two taps to report. But maybe that's not simple either.

You can override the report button and build a custom flyout with two buttons: one for ai slop reporting and another one for the common reporting with the original runnable. But this solution will be problematic for old versions.

0xrxL:
...unless I'll not add a way to save the onItemClick() listener of each button. 🤔

Override92:
Hey,
thanks for your effort and sorry for the late reply.
First of all, here is the API doc for submitting channels:

### Base URL

| Env  | URL |
|------|-----|
| Prod | `https://api.aisloplist.com` |

### Field rules

| Field | Rule |
|-------|------|
| `channel_handle` | 3–30 chars; letters/digits/`_`/`-`/`.`/`@`; not all-numeric; no `://` or domain-style `.` |
| `video_url`      | Must match `youtube.com/watch?v=…`, `youtube.com/shorts/…`, or `youtu.be/…` |
| `username`       | Non-empty string |

### Responses

| Status | Meaning | Body (text/plain) |
|--------|---------|-------------------|
| `200` | Accepted | `Channel <handle> with url <url> and username <name> submitted successfully` |
| `400` | Bad JSON / missing field / invalid handle or URL | `Error parsing JSON: …` |
| `405` | Wrong method (only `POST` allowed) | *(empty)* |
| `409` | DB conflict on persist | `Error while saving to database: …` |
| `429` | Rate limit hit | `Rate limit exceeded.` |
| `500` | Internal/forwarding failure | `Error while forwarding request to backend: …` |

### Example

curl -sS -X POST https://api.aisloplist.com \\
  -H 'Content-Type: application/json' \\
  -d '{"channel_handle":"@ExampleChannel","video_url":"https://youtu.be/dQw4w9WgXcQ","username":"alice"}'
  


I’m going to have a look at the PR now

Override92:
.../patches/components/AiSListFilter.java
`    /** Refresh the cached list from GitHub raw after this long since last successful fetch. */
    private static final long REFRESH_CHECK_INTERVAL_MS = 4 * 60 * 1000L;`
    
This is unintended, right? According to PR description this should be 4 hours if I'm not wrong.

LisoUseInAIKyrios:
Yes that's a typo. Needs another `* 60`

Override92:
One more thing I noticed while reading through. in parseList() (AiSListFilter.java:100-105) the trie callback accepts every hit:


```
(text, startIndex, matchLength, callbackParam) -> {
    ((MutableReference<String>) callbackParam).value = handle;
    return true;
};
```

So I think a listed handle would also match when it happens to be a prefix of a different channel's handle. [at]drew would, if I'm reading this right, also match [at]drewbinsky and hide that channel's videos.

I see KeywordContentFilter guards its callback with keywordMatchIsWholeWord() (KeywordContentFilter.java:192) from the shared base class BufferPhraseFilter. That helper on its own may not quite cover handles, though: it only rejects a match when the adjacent code point is a letter (BufferPhraseFilter.java:384-392), while handles can also contain digits, _, - and . — so [at]drew could still match [at]drew_official or [at]drew.tv. My own list already has [at]drew_ponder sitting next to [at]drew, so that shape does occur in practice.

If it's useful, one option would be to check the trailing boundary against the handle character set, reusing the existing getUtf8CodePointAt() helper (BufferPhraseFilter.java:418). The leading side shouldn't need a check, as @ can't occur inside a handle and is already part of the pattern:


```
private static boolean endsAtHandleBoundary(byte[] text, int endIndex) {
    Integer cp = getUtf8CodePointAt(text, endIndex);
    return cp == null || !(Character.isLetterOrDigit(cp)
            || cp == '_' || cp == '-' || cp == '.');
}
```
with the callback gated by:


`if (!endsAtHandleBoundary(text, startIndex + matchLength)) return false;`

I hope I've described it clearly.

LisoUseInAIKyrios:
Thanks for point that out.

I believe it's fixed with https://github.com/MorpheApp/morphe-patches/commit/24fd63e4749f50822008ce42521de8b504e83ecd

[YouTube](https://support.google.com/youtube/answer/11585688) says middle dot (·) is also a valid channel character.

HanabishiRecca:
Why is it not a standalone patch that can be disabled? I understand the intent here, but I don't want my client to constantly fetch data from random sources in the internet. Especially when I don't need that functionality in the first place.

LisoUseInAIKyrios:
It's not fetching anything for you.

The feature is disabled by default, and the list is fetched only if the feature is enabled.

HanabishiRecca:
I'm not so sure.

<details>
<summary>Details</summary>



</details>

LisoUseInAIKyrios:
I'll double check if it's fetching when the feature is off.

If it's fetching when it's off then it'll be fixed in the next pre-release.

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/AiSListFilter.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    @Override
+    protected void reparseIfNeeded() {
+        final long now = System.currentTimeMillis();
+        final long lastCheck = lastRefreshCheckMs.get();
+        if (now - lastCheck > REFRESH_CHECK_INTERVAL_MS
+                && lastRefreshCheckMs.compareAndSet(lastCheck, now)) {
+            Utils.runOnBackgroundThread(AiSListRequester::fetchAndStore);
+        }
```
HanabishiRecca, line 56: Glancing over the code, I don't see it ever checking for the feature being enabled.

MarcaDian: Fixed in https://github.com/MorpheApp/morphe-patches/commit/a35a429f971d97c6179af8a76c19c235bf2ca1e8

PR #2022, opened 2026-07-13
fix(Network proxy): Support HTTPS proxies for auxiliary requests

### LINKED ISSUE #1998: bug(YouTube/ YT Music): HTTPS proxy doesn't route all requests through the proxy
# HTTPS proxy doesn't route all requests through the proxy

## Description

I'm using the Network proxy patch with YouTube Music to bypass region restrictions in my country. I set up an HTTPS proxy (with Proxy HTTPS enabled) so that my proxy credentials aren't sent in cleartext over the network.

However, I noticed that when using an HTTPS proxy, not all requests go through the proxy. Some requests seem to fall back to a direct connection, which means they still use my real IP address. This causes inconsistent behavior — some parts of the app work fine, but others fail because the region restriction is still triggered for those requests.

If I turn on "Allow direct fallback," these requests silently bypass the proxy. If I turn it off, they just fail with a connection error. Neither option works well for region bypass.

With an HTTP proxy, everything works perfectly — all requests are routed through the proxy. But I'd prefer to use HTTPS to protect my credentials.

## Steps to reproduce

1. Set up an HTTPS proxy server (TLS between client and proxy)
2. In Morphe settings for YouTube Music, configure:
   - Proxy enabled: on
   - Proxy host and port: your HTTPS proxy
   - Proxy HTTPS: **on**
   - Allow direct fallback: on
3. Open YouTube Music and use it from a region where it's restricted
4. Observe that some requests work but others fail or behave inconsistently

## Expected behavior

When an HTTPS proxy is configured, all app network requests should be routed through the proxy — not just the main streaming/API requests via Cronet, but also any auxiliary requests that use `HttpURLConnection`.

## What I found

I looked at the source code and found that `openUrlConnection` in `NetworkProxyPatch.java` explicitly handles this case:

```java
if (config.httpsProxy) {
    if (config.allowDirectFallback) {
        return (HttpURLConnection) url.openConnection();  // falls back to direct
    }
    throw new IOException("HttpURLConnection proxy requests only support HTTP proxies");
}
```

So it seems like Java's `HttpURLConnection` doesn't natively support HTTPS proxies, and the current implementation just gives up on those requests.

## Suggestion

Would it be possible to add HTTPS proxy support for these requests? I understand it's not straightforward since Java's standard library doesn't support it out of the box, but maybe one of these approaches could work:

- **Custom TLS tunneling**: Establish a TLS connection to the proxy first, then send the HTTP CONNECT request over it. This would only use standard library classes like `SSLSocketFactory` / `SSLSocket`.
- **Route everything through Cronet**: Since Cronet already handles HTTPS proxies, maybe `HttpURLConnection`-based requests could be redirected through the Cronet engine too.
- **At minimum, a UI warning**: When the user enables HTTPS proxy, show a warning that some requests may not go through the proxy and could expose the user's real IP address.

I think this would be really helpful for users in restricted regions who need reliable proxy coverage for all app traffic.

## Environment

- App: YouTube Music
- App version: 9.15.51
- Patch version: v1.34.0

naijun0403:
### Description

Routes auxiliary extension requests through the app-wide, proxy-configured Cronet engine when an HTTPS proxy is enabled.

Previously, HTTPS proxy settings only covered the app's main Cronet traffic. Requests opened through `Requester` or direct `HttpURLConnection` call sites either bypassed the proxy when direct fallback was enabled or failed when it was disabled. This also prevented authenticated proxy users from completing spoof player and OAuth requests.

This change:

- Captures the app-wide Cronet engine without retaining transient engines
- Routes spoofing, OAuth, scrobbling, thumbnail, translation, transcript, and caption-cookie requests through the proxy-aware connection provider
- Preserves proxy authentication and the configured connect/read timeouts
- Only applies direct fallback when it's enabled and the request can still be retried safely
- Recovers from an unavailable Cronet engine without replaying requests whose I/O has already started

Closes #1998

### Additional context

Direct fallback remains opt-in. HTTPS proxy requests no longer bypass the proxy just because Java's `HttpURLConnection` can't represent an HTTPS proxy.

The patch is a no-op on YouTube below `20.47.62` and YouTube Music below `8.51.51`, where the required Cronet proxy APIs aren't available.

### Test results

### DIFF
--- a/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/requests/proxy/DelegatingHttpURLConnection.java
+++ b/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/requests/proxy/DelegatingHttpURLConnection.java
@@ -5,7 +5,7 @@
  * See the included NOTICE file for GPLv3 Section 7 terms that apply to this code.
  */
 
-package app.morphe.extension.shared.patches;
+package app.morphe.extension.shared.requests.proxy;
 
 import java.io.IOException;
 import java.io.InputStream;
--- a/extensions/music/src/main/java/app/morphe/extension/music/patches/scrobbling/lastfm/LastFM.java
+++ b/extensions/music/src/main/java/app/morphe/extension/music/patches/scrobbling/lastfm/LastFM.java
@@ -12,7 +12,6 @@ import org.json.JSONObject;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.net.HttpURLConnection;
-import java.net.URL;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
@@ -24,6 +23,7 @@ import java.util.Map;
 
 import app.morphe.extension.music.patches.scrobbling.ScrobbleManager;
 import app.morphe.extension.shared.Logger;
+import app.morphe.extension.shared.requests.Requester;
 
 public class LastFM {
     private static final String BASE_URL = "https://ws.audioscrobbler.com/2.0/";
@@ -94,8 +94,7 @@ public class LastFM {
 
         byte[] postDataBytes = postData.toString().getBytes(StandardCharsets.UTF_8);
 
-        URL url = new URL(BASE_URL);
-        HttpURLConnection conn = (HttpURLConnection) url.openConnection();
+        HttpURLConnection conn = Requester.openConnection(BASE_URL);
         conn.setRequestMethod("POST");
         conn.setRequestProperty("User-Agent", USER_AGENT);
         conn.setRequestProperty("Content-Type", "application/x-www-form-urlencoded");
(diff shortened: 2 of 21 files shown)


extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/requests/proxy/DelegatingHttpURLConnection.java
```diff
@@ -0,0 +1,333 @@
...
+import java.net.URL;
+import java.security.Permission;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Base class for an {@link HttpURLConnection} that forwards operations to another connection.
+ */
+abstract class DelegatingHttpURLConnection extends HttpURLConnection {
```
LisoUseInAIKyrios, line 24: All these url connection classes could be moved to `app.morphe.extension.shared.requests`

Or a new subpackage since this is specific to cronet:
`app.morphe.extension.shared.requests.cronet`

naijun0403: This was my mistake and has been fixed in https://github.com/MorpheApp/morphe-patches/pull/2022/changes/a9df02463a823a6669c4c2116df0d112bf0c742a

LisoUseInAIKyrios:
Looks good from here.

Ready to merge?

naijun0403:
> Looks good from here.
> 
> Ready to merge?

Yes, it's ready to merge. I don't see any remaining issues from my side.

PR #2035, opened 2026-07-14
fix(YouTube - StreamOrDetailsDataRequest): Sometimes player headers is null

0xrxL:
While I was using the "Save to Watch Later" button, I encountered a blocking issue during the initial phases of the code to fetch the response associated with that button. The problem occurs when I switch the app to the background and then bring it back to the foreground, and than using the button. I suspected the `Future` was the culprit, so I created a method to force its cancellation if the video details fetch takes longer than a reasonable timeframe (5 seconds).

This is currently an experimental attempt; it will be merged if it proves successful.


0xrxL:
UPDATE: The issue was instead localized during currentVideoRequestHeaders set. So I moved the code for doing this in a more safe point, hopefully this is enough to fix the issue.

PR #2066, opened 2026-07-17
fix(YouTube - Downloads): Add `;` to one of the fingerprint

sayand734:
### Description

Closes #

### Additional context

N/A

### Test results


LisoUseInAIKyrios:
The fingerprint still works because it uses startsWith matching.

Will fix this the next time the fingprint is patched.

Thanks.

Edit: I can't remember but this may have matched both `String` and `StringBuilder`.

PR #2084, opened 2026-07-18
bump(YouTube): Add experimental support for `21.29.363`

LisoUseInAIKyrios:
### Description

In 21.29 YouTube has removed all modern miniplayers except the default modern 4, including the 'minimal' miniplayer type.

Previously the miniplayer type was an int field and used in ~ 60 different places to give the different behavior of each player. 21.29 removes that field and all code using it. It seems infeasible to restore the old miniplayers with so much code removed.

Patch 21.28 or lower if you want a different miniplayer type (rip Modern 1 and minimal).

### Test results


MarcaDian:
The minimal miniplayer included a few fixes for icons and their colors. This can now be removed.

PR #2101, opened 2026-07-19
chore(YouTube Music - Scrobbling): Replace hardcoded user agent with dynamic version

krvstek:
### Description

Replaces hardcoded User-Agent and submission client version strings with dynamic values derived from `Utils.getAppVersionName()` across LastFM and ListenBrainz. Renames submission client from `YT Music Morphe` to `Morphe` and extracts a `CLIENT_VERSION` constant in ListenBrainz.

### Additional context

N/A

### Test results

### DIFF
--- a/extensions/music/src/main/java/app/morphe/extension/music/patches/scrobbling/listenbrainz/ListenBrainz.java
+++ b/extensions/music/src/main/java/app/morphe/extension/music/patches/scrobbling/listenbrainz/ListenBrainz.java
@@ -23,7 +23,8 @@ import app.morphe.extension.shared.requests.Requester;
 
 public class ListenBrainz {
     private static final String BASE_URL = "https://api.listenbrainz.org/";
-    private static final String USER_AGENT = "YT Music Morphe (https://github.com/MorpheApp/morphe-patches)";
+    private static final String USER_AGENT = "Morphe/" + Utils.getAppVersionName();
+    private static final String CLIENT_VERSION = Utils.getAppVersionName();
     
     public static class TokenValidation {
         public boolean valid;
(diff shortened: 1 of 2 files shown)


extensions/music/src/main/java/app/morphe/extension/music/patches/scrobbling/listenbrainz/ListenBrainz.java
```diff
@@ -23,7 +23,8 @@
 
 public class ListenBrainz {
     private static final String BASE_URL = "https://api.listenbrainz.org/";
-    private static final String USER_AGENT = "YT Music Morphe (https://github.com/MorpheApp/morphe-patches)";
+    private static final String USER_AGENT = "Morphe/" + Utils.getAppVersionName();
+    private static final String CLIENT_VERSION = Utils.getAppVersionName();
```
LisoUseInAIKyrios: This is the version of YT Music.

Should this instead use the patches version? 
`Utils.getPatchesReleaseVersion()`

I don't use scrobbling so I'm unsure which is more appropriate.

krvstek: I actually just used the same user-agent as SponsorBlock here.

However, we could use a user-agent that includes both the app version and the patch version, for example: `Morphe/[PatchVersion] (YTMusic/[AppVersion])`. As for the `submission_client`, we can strictly use the patches version exclusively.

Does that sound good to you?

LisoUseInAIKyrios: Yeah sounds good.

Issue #541, opened 2026-02-15
bug(YouTube): Video player hangs when redirecting from Shorts player


RichardReallyHard:
### Bug description

YT 20.40.45 patch 1.13.0-dev.2

how to reproduce:
1. (at anywhere) open a shorts video with video link (long video) or scroll to this kind of shorts
2. click that link to open in normal video player
3. weird bugs:
- video player stalls, but thumbnail & duration loads instantly
- below player parts (everything below seekbar) takes very long to load
- clicking play button will plays the short video at step 1
- returning to shorts player will change video page to minimise/popup mode, had to close it to watch other long video

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

ILoveOpenSourceApplications:
Does this issue occur on Shorts played from Home feed or only when the Shorts played are the ones from related videos section?

RichardReallyHard:
> Does this issue occur on Shorts played from Home feed or only when the Shorts played are the ones from related videos section?

I think I did tested from these starting point to open a short with video link:
(no active video page/fresh app start)
1. home feed
2. shorts player (1st video)
3. shorts player (scrolled for a bit)
4. subscriptions
5. acc profile watch history

the results are same

haven't try "open shorts inside a video page"

ILoveOpenSourceApplications:
This is due to setting a default quality for Shorts. If it's set to `Auto` then the issue isn't there anymore.

RichardReallyHard:
that kind of avoided the bug
but Auto always means low quality regardless of network speed 😂
is there any way to fix this ?
don't want to manually pick a higher quality on every playback 😥
if **_remember quality_** is ON and picked a higher quality on this shorts, the problem just returns


another similar issue (for normal videos):
is there any way to ensure a higher quality preload buffer ?
though the quality follows the default setting after this preload buffer, the preload buffer alway comes in 240p
need to manually fast forward or backward to "reload" a correct quality
isn't there a option to always skip preload buffer?

RichardReallyHard:
just discovered another bug:
open shorts player (auto quality on wifi)
-> click the video link to open in normal video player 
-> change quality (with **remember quality** for video / shorts both ON) 
-> got a notification saying the quality selected will apply to all **shorts** on wifi
(on wifi is correct but it should be **video** not shorts ?)

LisoUseInAIKyrios:
[...]

PR #2144, opened 2026-07-21
fix(YouTube - Remember video quality): Default video quality may not be applied

0xrxL:


inotia00:
iirc, this code was needed to force the Dash stream on clients that support both SABR and Dash stream (TV, visionOS).

Testing is required to see if both TV and visionOS work on YouTube 20.21.37 and YouTube 21.29.366.

LisoUseInAIKyrios (commented):
TV client seems to work, but this change breaks visionOS with 21.29.366

0xrxL:
@inotia00 @LisoUseInAIKyrios Ok. I reverted everything and converted the goal of this PR, to fix an issue that I was able to reproduce lately with default video quality set, on some videos. This issue always restart the playback, at the video starts.

ILoveOpenSourceApplications:
Can #541 be addressed in the PR as well? That also happens when quality of Shorts/Video is set to any value other than `Auto`.

LisoUseInAIKyrios:
Need to check if it's fixed by this change

PR #2202, opened 2026-07-25
chore(YouTube - Seekbar): Various UI fixes

### LINKED ISSUE #2197: bug(YouTube): "Enable seekbar thumbnails" causes issue with Shorts preview
### Bug description

Using the new enable seekbar thumbnails option causes overlapping thumbnails on YouTube shorts

Image

Please let me know if logs are required.

### App version patched

21.04.223, Morphe 1.37.0-dev.1

### Error logs

```shell

```

### Acknowledgements

### LINKED ISSUE #2209: bug(YouTube): Incorrect seekbar timing
### Bug description

YouTube app version patched - 

Patches version - 1.37.0-dev.2

The issue is that when using the seekbar to skip ahead, for example to 15:30, it actually stops at 16:03. The difference between the shown value when using the seekbar and the actual value where the video will play is not fixed; it could fluctuate between ~3 seconds to ~2 minutes. This is only an example; actual min and max were not tested.

Also, an unrelated issue, as you can see in the last 10 seconds of the video, the thumbnail seekbar preview is not always showing.

### App version patched
21.29.366

https://github.com/user-attachments/assets/377feec4-047b-4afa-adb7-478f28b84f33

### Error logs

```shell

```

### Acknowledgements

0xrxL:
- The preview thumbnail now follows only the trackball on the seekbar, rather than the tap position of the MotionEvent.
- The preview thumbnail follows the trackball on the seekbar even during a long press on the player for seeking.
- The preview thumbnail now works only on the video player.
- The preview thumbnail now disappears when dragging upwards.

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/SeekbarThumbnailPreviewPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/SeekbarThumbnailPreviewPatch.java
@@ -246,11 +246,9 @@ public class SeekbarThumbnailPreviewPatch {
                 final int previewHeightPx = currentFrameParams.height;
 
                 int targetX = trackBallPosX - (previewWidthPx / 2);
-                final int targetY =
-                        trackBallPosY -
-                        previewHeightPx -
-                        Dim.dp(THUMBNAIL_PREVIEW_DISTANCE_DP) -
-                        Dim.dp(THUMBNAIL_PREVIEW_TIMESTAMP_HEIGHT_DP);
+                final int targetY = trackBallPosY - previewHeightPx
+                        - Dim.dp(THUMBNAIL_PREVIEW_DISTANCE_DP)
+                        - Dim.dp(THUMBNAIL_PREVIEW_TIMESTAMP_HEIGHT_DP);
 
                 final int screenWidth = Dim.getScreenWidth();
                 if (targetX < 0) {
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/SeekbarThumbnailPreviewPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/SeekbarThumbnailPreviewPatch.java
@@ -12,11 +12,13 @@ import android.content.Context;
 import android.graphics.Bitmap;
 import android.graphics.Color;
 import android.graphics.Outline;
+import android.graphics.Point;
 import android.graphics.drawable.ColorDrawable;
 import android.graphics.drawable.GradientDrawable;
 import android.view.Gravity;
 import android.view.MotionEvent;
 import android.view.View;
+import android.view.ViewGroup;
 import android.view.ViewOutlineProvider;
 import android.widget.FrameLayout;
 import android.widget.ImageView;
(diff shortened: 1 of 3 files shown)


0xrxL:
@LisoUseInAIKyrios This will potentially solve your issue related to the bitmap freeze, during the first minute of video. But the fingerprints and the patch (not the integrations) should be refactored. So I'll mark this PR as a draft (despite it works).

P.S: Yeah...at the end I didn't slept at that time.

LisoUseInAIKyrios:
Everything is working. Tested on:
- 20.21.37
- 20.51.39
- 21.04.233
- 21.29.366

0xrxL:
@LisoUseInAIKyrios did you fixed the space between the seekbar and thumbnail in fullscreen? There's a height difference between portrait and landscape.

0xrxL:
I think that the action button space should be added to the fullscreen calculation

LisoUseInAIKyrios:
Fullscreen spacing looks correct to me.

I think portrait mode could use less vertical padding so the thumbnail and seektime is closer to the seekbar.

0xrxL:
> Fullscreen spacing looks correct to me.
> 
> I think portrait mode could use less vertical padding so the thumbnail and seektime is closer to the seekbar.

Yeah this needs to be tweaked, by adding the action buttons fullscreen height to the calculation, in order to align fullscreen thumbnail height to the portrait one, and than lowering both spacing to good value.

lucai9000:
Is it planned that the chapter name will be added below the thumbnail?

0xrxL:
> Is it planned that the chapter name will be added below the thumbnail?

We could add it in this PR. It wouldn't be difficult to implement...if the name would be already hooked somewhere. 🤔

inotia00:
> > Is it planned that the chapter name will be added below the thumbnail?
> 
> We could add it in this PR. It wouldn't be difficult to implement...if the name would be already hooked somewhere. 🤔

You just need to hook the class `TimelineMaker`.

This class is obfuscated on the latest YouTube.

### YouTube 20.21.37

#### Class: ajlc (Hooking point)

```
    public final TimelineMarker[] n(ajov ajov) {
        ajyy ajyy = (ajyy) this.a.get(ajov);
        if (ajyy == null) {
            return new TimelineMarker[0];
        }
        return (TimelineMarker[]) ((ajkq) ajyy.b).a.toArray(new TimelineMarker[0]);
    }
```

#### Class: com.google.android.libraries.youtube.player.features.overlay.timebar.TimelineMarker (Abstraction interface implementation point)

```
    public final String toString() {
        String valueOf = String.valueOf(this.d);
        StringBuilder stringBuilder = new StringBuilder("TimelineMarker[title=");
        stringBuilder.append(valueOf);
        stringBuilder.append(", startMillis=");
        stringBuilder.append(this.a);
        stringBuilder.append(", endMillis=");
        stringBuilder.append(this.b);
        stringBuilder.append(", markerIndex=");
        stringBuilder.append(this.c);
        stringBuilder.append(", displayProperties=null]");
        return stringBuilder.toString();
    }
```

LisoUseInAIKyrios:
Will merge this now because it fixes existing bugs. Adding the chapter title can be in another PR.

PR #2238, opened 2026-07-27
fix(YouTube - Swipe controls): Evaluate swipe zone before recycling MotionEvent

gajera1080:
## Description
Fixes a use-after-recycle bug in `BaseGestureController.submitTouchEvent()` where `isInSwipeZone(me)` was called after `me.recycle()`, reading coordinates from a recycled `MotionEvent`.

## Root Cause
`me.recycle()` returns the cloned `MotionEvent` to Android's memory pool. Calling `isInSwipeZone(me)` after recycling reads $X$ and $Y$ coordinates from recycled memory, causing undefined behavior or inaccurate zone detection.

## Solution
Evaluated `isInSwipeZone(me)` prior to calling `me.recycle()` and stored the result in a local `inSwipeZone` value.


LisoUseInAIKyrios:
Did you find this fix using ai?

Or is this entirely your own human found fix?

gajera1080:
Yes, I found and created this fix using AI assistance.

PR #2239, opened 2026-07-27
fix(YouTube Music - Bypass certificate checks): Resolve Android Auto crash

andersonlucasg3:
## Problem

On devices that have real Google Play services installed alongside microG (GmsCore), YouTube Music crashes when Android Auto connects to its `MediaBrowserService`:

```
java.lang.IllegalStateException: Missing DynamiteApplicationContext.
    at com.google.android.gms.common.GoogleCertificatesImpl.<init>(...)
    at java.lang.Class.newInstance(Native Method)
    ...
    at com.google.android.apps.youtube.music.mediabrowser.MusicBrowserService.f(...)
    ...onGetRoot(...)
```

The app's `GoogleSignatureVerifier` loads `GoogleCertificatesImpl` from the real GMS Dynamite module, whose constructor requires a `DynamiteApplicationContext` that was never initialized in the app process (GMS calls are redirected to microG by the GmsCore support patch). The exception is not caught (only `tpz`/`RemoteException` are), the process dies, Android Auto retries forever → infinite loading spinner (reported upstream in ReVanced/revanced-patches#6602 and related issues).

The current patch only neutralizes the SHA-256 partner fingerprint check, which is evaluated **after** the Dynamite-based verifier in the caller chain — so on these devices the crash happens before the patch can take effect.

## Fix

Also `returnEarly(false)` on the `GoogleSignatureVerifier` entry method (`c(String): boolean`): the check reports "not Google-signed" without touching Dynamite, and authorization in the caller still succeeds via the existing certificate fingerprint patch (`!r.c() && !m()` evaluates to false → access granted).

Behavior on devices without real GMS is unchanged: the method already returned `false` after a graceful fallback — this only skips the path that hard-crashes when GMS is present.

The new fingerprint anchors on the `"Failed to get Google certificates from remote"` log string (in `GoogleSignatureVerifier.a`) and scopes the `(String)Z` lookup to the same class — it is the only method with that signature there.

## Validation

Device: Xiaomi 17 Pro Max (popsicle), Android 16 / HyperOS 3 (xiaomi.eu), Google Play services 26.28.60 installed + GmsCore 0.3.13.2, YouTube Music 9.15.51.

- Before: `MediaBrowserService.onGetRoot` crashed with the exception above on every Android Auto connect (process death loop, spinner forever). Confirmed via Frida hooks.
- After (same smali change applied manually): `onGetRoot` serves a real browser root to `com.google.android.projection.gearhead`; the library loads on the car screen. No crashes.

- [ ] Patch compiles locally (build environment requires GitHub Packages auth for the `app.morphe.patches` Gradle plugin — not available in this dev setup)


LisoUseInAIKyrios:
Please confirm this change patches and the app works correctly with these YT Music targets:
- 9.28.51
- 9.15.51
- 7.29.52


Did you use Claude to find the method to patch? Nothing wrong with that, just confirming.

andersonlucasg3:
In fact, I used Kimi-K3. I'll test those targets tomorrow and confirm to you. 

Thanks.

andersonlucasg3:
Hey @LisoUseInAIKyrios, just finished testing the patch across all three versions you asked for. All are working fine.

andersonlucasg3:
Confirmed: the change patches cleanly and the app works correctly on all three requested targets.

Test setup:
- Patches built locally from this branch (patches-1.37.0.mpp, applied with Morphe desktop CLI 1.12.0, default patch set)
- Device: Xiaomi Redmi (23053RN02A, "fire"), Android 15, real Google Play services 26.29.31 installed + MicroG-RE 6.1.4

Results:

| YT Music | Patch applies (fingerprint resolves) | Installs & launches | Crash (Missing DynamiteApplicationContext) |
|---|---|---|---|
| 9.28.51 | ✅ | ✅ | None |
| 9.15.51 | ✅ | ✅ | None |
| 7.29.52 | ✅ | ✅ | None |

Notes:
- The new GoogleSignatureVerifier fingerprint resolved on all three versions (no fingerprint failures during patching).
- Verified via logcat after launch: no IllegalStateException / FATAL, process stays alive.
- 7.29.52 was tested with the armeabi-v7a APK (arm64 variant was not available from any accessible mirror); it runs fine on this arm64 device.
- Bonus: this also checks off the "patch compiles locally" item from the PR description — the build works once the machine has GitHub Packages credentials (gpr.user/gpr.key or GITHUB_ACTOR/GITHUB_TOKEN) and ANDROID_HOME set.

LisoUseInAIKyrios:
Need to rebase off dev branch. Otherwise looks good.

LisoUseInAIKyrios:
Thanks for the help 👍

PR #2257, opened 2026-07-29
feat: Add `Change installer source` universal patch

LisoUseInAIKyrios:
### Description

Adds a universal patch to spoof the installer source used by any internal app code.

This can fix apps that look for an app store installation source and block side loading usage.

Patch idea was first done by [inotia00](https://github.com/inotia00/revanced-patches/commit/aef0cbb7eb2fff3783f4f22a2f613e079040cd08), but the code here was wrote from scratch using Morphe patcher features.

### Test results

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -2,7 +2,7 @@
 
 [versions]
 morphe-patcher = "1.7.0"
-morphe-patches-library = "1.5.1-dev.2"
+morphe-patches-library = "1.5.2-dev.6"
 # Tracking https://github.com/google/smali/issues/100.
 smali = "d92701d947"
 agp = "9.1.0"
--- a/patches/src/main/kotlin/app/morphe/patches/all/misc/installer/ChangeInstallerSource.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/all/misc/installer/ChangeInstallerSource.kt
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2026 Morphe.
+ * https://github.com/MorpheApp/morphe-patches/pull/2257
+ *
+ * See the included NOTICE file for GPLv3 Section 7 terms that apply to this code.
+ */
+
+package app.morphe.patches.all.misc.installer
+
+import app.morphe.patcher.patch.resourcePatch
+import app.morphe.patcher.patch.stringOption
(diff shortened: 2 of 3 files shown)


0xrxL:
Do you think this should be mandatory in future?

LisoUseInAIKyrios:
Reddit has been using this for many months because some code checks the installer source. YT doesn't check anywhere.

LisoUseInAIKyrios:
There exist other installer sources such as system (preinstalled) apps, f-droid, uptodown, and other third party app stores.

But no apps are known to require those sources, so the packag names were not included with the option preset. Anyone can specify those by hand if they insist on experimenting.

sr-shishir:
I had a feeling Morphe would eventually address this. G00gle has already released an app on the Play Store that appears to install itself silently on some users' phones, and it prevents the installation of certain third-party apps. I've personally seen this happen to a few people I know. It seems G00gle has already started making its move.

Galaxyyss:
Is it the same as "install from Play Store" with Shizuku? Just as universal patch with more installation sources to spoof and choose from?

LisoUseInAIKyrios:
This patches the apps own internal checks (if any are present).

This does not change the actual installation source, and selecting "Google Play Store" with this patch will not re-enable app store updates.

MarcaDian:
What happens if user apply the patch *and* select an installation method in the manager? Will the manager take precedence?

LisoUseInAIKyrios:
The patch takes precedence, because it's bytecode patching all app code for any usage of:

- `Landroid/content/pm/PackageManager;->getInstallerPackageName(Ljava/lang/String;)Ljava/lang/String;`
- `Landroid/content/pm/InstallSourceInfo;->getInstallingPackageName()Ljava/lang/String;`
- `Landroid/content/pm/InstallSourceInfo;->getInitiatingPackageName()Ljava/lang/String;`
- `Landroid/content/pm/InstallSourceInfo;->getPackageSource()I`

LandonMoran:
Okay, but what's the point when you're there's a thing in the settings? It literally already spoofs what you installed with and the installer settings. So what is it in between? Doing it in the installer and doing it in the patching? Is there a difference

LisoUseInAIKyrios:
Exporting apks doesn't have a setting.

CLI doesn't have an installer setting yet.

Using play store installer source in Manager causes app updates to show (unless disable app updates is used).

PR #2261, opened 2026-07-29
feat(YouTube): Add `Playback in feeds` patch

### LINKED ISSUE #74: feat(YouTube): Disable video playback in feeds
### Feature description

This request proposes porting the specific patch functionality from ReVanced Extended (RVX) that adds a toggle within the mod settings to forcibly disable video playback in feeds. Instead of relying on the native YouTube "General" settings menu, which is frequently bugged or unresponsive in newer base versions. This patch would introduce a hard override switch in the Morphe settings menu. This switch would programmatically suppress the autoplay behavior in the Home and Subscription feeds regardless of the state of the native Google settings.

### Motivation

The primary motivation for this request is that the native "Playback in feeds" button in the standard YouTube settings is currently non-functional for many users on version 20.xx builds, rendering it impossible to disable autoplay through normal means. Even when troubleshooting steps such as disabling device spoofing or changing layout factors are attempted, the native menu entry often remains unresponsive to touch input due to underlying UI conflicts in the new base. Porting this patch is essential for feature parity with the discontinued RVX project and ensures that users can successfully disable data-consuming feed previews even when the stock Google settings UI is broken.

Image
Image

### Acknowledgements

MarcaDian:
### Description

Adds the `Playback in feeds` setting of YouTube to the Morphe settings, where it is always available even if YouTube hides it in the tablet layout or removes it remotely.

The mode is read from and written to YouTube itself, so the Morphe setting and the native setting always use the same value.

### Test results


MarcaDian:
The mode is stored by YouTube itself, so the native "Playback in feeds" preference keeps showing the old value until the settings screen is recreated, which is why a restart is asked for. The pending mode was removed because it was kept in memory and the restart ends the process before it could ever be applied.

MarcaDian:
Functionality verified on version `21.21.80`; for all other versions, only compilation was verified.

MarcaDian:
@Brocman Please test this artifact

https://github.com/MorpheApp/morphe-patches/actions/runs/30471609893/artifacts/8731994959

Brocman:
I apologize, I'm not a tester — just a regular user. I don't understand what to do with this.

roniui:
> https://github.com/MorpheApp/morphe-patches/actions/runs/30471609893/artifacts/8731994959

Working as intended (21.04.223)

PR #2269, opened 2026-07-30
feat(YouTube Music): Add `Third-party lyrics` patch

### LINKED ISSUE #189: feat(Youtube Music): Handle YouTube Music Lyrics Translation (Unlock or Hide)
### Feature description

Currently, the new Lyrics Translation feature in YouTube Music is visible in the UI, but it's fully locked behind a Premium paywall. When I tap the translate icon, it immediately triggers the "Get Premium" banner instead of translating the text.

I would like to request a patch to handle this in one of two ways:

Priority A (Unlock): Attempt to bypass this restriction. I'm not sure if this is a hard server-side check (using the Google Translate API only for premium accounts) or if it can be tricked by spoofing the client version or parameters.

Priority B (Hide): If unlocking is impossible due to server-side limitations, please add a patch option to remove/hide the translate button entirely from the Lyrics screen. Right now, it just sits there as a "trap" that leads to an ad.

Image
Image

### Motivation

he main goal of using Morphe is to have a seamless, ad-free experience. Having a button in the main UI that serves no function other than triggering a Premium upsell banner goes against the core philosophy of the project.

Explicit use case & Benefits:

UX Improvement: Users won't accidentally click a button expecting a feature, only to be interrupted by a popup.

Cleaner Interface: If the feature can't be used, the UI element shouldn't be there cluttering the lyrics view.

Consistency: It aligns with other patches that hide non-functional or promotional elements (like the "Upgrade" tab or sample buttons).

Basically, if we can't use it, we shouldn't see it.

### Acknowledgements

### LINKED ISSUE #597: feat(YouTube Music): Lyrics romanization and better lyrics
### Feature description

feat: Add japanese lyrics romanization and the better lyrics for the lyrics section if possible pls
For references you can see the ytm clients like metrolist and vivi's lyrics section

<details>

Image

</details>

### Motivation

Me and many others love some language songs than english but can't read the lyrics or sing along so this would be useful for us

### Acknowledgements

MarcaDian:
Adds the `Third-party lyrics` patch, which fills the lyrics panel with lyrics from `LRCLIB` or `KuGou`, falling back to the built-in lyrics when none are found. Synced lyrics follow playback and can be tapped to jump to that moment, and the panel has
its own copy and translate buttons. The lyrics service, text size and timing offset are configurable.

Translation uses the device language and is cached, as are the lyrics themselves.
___


MarcaDian:
The only side effect I am currently observing occurs when the native `Lyrics` button is unavailable because the native source lacks the lyrics, even though they might be available via third-party providers.

dhtzs:
@MarcaDian thanks for the patch! Is there any way the side-effect could somehow be fixed?

MarcaDian:
Since it is a server-side controlled Litho component, I don't know yet. Even the text you see in the original panel is server-side.

MarcaDian:
It's done on my end.

patches/src/main/resources/addresources/values/music/strings.xml
```diff
@@ -41,10 +41,11 @@ Second \"item\" text"</string>
...
     <string name="morphe_settings_music_screen_4_player_title">Player</string>
-    <string name="morphe_settings_music_screen_5_scrobbling_title" translatable="false">Scrobbling</string>
-    <string name="morphe_settings_music_screen_6_return_youtube_dislike_title" translatable="false">Return YouTube Dislike</string>
-    <string name="morphe_settings_music_screen_7_sponsorblock_title" translatable="false">SponsorBlock</string>
-    <string name="morphe_settings_music_screen_8_misc_title">Miscellaneous</string>
+    <string name="morphe_settings_music_screen_5_lyrics_title">Lyrics</string>
+    <string name="morphe_settings_music_screen_6_scrobbling_title" translatable="false">Scrobbling</string>
+    <string name="morphe_settings_music_screen_7_return_youtube_dislike_title" translatable="false">Return YouTube Dislike</string>
+    <string name="morphe_settings_music_screen_8_sponsorblock_title" translatable="false">SponsorBlock</string>
+    <string name="morphe_settings_music_screen_9_misc_title">Miscellaneous</string>
```
LisoUseInAIKyrios, line 48: TODO: Rename these keys on Crowdin before merging

dhtzs:
@MarcaDian @LisoUseInAIKyrios, is there an alternative icon we can use for the 'Copy lyrics' button? The current share icon doesn't accurately reflect the action.

MarcaDian:
I tried to introduce as little external code as possible and reuse all existing resources.

fontzhh:
any updates  about side effect that cannot show lyrics as Lyrics " not available in native player " , but its available in 3rd party sites?

btw it showes the lyrics of 3rd party site if you just switched to next track while opening the lyrics of available track previously

MarcaDian:
There is no solution at the moment. The button is entirely server-controlled. Adding a clunky button to the player that duplicates the existing one is not a viable option.

bys-exe:
Thanks alot man

hyperion002:
Thanks for this feature.
Any future plans of supporting Romanisation of lyrics? Something similar to BetterLyrics extension?

MarcaDian:
Create a separate request to implement this feature if one has not already been created.

jsbsvjend:
Lyrics translation breaks when the app language is set to Portuguese. In English, selecting lyrics shows the custom menu (Copy / Translate) and translates through the external service. In Portuguese, the custom menu fails to load and shows the native buttons (Share / Translate) instead, which triggers the YouTube Music Premium paywall screen upon clicking. The patch only injects the external translation interface when the app is in English or other language. 
Screenshot_20260821-013424
Screenshot_20260821-013320

MarcaDian:
Open separate issue.

PR #2273, opened 2026-07-30
fix: Mirror in RTL locales and restyle to match the manager

MarcaDian:
### Description

Fixes the `About` and `Credits` dialogs rendering left to right in RTL locales, where the row chevrons pointed the wrong way and pre-release version numbers were reordered by the bidi algorithm. Both dialogs now share a single stylesheet that follows the `Morphe` app's look, with Material Symbols and brand marks replacing the hand-drawn SVG icons.
___


MarcaDian:
It looks like it's ready on my end.

PR #2282, opened 2026-07-31
feat(YouTube - Playback speed): Add "Default playback audio pitch" setting

### LINKED ISSUE #1096: feat(YouTube): Pitch Controls
### Feature description

Pitch controls/gesture completely adds new optional when it's enabled and so it allows you to customize pitches like for example  
-12 by default but you can customize reaching -48 pitch or up to +48 pitch its similar ones like pipepipe or newpipe its actually fit so perfectly fine

### Motivation

There could be my favorite feature! You would add this in the next update!! Pls

### Acknowledgements

### LINKED ISSUE #1585: feat(YouTube): Speed changes pitch instead of stretching
### Feature description

Youtube uses an audio-stretching algorithm to preserve pitch when changing speed by repeating or cutting out very small parts of the audio. I think it'd be nice to have an option to turn it off and let the pitch change naturally, like a setting in the [Global Speed](https://github.com/polywock/globalSpeed) browser extension.

### Motivation

The default audio-stretching algorithm heavily distorts audio at low speeds, which is undesirable for music and can sound awkward. I personally like listening to music at different pitches, it makes it sound new again.

### Acknowledgements

ragavpr:
### Description

- Adds support to independently control playback `AudioPitch` from `CustomPlaybackSpeedInterface`.
- Uses `VideoInformation` to store the state, can remember it as well.
- Adds settings and actions equivalent to the video speed value.
- Forces `ExoPlayer`'s `PlaybackParameters` constructor to use audio pitch value from `VideoInformation` extension.
- This currently deviates from YT's default behaviour with `AudioTimeStretching` setting disabled by default, this is the same setting toggled directly in the playback controls with the link toggle button.

EDIT: This only patches ExoPlayer, the previous YT patch is now removed.

Closes #1096
Closes #1585 

### Additional context

It is similar to NewPipe's speed/pitch controls.

### Test results

- [x] Tested on both the **minimum** (20.14.43 on older main branch) and **maximum** (21.04.223 on 1.39.0-dev.9) supported versions
- [X] Tested on **experimental** supported versions (21.31.523 on 1.39.0-dev.9) (Optional)


ragavpr:
Added icons and made it scrollable if there isn't enough vertical space in landscape mode.

LisoUseInAIKyrios:
What does "Audio time stretching" do?

ragavpr:
> What does "Audio time stretching" do?

Audio Time Stretching is the process to change audio pitch independently to the playback speed (tempo), the algorithms to do this aren't perfect and real-time ones usually break the audio (especially in music).

The setting "Audio Time Stretching" when enabled, allows any video speed / audio pitch combinations to be set independently.
If disabled, audio speed is forced to be equal to video speed (since they are equal, ExoPlayer will skip audio time stretching processing for the audio, and the pitch varies naturally without any breaks in the audio).

The default YT behavior is to preserve the pitch (the same at 1x, so voices don't sound too deep / high) when the audio is played back at different speed.

LisoUseInAIKyrios:
So, it's a setting to enable the pitch control UI buttons?

There should be a way to hide the pitch control UI in the speed menu.

ragavpr:
Currently there is no setting to hide / show the pitch controls, it is always shown in the UI.

Instead of adding a toggle in the speed menu itself, will a toggle in the morphe/video settings menu do? (just above the audio time stretching one).

[currently implementing this]

LisoUseInAIKyrios:
Yes it can be in the app settings and not in the dialog.

If pitch controls are enabled, is there any reason to turn off the time stretching? If there isn't a reason that setting could be removed and it's always active if pitch controls are enabled.

ragavpr:
New "Enable Audio Pitch controls" toggle,

audio pitch control -> disabled = any video speed, 1x audio pitch (default YT)
audio pitch control -> enabled = shows the audio pitch section

audio time stretching -> disabled = any video speed can be set but audio pitch is matched (pull req. default) (daycore / nightcore mode)
audio time stretching -> enabled = any video speed and any audio pitch can be set.

Also, it seems you have reverted speed menu defaults to 15 presets, it just makes the menu one row bigger (and makes the link button go out of view in landscape)

LisoUseInAIKyrios:
Should "Audio time stretching" be on all the time? Why would someone want pitch on but audio time stretching off?

> audio time stretching -> disabled = any video speed can be set but audio pitch is matched

How is this different from:
> audio pitch control -> disabled = any video speed, 1x audio pitch (default YT)

If "audio time stretching" is kept it needs a name/summary that makes it more clear what it does.

ragavpr:
Total Playtime = Total Duration / PlaybackRate
Playbackrate can be measured in (fps for video, or sampleRate in Hz for audio)

Let's say, we put 2x video speed in default YT, it sends `[videoPlaybackRate (speed) = 2 and audioPlaybackRate (pitch) = 1 (default)]` to ExoPlayer, since they are not same, the audio **needs to be time-stretched** (some algorithm that chops up audio and plays at default speed and skipping the rest, even though this maintains the final pitch, it definitely breaks the audio).

Now, we enable `Pitch UI controls` with `AudioTimeStretching` off, we put 2x video speed again, the audioPlaybackRate (pitch) is matched (set equally) to the videoPlaybackRate (speed), ExoPlayer gets `[videoPlaybackRate = 2 and audioPlaybackRate = 2]`, since they are equal, **ExoPlayer will skip time-stretching algorithms** and play the audio at 2x playbackRate without any processing, (this will make the audio sound higher pitch but will have no breaks in the audio (similiar to nightcore music videos)), and is the main use case for this PR.

If `AudioTimeStretching` is on, user is given more freedom on setting video and audio playback rates independently, (might be useful for music producers, maybe idk).

I'd recommend to keep it, even though technically `AudioTimeStretching`, `videoPlaybackRate` and `audioPlaybackRate` are independent, for "video/audio" combination "speed/pitch" pair is used to represent them.

I'll add these as description for `AudioTimeStretching` setting,

`AudioTimeStretching` Disabled: "Prevents audio time-stretching by syncing video and audio playback rates.".
`AudioTimeStretching` Enabled: "Allows audio time-stretching (audio artifacts may occur).".

ragavpr:
Pretty much made the code a lot easier to read, along with fixing a bug (allowing the slider to go out of limits) that was introduced a while ago, I think this is finished, will test and mark as ready after a while.

LisoUseInAIKyrios:
In the next day or two I'll try to finish local changes I am working on. 

21.30+ needs a bytecode change and I will try to handle that.

ragavpr:
I'll first try to attempt a patch that won't touch YT at all and only target ExoPlayer, that should work for wider range of versions.

LisoUseInAIKyrios:
I think this is ready. Will merge this tomorrow.

98xe:
https://github.com/MorpheApp/morphe-patches/actions/runs/31683613815

Morphe settings fail to load.


<details>
  <summary>Screenshot</summary>

Screenshot_20260813-103924_YouTube Morphe



</details>

PR #2295, opened 2026-08-01
feat(YouTube): Add "Add-on support" patch for third party patch bundles

MarcaDian:
Adds the runtime hooks that third party add-on patch bundles attach to, since the manager loads every bundle in its own class loader and an add-on cannot reference any patch of this bundle at patch time. An add-on registers itself in `AddOnManager.registerAddOns()`, subscribes to the player button and video hooks of `AddOnApi`, claims one of the reserved legacy player button slots, and declares its preferences in `morphe_addon_prefs.xml`, which the settings patch merges into the Morphe settings.


LandonMoran:
What add-ons would this allow to work

PR #2304, opened 2026-08-02
feat(YouTube - Hide layout components): Add "Hide 'Invite others to message' card" setting

### LINKED ISSUE #2303: feat(YouTube): Hide the Messages section on the Notifications tab
### Feature description

Since around June 2026, YouTube shows a "Messages" section at the top of the Notifications tab, containing an "Invite others to message" card ("Get a link to start a chat and send videos"). It cannot be dismissed or disabled in YouTube's own settings.

Add a switch (Settings → Feed) that hides this section, both the "Messages" header and the invite card, leaving the rest of the Notifications tab (section headers like "Today"/"This week" and the notification rows) untouched.

For reference, the equivalent annoyance is reported in the RVX fork as anddea/revanced-patches#1559, with no fix there either.

YouTube screenshot showing the Messages section at the top of the Notifications tab

### Motivation

- The section is promotional: it advertises YouTube's new DM feature to users who have never used it, permanently occupying the top of the Notifications tab.
- It cannot be dismissed through any official means, so every visit to the tab starts with a screenful of unwanted UI before the actual notifications.
- Morphe already has patches for hiding comparable promotional sections elsewhere in the feed; this fills the same need on the Notifications tab.
- A working implementation exists (I will open a PR referencing this issue), so the cost to the project is review rather than development.

### Acknowledgements

kimsey0:
### Description

Adds a "Hide 'Invite others to message' card" setting to the **Hide layout components** patch (switch under Settings → Feed, enabled by default) that hides the "Invite others to message" card shown in the Messages section at the top of the Notifications tab since YouTube introduced its DM feature in mid-2026. The section is not dismissible through any official YouTube setting.

The card is a top-level Litho cell on the `FEactivity` feed: `linear_layout.eml-fe|76f740a55d9fbbe`, wrapping `connections_inbox_zero_state`. `LayoutComponentsFilter` registers an identifier callback on `linear_layout.e`, restricted to `contentIndex == 0`, and confirms via the language-independent `connections_inbox_zero_state` buffer string. Only then is the selected navigation button checked against `NavigationButton.NOTIFICATIONS`.

Closes #2303

### Additional context

**Per review, the "Messages" shelf header above the card is not hidden.** Every shelf header on the Notifications tab ("Messages", "Notifications", "Today", "This week", "Older") shares the same identifier, the same eml hash, and a byte-identical buffer apart from the title string, which is localized server-side (it is not an app string resource). There is therefore no language-independent way to match the header, and matching server-localized text is not maintainable. Users who want the header gone anyway can hide it with a custom filter at their own risk; I will document that in #2303. A comment in `LayoutComponentsFilter` records this limitation so it is not re-attempted.

Other decisions worth flagging for review:

- **Default ON**, on the grounds that the card is promotional (same reasoning as the ad-shelf defaults). Easy to flip if you disagree.
- Switch placed on `PreferenceScreen.FEED`, matching the other "hide {component}" feed switches.
- The tab check relies on `navigationBarHookPatch`, which `hideLayoutComponentsPatch` already depends on.
- The navigation button is checked last, after `contentIndex` and the buffer check, so that `getSelectedNavigationButton()` is only reached for a component that has already matched. That call waits on the back button latch and can block the litho thread for up to 120 ms, and `linear_layout.e` is generic enough to appear all over the app.

Development was AI-assisted (Claude Code). I have reviewed the code and verified all behavior on-device myself.

### Test results

Tested on a Pixel 8 running Android 17. Additionally tested on an Android 14 x86_64 emulator with a signed-in account, on both ends of the supported range:

- **21.04.223 (max supported)**: the server serves the Messages section. With the toggle ON, the "Invite others to message" card is gone; logcat confirms `LayoutComponentsFilter filtered identifier: linear_layout.eml-fe|76f740a55d9fbbe`. With the toggle OFF the card returns and nothing is filtered. The "Messages" header remains visible in both states, as intended.
- **Back button behavior**: after backing out of nested settings screens with the hardware back button and going straight to the Notifications tab, the card is still filtered correctly, and no litho thread blocked on the nav button latch.
- **20.21.37 (min supported)**: the server does not serve the Messages section to this client at all (its Notifications tab still uses the older All/Mentions layout). The filter is verified inert on that version.
- Notification rows, Home, and Subscriptions are unaffected on both versions, in both toggle states.
- The preference renders correctly on both versions under Morphe settings → Feed, enabled by default, sorted between "Hide image shelf" and "Hide 'Latest videos' button".


LisoUseInAIKyrios:
This could be part of `Hide layout components` patch.

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/MessagesSectionFilter.java
```diff
@@ -0,0 +1,96 @@
...
+                              String path,
+                              byte[] buffer,
+                              BufferAsciiStrings asciiStrings,
+                              StringFilterGroup matchedGroup,
+                              FilterContentType contentType,
+                              int contentIndex) {
+        // Both identifiers are generic and used all over the app,
+        // so only filter the Notifications tab.
+        if (contentIndex != 0
+                || NavigationButton.getSelectedNavigationButton() != NavigationButton.NOTIFICATIONS) {
```
LisoUseInAIKyrios: Checking the navigation button can temporarily block filtering due to workarounds required to fix the back button behavior.

Instead, check the navigation button very last after all buffer checks pass.

kimsey0:
@LisoUseInAIKyrios, merged into the `Hide layout components` patch and fixed the navigation button check.

Do you have any opinion on the header title match? I can add all the localized titles, but I don't know how to avoid depending on the server-side translation string.

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/components/LayoutComponentsFilter.java
```diff
@@ -110,10 +114,44 @@ public LayoutComponentsFilter() {
...
+        // keep any mismatch contained to that tab.
+        messagesSectionHeader = new StringFilterGroup(
+                Settings.HIDE_MESSAGES_SECTION,
+                "shelf_header.e"
+        );
+
+        messagesSectionHeaderTitleBuffer = new ByteArrayFilterGroup(
+                null,
+                "Messages"
+        );
```
LisoUseInAIKyrios: I just noticed this uses localized text.

This isn't ideal because Yt is offered in different languages.

We previously tried using localized text, but the localizations kept changing or we overlooked some langauges. It was not a good solution.

Please look for something else in the buffer that is unique and not localized.

kimsey0: I already tried, and I cannot find anything. See the pull request description.

LisoUseInAIKyrios: > In other app/account languages the section degrades to an orphaned header with no card under it.

This is less than idea. So close yet so far.

kimsey0: Yes, I would also prefer a better solution, but there really doesn't seem to be anything setting the headers apart other than the string.

Personally, I would rather have the card hidden even if the empty header is shown than nothing. The card is very visually annoying.

As mentioned, I'm happy to go through all currently supported languages in the YouTube app and gather translation strings.

LisoUseInAIKyrios: Unless the translation strings are in strings.xml (very unlikely), then it's server side and they can change at anytime.

Previously we tried a patch that used localized server side strings but the strings were slowly changed by YT until most languages didn't work anymore. We want to keep future maintenance at a minimum and using localized strings observed in server side data doesn't give us that.

My only suggestion is maybe this can be hidden another way, but perhaps not.

Can this be a custom filter so users can use this on their own, and if it breaks for them it breaks for them? There is no way to filter based on navigation tab, but path 'contains' can be used along with a single buffer filter expression.

kimsey0: What about following the "filter strings" convention this patch already uses in four places (`morphe_hide_channel_tab_filter_strings`, `morphe_hide_feed_flyout_menu_filter_strings`, `morphe_hide_account_menu_filter_strings`, `morphe_hide_comments_carousel_filter_strings`)?

- The switch keeps hiding the invite card in all languages via the non-localized `connections_inbox_zero_state` buffer string, so no maintenance risk there.
- The header match becomes an empty-by-default `morphe_hide_messages_section_filter_strings` text preference where users enter the section title in their own language, exactly like the comments carousel filter (whose "Ask" / "Hype this video" titles are also server-localized). No localized strings ship with Morphe, and if YouTube changes the title, users update their own entry.

Compared to a plain custom filter, this keeps the Notifications tab scoping and the first-component restriction (which custom filters can't express), while still meeting the "users own the string" requirement.

If you'd rather not have the text preference at all, I'm also fine trimming the PR to card-only hiding and documenting a custom filter for the header in #2303.

LisoUseInAIKyrios: The filter strings solution isn't great either and causes a lot of confusion, but those features are highly desirable we accept the limitation.

Trimming this to just hide the card only, with no localized text filters and no edit texts the user needs, might be ok.

kimsey0: Alright! I cut it down to just hiding the card.

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -78,6 +78,10 @@ Changes include:
     <string name="morphe_hide_hyped_label_title">Hide Hyped label</string>
     <string name="morphe_hide_image_shelf_title">Hide image shelf</string>
     <string name="morphe_hide_image_shelf_summary">Hides image shelves in search results</string>
+    <!-- 'Invite others to message' should be translated using the same localized wording YouTube displays
+          for the card in the Messages section at the top of the Notifications tab. -->
+    <string name="morphe_hide_invite_to_message_card_title">Hide \'Invite others to message\' card</string>
+    <string name="morphe_hide_invite_to_message_card_summary">Hides the \'Invite others to message\' card shown in the Messages section at the top of the Notifications tab</string>
```
```suggestion
    <string name="morphe_hide_invite_to_message_card_summary">Hides the invite card shown in the Notifications tab Messages section</string>
```

iSaiAss13:
> This could be part of `Hide layout components` patch.

Could you guys please also remove the label "Messages"?

kimsey0:
@iSaiAss13, you can hide the header by doing the following:

1. Enable **Morphe settings → General → Custom filter**.
2. Add this line to **Custom filter strings**:
   ```
   ^shelf_header.e$Messages
   ```
   Replace `Messages` with the exact title YouTube shows in your app's language (case-sensitive).
3. Restart YouTube.

iSaiAss13:
> @iSaiAss13, you can hide the header by doing the following:
> 
> 1. Enable **Morphe settings → General → Custom filter**.
> 2. Add this line to **Custom filter strings**:
>    ```
>    ^shelf_header.e$Messages
>    ```
>    Replace `Messages` with the exact title YouTube shows in your app's language (case-sensitive).
> 3. Restart YouTube.

Thank you so much! Worked just fine! ♥️

PR #2314, opened 2026-08-03
chore(YouTube Music - Third-party lyrics): Use the "Show" naming convention for the button settings and fix auto-translation

dhtzs:
### Description

Makes the state of the `Third-party lyrics` button settings explicit in their keys. Setting titles describe the state the toggle turns on, so both keep their "Show ..." wording and stay enabled by default; only the keys and the title phrasing change.

| Before | After |
| --- | --- |
| `morphe_music_lyrics_copy_button` — "Show copy lyrics button" | `morphe_music_lyrics_show_copy_button` — "Show Copy button" |
| `morphe_music_lyrics_translate_button` — "Show translate lyrics button" | `morphe_music_lyrics_show_translate_button` — "Show Translate button" |

Both default to on, so there is no behaviour change.

It also fixes a null dereference on the auto-translate path, reported by Copilot during review. `LYRICS_TRANSLATE` is persisted, so a track change could call `onTranslateClicked()` with the Translate button hidden and `translateView` null. `Objects.requireNonNull` then threw, and because the method catches `Exception` this never crashed — it silently stopped translating while leaving the setting saved as on. The guard now returns before the state is saved, matching how `updateTranslateLabel()` already null-checks the same field.

The setting is deliberately not forced off when the button is hidden, so hiding and later unhiding the button resumes translation as the user left it.

This part is not caused by the rename — the previous `morphe_music_lyrics_translate_button` switch could hide the button too, so the same path was already reachable.

Closes #

### Additional context

The two renamed keys lose their existing translations until the next Crowdin sync. `AddResourcesPatch` skips translated strings whose key is missing from the default locale, so both settings fall back to English in the meantime. The wording is unchanged in meaning, so the translations should carry over cleanly once Crowdin picks up the new keys.

No settings migration is included, as the old keys only ever shipped in the `v1.39.0-dev.2` and `v1.39.0-dev.3` prereleases. The polarity is unchanged, so `migrateOldSettingToNew` would work here if one is wanted.

### Test results

### DIFF
--- a/extensions/music/src/main/java/app/morphe/extension/music/patches/lyrics/ui/LyricsPanelView.java
+++ b/extensions/music/src/main/java/app/morphe/extension/music/patches/lyrics/ui/LyricsPanelView.java
@@ -173,7 +173,7 @@ public final class LyricsPanelView extends FrameLayout implements LyricsManager.
         buttonRow.setGravity(Gravity.CENTER);
         buttonRow.setVisibility(GONE);
 
-        if (Settings.LYRICS_COPY_BUTTON.get()) {
+        if (!Settings.LYRICS_HIDE_COPY_BUTTON.get()) {
             TextView copyView = new TextView(context);
             applyButtonStyle(copyView, COPY_ICON);
             copyView.setText(str("morphe_music_lyrics_copy"));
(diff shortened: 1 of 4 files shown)


LisoUseInAIKyrios:
These buttons have different semantics, because these buttons are not present in the unpatched app. So there is nothing to hide if they are not normally there, hence 'show'.

But I suppose they could be the same for consistency.

ILoveOpenSourceApplications:
Setting titles should reflect the state when toggled on. Renaming it to 'Hide' instead of 'Show' makes the toggle behavior confusing.

MarcaDian:
So, the PR is simply renaming the keys and strings?

dhtzs:
@MarcaDian plus a bug fix.

PR #2332, opened 2026-08-05
feat(YouTube): Add `Fullscreen cutout padding` patch

Cistoran:
### Description

Adds a `Fullscreen cutout padding` patch that moves the fullscreen video away from the camera cutout, so the camera sits in the letterbox bar instead of on top of the picture.

On a foldable or tablet, a 16:9 video in landscape is letterboxed with a large black bar above and below it. On these screens the front camera often falls inside the picture rather than inside that bar. Moving the video into the existing slack costs no video size, and the offset is clamped to the space available, so the video is never resized, cropped or pushed out of view. A screen with no letterbox has no slack, so the patch does nothing on a normal phone screen.

**Which view is moved.** The video surface sits inside wrappers that are sized to the picture rather than the screen, and each of them clips its children. Moving the surface itself only crops it. The patch walks up from the surface while each parent has the same bounds and moves the outermost such wrapper, whose own parent is sized to the screen. Player controls are siblings of that wrapper, so they keep their original layout and nothing is pushed off screen. `setTranslationY` is used rather than a margin change, so there is no re-layout and the surface is never resized.

**Automatic vs manual.** Automatic mode uses the cutout reported by `WindowInsets.getDisplayCutout()`. Devices with an under-display camera report no cutout at all, so a manual distance is used as a fallback in that case, without the user having to change the mode.

For the manual side, the two landscape rotations are always 180 degrees apart, but they are `90`/`270` when the display's natural orientation is portrait and `0`/`180` when it is landscape, as on many tablets and foldable inner screens. Which of the two puts the camera above the video also depends on how the panel is mounted relative to its display (`installOrientation`), so the camera side is selectable, and the side can also be pinned to top or bottom.

Settings are grouped in a sub-screen under Player, and default to off.

### Additional context

Verified against the system's own geometry on a Galaxy Z Fold 8 Ultra. Its inner panel is 2256x2504 and `dumpsys window displays` reports `udcCutout` bounds of `Rect(1665, 22 - 1747, 104)`. In landscape the video is 2504x1408 with 424px bars, and the camera intrudes 167px into the picture, which matches the offset the patch applies at the equivalent setting.

That device also demonstrates the under-display case: `initCutout` and `baseCutout` are both null and `getDisplayCutout()` returns null, even though the window is `layoutInDisplayCutoutMode=always` and the system marks it `mOverlappingWithCutout=true`. There is no public API that exposes the cutout there, which is why the manual fallback exists.

Known limitations:

- The movable-view search assumes the video surface sits inside wrappers sized to the picture, within a parent sized to the screen. If a future layout differs, the search returns the surface itself, the available slack computes to zero and the patch does nothing rather than cropping.
- Only the manual path could be exercised on the test device, since it reports no cutout. The automatic path is implemented and logged but has not been confirmed on hardware that reports a real cutout.
- Only one device was available, so the default mapping of rotation to camera side is verified on that panel alone. Devices mounted the other way are handled by choosing the opposite camera side.

No existing issue to close.

### Test results

Patched and installed on `20.21.37` and `21.04.223`, both applying cleanly with the full patch set. Behaviour checked on a Galaxy Z Fold 8 Ultra, Android 17: the video shifts in both landscape rotations, stays put in portrait, and the offset is stable across repeated rotations and fullscreen enter/exit, with no cropping.

Not tested.


patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -916,6 +916,24 @@ Settings → Playback (or Autoplay) → Autoplay next video"</string>
...
+    <string name="morphe_fullscreen_cutout_padding_mode_title">Cutout detection</string>
+    <string name="morphe_fullscreen_cutout_padding_mode_entry_1">Automatic</string>
+    <string name="morphe_fullscreen_cutout_padding_mode_entry_2">Manual</string>
+    <string name="morphe_fullscreen_cutout_extra_margin_title">Automatic extra clearance</string>
+    <string name="morphe_fullscreen_cutout_extra_margin_summary">Additional space between the camera and the edge of the video</string>
+    <string name="morphe_fullscreen_cutout_manual_side_title">Add space on</string>
+    <string name="morphe_fullscreen_cutout_manual_side_entry_1">Follow rotation, camera in right half</string>
+    <string name="morphe_fullscreen_cutout_manual_side_entry_2">Follow rotation, camera in left half</string>
+    <string name="morphe_fullscreen_cutout_manual_side_entry_3">Always top</string>
+    <string name="morphe_fullscreen_cutout_manual_side_entry_4">Always bottom</string>
```
LisoUseInAIKyrios, line 933: Is this setting applied only when cutout detection is set to manual?

If so an availability could be added to show the setting only when 'manual' cutout detection is selected.

0xrxL:
@Cistoran Any update about this?

MarcaDian:
If this PR has been abandoned, it should probably be closed. Then, if the author reappears, it can be reopened.

PR #2333, opened 2026-08-05
feat(YouTube - Add to queue): Add "Add feed flyout queue menu item" setting

0xrxL:
### Description

Since Google is testing the removal of the stock "Add to queue" button on some accounts, I thought it would be useful to inject new buttons directly into the flyout menu. This will not only restore the "Add to queue" button but could potentially fix other issues—such the lack of a "Save to watch later" button into the flyout menu of some videos (maked as kids)

Screenshot_20260805_111330

### Test results

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/AddToQueuePatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/AddToQueuePatch.java
@@ -53,6 +53,7 @@ public final class AddToQueuePatch {
             initializeNewDivider(flyoutPanel, currentInjectIndex);
         }
     }
+
     @SuppressWarnings({"UnusedReturnValue", "SameParameterValue"})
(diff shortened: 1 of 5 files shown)


0xrxL:
Tested on 21.31.xx for the moment.

EDIT: Now works from 20.21 to 21.xx

MarcaDian:
Awesome 😎

MarcaDian:
This can now also be used to add videos to the AIList.

AmazingKo:
@0xrxL Can you try now to remove the `'P'` label from the original button please? bcz I'd prefer using the original one until Google officially deprecates/removes it.

LisoUseInAIKyrios:
We have a working solution we wrote from scratch and doesn't use anyone else's code, but we are holding off adding it for reasons previously explained.

LisoUseInAIKyrios:
This needs refactoring so other patches can add flyout menu items, but that can be in another PR.

PR #2337, opened 2026-08-05
feat(YouTube - Sponsorblock): Add LRU segment caching for instant video re-entry

gajera1080:
## Description
Adds an in-memory LRU cache for SponsorBlock segments to eliminate redundant network requests and display segment markers instantly when re-opening recently played videos.

## Changes
- Added a **30-entry** `LruCache<String, SponsorSegment[]>` in `SegmentPlaybackController`.
- Loaded cached segments immediately on video re-entry (playlist track changes, history re-opens, or background/PiP resume).
- Reset segment execution flags (`didAutoSkipped`, `recordedAsSkipped`) per session when loading cached segments.
- Evicted LRU cache on SponsorBlock settings re-initialization and channel whitelisting.

## Verification
- Verified build with `./gradlew build`.
- Also tested on `21.04.223`.


LisoUseInAIKyrios:
This is an issue with recently released videos where there may be new segments that previously were not present.

 Refetching again isn't a big deal and I don't see users opening previous videos that often.
 
 I don't think this change is needed.

gajera1080:
Thanks for the review and the explanation. I understand the concern about recently added segments and the tradeoff. I'll close this PR.

PR #2352, opened 2026-08-06
fix(YouTube): Prevent chapter titles from being truncated

### LINKED ISSUE #2037: feat(YouTube): Chapters are being cut off
### Feature description

The episode titles displayed in the player at the bottom, next to the time, are limited to a certain length, even though there's plenty of space to fit the full title.
It could be made so that the full title is displayed if there's enough room for it. In other words, make its length adaptive.

Image

### Motivation

I think that's clear.

### Acknowledgements

ILoveOpenSourceApplications:
### Description

Closes #2037.

### Additional context

N/A

### Test results

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/ChaptersHookPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/ChaptersHookPatch.java
@@ -7,8 +7,6 @@
 
 package app.morphe.extension.youtube.patches;
 
-import android.os.Build;
-
 import androidx.annotation.Nullable;
 
 import java.util.Arrays;
@@ -40,16 +38,10 @@ public class ChaptersHookPatch {
                 if (currentMillis >= marker.patch_getStartMillis()
                         && currentMillis < marker.patch_getEndMillis()) {
                     CharSequence title = marker.patch_getTitle();
-                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
-                        return title != null && !title.isEmpty()
-                                ? title
-                                : null;
-                    } else {
-                        //noinspection SizeReplaceableByIsEmpty
-                        return title != null && title.length() > 0
-                                ? title
-                                : null;
-                    }
+                    //noinspection SizeReplaceableByIsEmpty
+                    return title != null && title.length() > 0
+                            ? title
+                            : null;
                 }
             }
         }
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/ChaptersHookPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/ChaptersHookPatch.java
@@ -7,6 +7,8 @@
 
 package app.morphe.extension.youtube.patches;
 
+import android.os.Build;
+
 import androidx.annotation.Nullable;
 
 import java.util.Arrays;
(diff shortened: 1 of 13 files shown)


extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/ChaptersHookPatch.java
```diff
@@ -38,9 +40,16 @@ public static CharSequence getChapterTitleAtTime(long currentMillis) {
...
-                    return title != null && title.length() > 0
-                            ? title
-                            : null;
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
+                        return title != null && !title.isEmpty()
+                                ? title
+                                : null;
+                    } else {
+                        //noinspection SizeReplaceableByIsEmpty
+                        return title != null && title.length() > 0
```
LisoUseInAIKyrios: I think it's better to use the length() check and ingore the warning than to have two code paths with so little changes.

extensions/youtube/src/main/java/app/morphe/extension/youtube/videoplayer/PlayerOverlayButton.java
```diff
@@ -351,4 +371,38 @@ public static TextView addButtonWithTextOverlay(View sourceButton,
 
         return textOverlay;
     }
+
+    /**
+     * Unconditionally removes YouTube's native maxWidth restrictions from the chapter title.
+     */
+    public static void initializeButton(View controlsViewGroup) {
+        Utils.runOnMainThread(() -> {
```
LisoUseInAIKyrios: If this is already on the main thread then can run the code here.

If it needs to be pushed onto the back of the main thread queue (call runOnMainThread from the main thread), then should add a comment explaining this.

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/ChaptersHookPatch.java
ILoveOpenSourceApplications: I've already made similar changes to other places, which is why I did the same here as well.

LisoUseInAIKyrios: If there's a noinspection to ignore the warning, then can use that.

I wish it recognized `isEmpty()` isn't available in the target android version and `length()` must be used.

ILoveOpenSourceApplications: Me too, but I prefer keeping the explicit API version check so we can natively use `isEmpty()` on API 35+ while safely falling back to `length() > 0` for older versions. It ensures we're utilizing the optimal method for each platform level.

LisoUseInAIKyrios: Using one or the other isn't going to make any difference for performance. I think it's better to keep the code simple and suppress the warning or we just ignore it.

extensions/youtube/src/main/java/app/morphe/extension/youtube/videoplayer/PlayerOverlayButton.java
```diff
@@ -111,6 +122,15 @@ void updateMargin(int buttonWidth, int totalButtons) {
...
 
             if (container.getLayoutParams() instanceof ViewGroup.MarginLayoutParams params) {
+
+                try {
+                    Field maxWidthField = params.getClass().getField("matchConstraintMaxWidth");
+                    if (maxWidthField.getInt(params) != 0) {
+                        maxWidthField.setInt(params, 0);
+                    }
+                } catch (Exception ignored) {
+                }
```
LisoUseInAIKyrios: Does this code work?

I don't see a field `matchConstraintMaxWidth` on `MarginLayoutParams`

ILoveOpenSourceApplications: It works, yes.
<p>

| Before | After |
|--------|--------|
| Screenshot_20260807_152004_YouTube | Screenshot_20260807_152457_YouTube | 

</p>
</details>

LisoUseInAIKyrios: Android 16 the field doesn't exist. This code can be removed.

ILoveOpenSourceApplications:
Does this need to be tested on old and new supported versions or can be merged as is?

LisoUseInAIKyrios:
Check 20.21.37 and if it works then merge

ILoveOpenSourceApplications:
Looks good in `20.21.37`.
<p>

image

</p>
</details>

PR #2416, opened 2026-08-11
feat(Reddit): Disable forced in-app update popup

armdebug:
### Description

Adds `disableAppUpdatePatch` to hide forced in-app update prompts.

### Additional context

N/A

### Test results

- [x] Tested on `v2026.14.0`


LisoUseInAIKyrios:
To avoid confusion, maybe this should be an internal patch that is always applied.

"Disable app updates" or some variation of that name might be confused with Play Store update prompts, which this patch does not change.

Edit: Changed patch to be always applied

patches/src/main/kotlin/app/morphe/patches/reddit/misc/settings/Fingerprints.kt
```diff
@@ -110,3 +110,13 @@ internal object WebBrowserActivityOnCreateFingerprint : Fingerprint(
...
     strings = listOf("com.reddit.extra.initial_url")
 )
+
+internal object GooglePlayUpdateCheckFingerprint : Fingerprint(
+    returnType = "Ljava/lang/Object;",
+    parameters = listOf(
+        "Lkotlin/coroutines/jvm/internal/ContinuationImpl;"
+    ),
+    filters = listOf(
+        string("getAppUpdateInfo(...)")
```
LisoUseInAIKyrios: This string is not present in the latest experimental.

The fingerprint I changed it to is found in all versions of Reddit, and uniquely identifies exactly 1 method.

armdebug: Ok, got it

PR #2436, opened 2026-08-13
fix(YouTube - Captions): Auto captions do not show when video is unmuted

oliver-leung:
## Summary

"Always show" and "Show when unmuted" auto captions settings were not working — captions only appeared when muted, then disappeared when unmuted.

## Cause

`disableAutoCaptions()` returned `original` for the `BOTH_ENABLED` / `WITH_VOLUME_ONLY` cases. This relied on YouTube's subtitle manager natively returning `true` during unmuted playback.

Starting around YouTube 20.26, auto-captioning moved behind a mute-only feature flag (`45692436` / `NoVolumeCaptionsFeatureFlagFingerprint`). The subtitle manager's native decision during unmuted playback became `false`, so `return original` effectively meant "no captions when unmuted."

The muted path was handled correctly (`disableMuteAutoCaptions` overrides the feature flag), but `disableAutoCaptions` was never updated to actively force captions on for unmuted playback.

## Fix

Simplified `disableAutoCaptions` to explicitly force captions on or off during the initial video load window based on the user's setting, then defer to `original` after the 150ms guard to respect manual CC button toggles.

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/AutoCaptionsPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/AutoCaptionsPatch.java
@@ -31,18 +31,16 @@ public class AutoCaptionsPatch {
      * Injection point.
      */
     public static boolean disableAutoCaptions(boolean original) {
-        AutoCaptionsStyle style = Settings.AUTO_CAPTIONS_STYLE.get();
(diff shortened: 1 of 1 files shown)


LisoUseInAIKyrios:
Please check these changes work as expected with YT:
20.21
20.31
21.04

0xrxL:
@oliver-leung Any update about this? It works on old versions?

0xrxL:
@LisoUseInAIKyrios Since the original author didn't answered, I think this could be merged to allow people to test it. It's not a huge change after all.

PR #2446, opened 2026-08-14
perf(YouTube): Cache app version comparison in LithoFilterPatch

gajera1080:
### Description

In `LithoFilterPatch.java`, `is_20_22_or_greater()` is called inside `isFiltered()` on every single Litho component evaluation during feed scrolling and search rendering.

Previously, `is_20_22_or_greater()` dynamically resolved `Utils.getAppVersionName()` and performed a character-by-character `compareTo("20.22.00")` string comparison on every invocation (hundreds of times per second during fast scrolling).

Since the application version is immutable throughout the process lifecycle, this PR caches the comparison once into a `static final boolean IS_20_22_OR_GREATER` at class load time, eliminating redundant string operations in the critical UI rendering path while maintaining full backward compatibility.

---

### Changes
* Added `static final boolean IS_20_22_OR_GREATER` in `LithoFilterPatch.java`.
* Inlined `IS_20_22_OR_GREATER` in `isFiltered()`.
* Removed redundant private `is_20_22_or_greater()` method.

---

### Test results

- [x] Tested on `21.04.223`

### DIFF
--- a/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/patches/components/LithoFilterPatch.java
+++ b/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/patches/components/LithoFilterPatch.java
@@ -82,6 +82,9 @@ public final class LithoFilterPatch {
      */
     private static final int LITHO_LAYOUT_THREAD_POOL_SIZE = 1;
 
+    private static final boolean IS_20_22_OR_GREATER =
+            Utils.getAppVersionName().compareTo("20.22.00") >= 0;
+
     private static final byte[] EMPTY_BYTE_ARRAY = new byte[0];
 
     /**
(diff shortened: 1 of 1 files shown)


LisoUseInAIKyrios:
This change crept in without being noticed.

There already is a class with static constants for checking the app version 

Use this instead

https://github.com/MorpheApp/morphe-patches/blob/47dd5b417b5b79f3361a030f03a3efc2c7682ec6/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/VersionCheckPatch.java#L10

LisoUseInAIKyrios:
Actually, this needs other changes because the class is shared with YT Music.

Fixed with this: https://github.com/MorpheApp/morphe-patches/commit/2c2685c981f3fb87bc57e3fa0c07309cf0403898

Thanks for pointing this out.

gajera1080:
Thanks for fixing it and for the clarification.

PR #2462, opened 2026-08-15
fix(YouTube - Seekbar): Dynamically set round edges for 'most replayed' label background

### LINKED ISSUE #2435: bug(YouTube): Seekbar thumbnail jumps when an element gets added/removed from its details
### Bug description

Notice the abrupt jumping in the seekbar thumbnails.

https://github.com/user-attachments/assets/69932524-6ec3-438a-9e64-7dce4fe23038

I believe the bug occurs whenever the playhead passes by a section of a video that causes a new element (such as chapter names, "Most replayed" indicator, etc.) to be added/removed from the thumbnail's details/description.

In the video demo, for example,

- "Most replayed" sections: The thumbnail jumps whenever scrubbing in or out of a "Most replayed" section, as an element ("Most replayed", in this case) is added or removed.
- End of the YT video: Scrubbing from the end of the video (where there's zero details) back into a chapter causes an element (chapter names, in this case) to be added, triggering the bug.

In both cases, the bug occurs when something is either added or removed the from the thumbnail's details.

### App version patched

21.33.321, 1.40.0-dev.5

### Error logs

```shell

```

### Acknowledgements

0xrxL:


LisoUseInAIKyrios (commented):
Looks good

PR #2470, opened 2026-08-16
fix(Disable Play Store updates): Change manifest version code but preserve existing internal version code checks

LisoUseInAIKyrios:
Add bytecode patch to preserve the original version code of the app for all internal checks.

This should fix strange behavior in some apps such as Facebook, which can show weird black icon overlays on the feed images if disable play store patch is included. Presumably the app is sending the version code with server requests and the response has newer/different data the app doesn't understand.

This change may also improve compatibility with logging in or any other internal app checks, but is not confirmed.

Tested on Reddit but should work without issues on all apps.

### Test results

### DIFF
--- a/extensions/all/change-version-code/build.gradle.kts
+++ b/extensions/all/change-version-code/build.gradle.kts
@@ -1,0 +1,23 @@
+import java.lang.Boolean.TRUE
+
+extension {
+    name = "extensions/all/versioncode/change-version-code.mpe"
+}
+
+android {
+    namespace = "app.morphe.extension"
+    compileSdk = 36
+
+    defaultConfig {
+        minSdk = 26
+    }
+
+    compileOptions {
+        sourceCompatibility = JavaVersion.VERSION_17
+        targetCompatibility = JavaVersion.VERSION_17
+    }
+}
+
+dependencies {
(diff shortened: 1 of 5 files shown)


EpirusTr:
I've been including this patch in my package since version 1.18.0.

I was wondering if it would affect YouTube Morphe.

The assigned version code hasn't changed and it can be installed over the previous one.

(Original version code: 1554771392)

<details>

Screenshot_20260816_204209_com estrongs android pop pro

</details>

LisoUseInAIKyrios:
There is no reason to use `Disable Play Store updates` with YouTube, because patched YouTube already must change the package name which also turns off play store updates.

This pull request does not change anything with the manifest version code, so existing installs will continue to work and can be upgraded as usual if disable play store updates patch is still used.

But I would reinstall YT without the play store patch because the patch is not solving any problem.

EpirusTr:
When I added it for the first time, there was no detailed information. Also, I didn't know much.

I add the patches I made to the 4pda.to website.

It doesn't matter to me, but if I cancel the patch, the original version code will be active and no one there will be able to install it over the previous ones.

I don't want to say delete the previous one and install the one like this.

I'll continue like this for now.

Thank you again, good work...

Issue #2412, opened 2026-08-11
bug(YouTube): "Failed to fetch visitor data" error message


xsoul33:
### Bug description

In some wifi i am having the playback issue for both youtube and ytm i have enen tried different spoofing client and it is still not resolved buy when try to connect to mobile network it resolved it is happening only in wifi network and also showing the below error when try to use the app and also client not found error

<img width="3692" height="1540" alt="Image" src="https://github.com/user-attachments/assets/ba8a5d13-5f09-4500-b2da-737599e932a9" />

### App version patched

21.04.223 of youtube and latest morph version 

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

PrzemekSkw:
Hi, I also have that issue after try Morphee on brand new Motorola. I use One Plus before and everything works fine. 
Now only turn off spoofing help me but then I get info that sponsorblock don't work. 
I try every client in spoofing menu:
- TV, Simply TV, Android VR, Android Studio, VisionOS).
I cannot login when Android VR is choosen because oauth2 don't work. 
My playback issues are only on YT Morphee. YT music works fine. 
All videos start on YT but I have to wait about 40 sec. 
Cannot set resolution when playback.
My issues are:
- dislike is unavailable (blocked unpatched use of SSL stack) - actually I see thumb down but don't see thumb up.
- JavaScriptManager: player JsUrl not found
- VisitorIdRequester: failed to fetch visitor data.
Reinstalling, clearing cache, set permissions don't work.
Regards.

nick-77:
Posting debug logs as asked.

Error shows up as "network error", always happens when client switches to "TV", even with community posts. 

Using Android VR (signed in) client 

[Debug Logs](https://drive.google.com/file/d/1zhcMZwso_JAfi0eChEwE7H-2Y3VhwsV2/view?usp=drivesdk)

LisoUseInAIKyrios:
@nick-77 Please sign out of Android VR and sign-in again.

Also, if visionOS works it can be a better choice than VR.

LisoUseInAIKyrios:
Also, the "connection error" should only show if you have miscellaneous debugging enabled.

Debug is very verbose and can shows a lot of popups to get the attention of who is using the app, but leaving debugging on is not intended for end users.

nick-77:
Yes, that was it.

[...]

PR #2479, opened 2026-08-17

inotia00:
### Exclude `Network proxy` patch by default

There is an issue where `Blocked unpatched use of SSL stack` is printed in the debug logs.

I think this is due to the `Network proxy` patch.

This patch is temporarily excluded by default.

### Reduce the Visitor ID lifespan to 30 days and do not remove it via AuthUtils

Since the Visitor ID is updated by the player response as well, this logic has been removed.

### Override all player configs in `Android VR`

All playerConfigs are now overridden in `Android VR`.

It is necessary to test whether there are any issues when the `Override initial video quality` setting is turned on and off.

### Skip fetching if not signed in to `Android VR`

You can no longer fetch streaming data from `Android VR` while logged out or incognito mode.

To continue using Android VR, please use `Android VR sign in`.

### Add `Android VR Downgraded` client

This is a legacy client where the audio track menu is not available.

If you are experiencing playback issues on `Android VR`, you can try this client.

### Build artifact

https://github.com/MorpheApp/morphe-patches/actions/runs/32224687134/artifacts/9355177939

### Test results


extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/spoof/requests/StreamingDataRequest.java
```diff
@@ -333,11 +333,27 @@ private static StreamData buildPlayerResponseBuffer(ClientType clientType,
...
 
-            return new StreamData(streamingDataBuffer, playerConfigBuffer);
+            if (clientType.requireSABR && playerResponse.hasPlayerConfig()) {
+                PlayerConfig.Builder playerConfigBuilder = playerResponse.getPlayerConfig().toBuilder();
+
+                // In some clients, 'playerGestureConfig' is missing from the response.
+                // Add 'playerGestureConfig' using proto builder.
+                PlayerGestureConfig.Builder playerGestureConfigBuilder = playerConfigBuilder.getPlayerGestureConfig().toBuilder();
+                playerGestureConfigBuilder.setDownAndOutPortraitAllowed(true);
+                playerGestureConfigBuilder.setDownAndOutLandscapeAllowed(true);
```
inotia00, line 353: This can also be fixed via patch:

https://github.com/inotia00/revanced-patches/commit/0b35694d18958f3787128bb88187db661c098dd4

This method was used because using proto builder is simpler.

0xrxL:
I'll review it soon. Anyway you'll add WEB client in this PR too?

0xrxL:
Anyway...no issues here on 21.33

inotia00:
> I'll review it soon. Anyway you'll add WEB client in this PR too?

The `WEB` client is still in WIP.

It requires a login, but for some reason, the login cookie expires every 2 to 6 hours (This means you have to log in again every 2 to 6 hours).

Since I was unable to fix this issue, I have given up on implementing the web client and am working on other tasks.

inotia00:
> Override initial video quality

Is it okay if the `Override initial video quality` setting is removed? (Enabled by default)

0xrxL:
> > Override initial video quality
> 
> Is it okay if the `Override initial video quality` setting is removed? (Enabled by default)

You means to straight remove the flag override, or keep it always to false?

0xrxL:
> > > Override initial video quality
> > 
> > 
> > Is it okay if the `Override initial video quality` setting is removed? (Enabled by default)

You means to straight remove the flag override, or keep it always to false? Because this latter option sounds ok for me.

LisoUseInAIKyrios (commented):
If proxy patch isn't applied for the stream fetch, does proxy patch still work to bypass region restrictions?

Figim:
Will Morphe not have mweb client support? 😔

0xrxL (approved):

0xrxL:
Looks like that the flag override can be left there (always set to off) without the toggle.

patches/src/main/resources/addresources/values/youtube/strings.xml
```diff
@@ -1179,12 +1179,8 @@ Tap here to open DeArrow.ajay.app"</string>
     <!-- video.quality.rememberVideoQualityPatch -->
     <!-- Translations should match the language used in 'morphe_custom_playback_speeds_auto'. -->
     <string name="morphe_video_quality_default_entry_1">Auto</string>
-    <string name="morphe_override_initial_video_quality_title">Override initial video quality</string>
     <!-- 'Force AVC (H.264) codec' should be the same translation used for 'morphe_force_avc_codec_title'. -->
     <!-- 'Spoof video streams' should be the same translation used for 'morphe_spoof_video_streams_screen_title'. -->
```
ILoveOpenSourceApplications:
```suggestion

```

ILoveOpenSourceApplications (commented):
Can remove these as well.

0xrxL:
@inotia00 I built a local source with VR 1.83.11 and works flawlessly on 21.33. Do you think is the case to include it in this PR?

P.S: With this PR, I can reproduce the infinite video loading (on some videos) regardless if platypus flag is overridden or less. So I removed the flag overridden.

0xrxL:
After some testing, I noticed that certain videos — specifically Shorts opened via the video player — experience playback issues when set to the highest supported quality (480p in my case). In fact, as soon as I force the next lower preferred quality setting, playback begins normally. Consequently, I implemented a routine that automatically switches to the next available quality level after few attempts, thereby bypassing the issue of continuous restarting.

0xrxL:
@inotia00 If you have a better solution, use it, because my solution is really ugly.

inotia00:
> If proxy patch isn't applied for the stream fetch, does proxy patch still work to bypass region restrictions?

Since the proxy was temporarily disabled in the `Spoof video streams` patch, it does not bypass regional restrictions.

I temporarily disabled the proxy only in this PR to check if #2412 has been fixed.

If it is confirmed that this issue is caused by the proxy, the proxy patch should be fixed.

inotia00:
PlayerConfig responses vary depending on the client version.

Do not update the YouTube VR client version, as new issues may occur.

LisoUseInAIKyrios:
Since we can't get confirmation that disabling the proxy for spoof streams fixes the ssl warning, should the proxy patch be restored and add a user dialog warning to the proxy setting?

Figim:
"android vr" on ytdlp doesn't work with http 403 error. Does Android br require Potoken? 
Will it affect Morphe too? 

https://github.com/yt-dlp/yt-dlp/issues/17456

inotia00:
> "android vr" on ytdlp doesn't work with http 403 error. Does Android br require Potoken? Will it affect Morphe too?
> 
> [yt-dlp/yt-dlp#17456](https://github.com/yt-dlp/yt-dlp/issues/17456)

Although SABR-related playerConfigs already existed in Android VR 1.65.10, SABR has not been enforced so far.

While I haven't checked in detail, it is highly likely that Android VR 1.65.10 also switched to a SABR-only stream.

Since yt-dlp does not support SABR, a 403 error may occur when attempting to download via dash stream.

Morphe is free from this issue because it supports both SABR and Android VR sign-in.

inotia00:
> Since we can't get confirmation that disabling the proxy for spoof streams fixes the ssl warning, should the proxy patch be restored and add a user dialog warning to the proxy setting?

It seems the issue is reproduced when the proxy patch is included (regardless of whether the user has enabled or disabled it in the settings).

If there is no user feedback, it seems better to exclude proxy patch by default.

inotia00:
> > "android vr" on ytdlp doesn't work with http 403 error. Does Android br require Potoken? Will it affect Morphe too?
> > [yt-dlp/yt-dlp#17456](https://github.com/yt-dlp/yt-dlp/issues/17456)
> 
> Although SABR-related playerConfigs already existed in Android VR 1.65.10, SABR has not been enforced so far.
> 
> While I haven't checked in detail, it is highly likely that Android VR 1.65.10 also switched to a SABR-only stream.
> 
> Since yt-dlp does not support SABR, a 403 error may occur when attempting to download via dash stream.
> 
> Morphe is free from this issue because it supports both SABR and Android VR sign-in.

I looked into it, and it doesn't seem related to SABR.

Instead, it looks like YouTube VR effectively needs to sign in now.

0xrxL:
@inotia00 Android VR is the unique broken client at the moment. Apart from log-in issues, also changing the quality resolution cause seeking issues...which can be reproduced from Liso too. For this reason Liso wondered if wouldn't be the case to downgrade the client to 1.6x version, to solve these issues.

inotia00:
> @inotia00 Android VR is the unique broken client at the moment. Apart from log-in issues, also changing the quality resolution cause seeking issues...which can be reproduced from Liso too. For this reason Liso wondered if wouldn't be the case to downgrade the client to 1.6x version, to solve these issues.

Yeah, I asked you for a review to test this.

What about the dev branch (before the PR was merged) rather than the PR branch? Does Android VR still work?

0xrxL:
> What about the dev branch (before the PR was merged) rather than the PR branch? Does Android VR still work?

Also the dev branch is affected by the same issue. But I noticed again yesterday. Probably one of your last changes related to player params (in the previous PR), forcely enabled a problematic flag that restart the playback when the quality is somehow changed.

inotia00:
Restoring Android VR 1.65 is not that bad.

In Android VR 1.65, audio tracks are not available in the flyout menu.

Since the force original audio track patch will not work in Android VR 1.65, the deleted code must be restored.

Figim:
I think the client's current version is better. It also needs to stay up to date.

0xrxL:
> I think the client's current version is better. It also needs to stay up to date.

Same... But this flag should be disabled, otherwise sooner or later the behaviour that I reproduced will be extended to other users.

LisoUseInAIKyrios:
> > Since we can't get confirmation that disabling the proxy for spoof streams fixes the ssl warning, should the proxy patch be restored and add a user dialog warning to the proxy setting?
> 
> It seems the issue is reproduced when the proxy patch is included (regardless of whether the user has enabled or disabled it in the settings).
> 
> If there is no user feedback, it seems better to exclude proxy patch by default.

I just noticed you said it fails when the patch is included, which is odd because it seems the patch doesn't do anything when it's included but not enabled.

Disabling the patch for now makes sense, but surely it can be fixed to actually do nothing when included but not enabled.

A user dialog message could be added or the patch description updated to mention the patch may not work with some devices

LisoUseInAIKyrios:
> Restoring Android VR 1.65 is not that bad.
> 
> In Android VR 1.65, audio tracks are not available in the flyout menu.
> 
> Since the force original audio track patch will not work in Android VR 1.65, the deleted code must be restored.

VR 1.65 with no force original audio is ok. Another spoof client is always good even if it's limited.

inotia00:
`Android VR Downgraded` has been added.

Anyone can test it via the build artifact:

https://github.com/MorpheApp/morphe-patches/actions/runs/32224687134/artifacts/9355177939

I have requested a review again because minor fixes, such as user dialog, are needed.

0xrxL:
Looks like the problem is extended to Android TV too (another SABR client, what a coincidence).


https://github.com/user-attachments/assets/45d792cb-d756-4a8a-99ce-ab832de571e4

inotia00:
There are no changes regarding TV SABR in this PR.

Please check if the issue is reproduced on the dev branch.

EpirusTr:
I tested it for version 20.21.37:

- I first exited VR, as it's now mandatory.

visionOS was activated in videos while Android VR and Android VR Downgraded were selected.

- Then I logged back into VR. Both clients worked correctly.

- The Network Proxy patch is already incompatible with version 20.21.37.


<details>


1Screenshot_20260819_113600_app morphe android youtube


2Screenshot_20260819_111924_app morphe android youtube


3Screenshot_20260819_112052_app morphe android youtube


4Screenshot_20260819_111206_app morphe manager

</details>

LisoUseInAIKyrios (commented):
We can merge this now.

Any TV issues can be fixed in another PR.

BlackGold8282:
Would it be possible to change it so that VisionOS is used instead of TV when VR is not logged in?

LisoUseInAIKyrios:
VisionOS may stop working or may not work for some regions/users.

Tv client is the safer first fallback client.

naijun0403:
As a side note about `Network proxy`, I wonder if the SSL warning comes from `Requester` falling back to `URL.openConnection()`, rather than from proxying itself. Even when the proxy is disabled, that can still reach YouTube's guarded Java SSL stack. This may be worth checking separately.

0xrxL:
> There are no changes regarding TV SABR in this PR.
> 
> Please check if the issue is reproduced on the dev branch.

I meant that I noticed that the same problem is also shared with the TV client (which is part of the SABR clients). And yes, as I wrote in a previous post, I can also reproduce it in the dev channel.

PR #2488, opened 2026-08-17
feat(YouTube): Add "Back button exits the feed" setting

### LINKED ISSUE #1088: bug(YouTube): Pressing back exits the app instead of returning to previous scroll position

ILoveOpenSourceApplications:
### Description

Closes #1088.

### Additional context

N/A

### Test results

### DIFF
--- a/patches/src/main/kotlin/app/morphe/patches/youtube/misc/fix/backtoexitgesture/FixBackToExitGesturePatch.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/youtube/misc/fix/backtoexitgesture/FixBackToExitGesturePatch.kt
@@ -14,13 +14,10 @@ import app.morphe.patcher.extensions.InstructionExtensions.addInstruction
 import app.morphe.patcher.patch.bytecodePatch
 import app.morphe.patches.youtube.misc.extension.sharedExtensionPatch
 import app.morphe.patches.youtube.misc.playertype.playerTypeHookPatch
-import app.morphe.patches.youtube.misc.playservice.is_20_40_or_greater
 import app.morphe.patches.youtube.misc.playservice.versionCheckPatch
-import app.morphe.patches.youtube.shared.YouTubeMainActivityOnBackPressedFingerprint
 import app.morphe.util.addInstructionsAtControlFlowLabel
 import app.morphe.util.getReference
 import app.morphe.util.indexOfFirstInstructionOrThrow
-import app.morphe.util.insertLiteralOverride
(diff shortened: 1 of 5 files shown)


ILoveOpenSourceApplications:
@sr-shishir, you were the one who had the issue with back swipe not exiting the app right? Can you test this PR and let me know if you can exit without any issues or not?

Edit: https://github.com/MorpheApp/morphe-patches/actions/runs/32080473225/artifacts/9304985385

sr-shishir:
Perfectly working. 👌
Tested on 
- 21.04.223 ✔
- 21.33.321 ✔

patches/src/main/kotlin/app/morphe/patches/youtube/misc/fix/backtoexitgesture/FixBackToExitGesturePatch.kt
```diff
@@ -44,44 +41,33 @@ internal val fixBackToExitGesturePatch = bytecodePatch(
...
-                0,
-                "invoke-static { }, $EXTENSION_CLASS->onBackPressed()V"
-            )
+                    const/4 v0, 1
+                    return v0
+
+                    :cond_continue_native
+                """.trimIndent()
+
+            it.method.addInstruction(0, smali)
```
LisoUseInAIKyrios: This needs to use `addInstructionsWithLabel` and `:cond_continue_native` needs a `nop` after it.

The code here effectively makes no changes, and the built-in back button behavior is used.

But, it all seems to work if the 'scroll to top' feature flag is selectively overridden and the exiting patch remains.

MarcaDian:
You overwrote Liso's commit

ILoveOpenSourceApplications:
I just saw that. Fixing it on my end.

sr-shishir:
Would it make sense to add a toggle for this? For example, with the toggle disabled, a single tap on the back button would immediately exit the app. If someone prefers to use the back button to scroll all the way to the top of the feed instead, they could simply enable the toggle. Wouldn't having an option like this be more flexible for users?

LisoUseInAIKyrios:
The latest change has a setting in Miscellaneous

ILoveOpenSourceApplications:
I guess this is now ready to be merged?

tanvir8282:
Pressing back button always exiting the app whether the new settings is enabled or not....but it needs 3 take to exit while disabling the feature.....

PR #2494, opened 2026-08-18
chore: Code cleanup

0xrxL:


extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/LoadVideoPatch.java
```diff
@@ -135,7 +135,8 @@ public static void openIntent(String url, boolean closeCurrentPlayerInstance) {
     }
 
     // This method opens a video based on hardcoded parameters found in an obfuscated class.
-    public static void openVideoWithInternalIntent(String videoIDWithParams) {
+    // It's generally more fast of the solution that use 'Intent.FLAG_ACTIVITY_NEW_TASK'.
```
er13, line 138: @0xrxL 
You probably meant to say:
```
It is generally faster than the solution that uses `Intent.FLAG_ACTIVITY_NEW_TASK`.
```

0xrxL: Fixed in dev 👍

er13: @0xrxL 
You missed one letter (it is, however, not worth a separate commit):
`... than the solution that uses ...`

0xrxL: > @0xrxL You missed one letter (it is, however, not worth a separate commit): `... than the solution that uses ...`

Fixed again and squashed in a single commit.

PR #2510, opened 2026-08-19
fix(Network proxy): Avoid platform HTTPS fallback

naijun0403:
### Description

Routes extension HTTPS requests through usable native Cronet engines when the `Network proxy` patch is included, including when the proxy setting is disabled.

Previously, `Requester` could fall back to `URL.openConnection()`, which can reach YouTube's guarded Java SSL stack.

This change:

- Registers native Cronet engines as they are built and prioritizes the main engine.
- Rejects Java fallback Cronet and prevents platform HTTPS fallback.
- Keeps Java proxy handling restricted to plaintext HTTP.
- Routes AiSList requests through `Requester`.
- Ensures Cronet POST output streams work consistently on YouTube and YouTube Music.

Related to #2412

### Additional context

This change follows up on a possible cause suggested in #2479: `Requester` falling back to `URL.openConnection()` even when the proxy is disabled.

The original warning was not reliably reproducible, so there is no affected-device before-and-after confirmation yet. For that reason, `Network proxy` remains excluded by default until the fix is confirmed on affected devices.

### Test results

### DIFF
--- a/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/patches/NetworkProxyPatch.java
+++ b/extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/patches/NetworkProxyPatch.java
@@ -24,7 +24,6 @@ import org.chromium.net.Proxy;
 import org.chromium.net.ProxyOptions;
 
 import java.io.IOException;
-import java.lang.ref.WeakReference;
 import java.net.HttpURLConnection;
 import java.net.InetSocketAddress;
 import java.net.URL;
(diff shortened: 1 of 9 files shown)


LisoUseInAIKyrios:
This all looks good. Will merge tomorrow morning if there's no new concerns.

roniaxe:
I suspect this patch causes YouTube to stop working. That's what I see every time I open YouTube now after using the latest patch.

Screenshot_20260824-080413

LisoUseInAIKyrios:
https://morphe.software/faq#troubleshooting-5

Issue #1838, opened 2026-06-25
bug(YouTube): Notification icons become monochrome if custom branding is applied


archi-X-beliderscene:
### Bug description

This a duplicate following #1812, but I demonstrated the behaviour improperly, here's the correct version

Notifications of patched YouTube with custom branding
<img width="1200" height="623" alt="Image" src="https://github.com/user-attachments/assets/526cc723-0ca7-47b8-8f65-978e3e5714ff" />

Here's normal, colored notifications
<img width="1200" height="1180" alt="Image" src="https://github.com/user-attachments/assets/f487653e-739e-4805-83a9-83a68d775c0f" />
(And with standard YouTube they also have been colored red-white logo, but I disabled currently after setting up morphe) 
 
### App version patched

21.24.360 experimental

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

rajarshikhatua:
similar bug, app switches to custom app icon with latest pre-release patches, and even after switching to original icon manually it sends notification with custom icon style, even though it's set to original icon.

[Somebody on Reddit too](https://www.reddit.com/r/MorpheApp/s/5GsDB8sIIW)

<img width="1080" height="616" alt="Image" src="https://github.com/user-attachments/assets/7ee9ade2-1c8f-4045-8a61-cd3a68770f58"/>

MarcaDian:
Re-patch using the "Original" option within the "Custom branding" patch. We have added a notification that explains this when selecting an icon from the list.

PR #2518, opened 2026-08-20
feat(Custom branding): Add an app icon patch option and support mounted installations

### LINKED ISSUE #2512: bug(YouTube Music): Wrong icon in Android Auto
### Bug description

I've noticed for quads on wild that YouTube music morphe is showing the wrong app icon in Android auto 

Image

I've been using morphe for quite a while and this bug has been persistent for every version I've ever tried 

### App version patched

9.33.52 

### Error logs

```shell

```

### Acknowledgements

### LINKED ISSUE #2751: bug(YouTube): Icon doesn't change in lawnchair legacy launcher
### Bug description

The icon for YouTube morphe (unrooted) doesn't change in lawnchair legacy launcher and stay as default youtube icon but changes in device's recent menu

Image
Image

### App version patched

21.07.247

### Error logs

```shell

```

### Acknowledgements

MarcaDian:
### Description

Adds an `App icon` patch option, and makes `Custom branding` work for mounted (root) installations.

**App icon**

The icon can now be picked while patching: `App default`, `Original`, `Black`, `Dark`, `Light`, `Play`, `Play black` or `Custom`. A regular installation starts with that icon and can still change it later in the app settings. Patching again with a different icon replaces what was selected in the app, and patching with the same icon leaves the selection alone, so a normal app update does not undo it.

The chosen icon is now also used where the icon selected in the app settings cannot reach, such as Android Auto and the app list of the system settings. Those kept showing the unpatched icon.

**Mounted (root) installations**

Branding did nothing at all for a mounted install, because the system keeps reading the manifest of the stock app. The icon and the app name are applied while patching instead, and the notification icon follows the chosen style as well.

A mounted installation still cannot change its branding from the app settings, and those preferences are hidden there, so changing it means patching again. `Custom branding` also stays off by default for mounted installations.

**Also fixed**

- Launchers that prefer an app's round icon kept showing the unpatched icon.
- The `Original` icon gave YouTube's pre 2024 logo instead of the icon the app actually uses.

**Note**

A launcher caches app icons and has no way to notice that a mounted app was changed, not even after a reboot. So the app asks the launcher to reload its icon the first time it is opened after patching. Until the app is opened once, the launcher keeps showing the previous icon.


ILoveOpenSourceApplications:
Can #1838 be addressed in this PR as well?

MarcaDian:
<s>Probably yes.</s> Actually - no.

LisoUseInAIKyrios:
Root branding doesn't work with Google ASOP launcher, and likely other launchers too.

It shows original YT icon except in weird places such as dragging the icon around the launcher.

I believe the launcher is storing a copy of the app icon and continues to use the original icon when the app was first seen. Since mount shows the app has not changed, the launcher fails to refresh the icon.

LisoUseInAIKyrios:
Now root icon works 🥳

jay:
This PR may have changed things so that when Custom Branding for icon is set to default then the apk icon is the Morphe icon instead of the original application icon like it used to be. I changed the setting to original to get the original application icon back for the apk. I don't know if that was intended or not so I'm mentioning it.

Screenshot_20260913-172058 Files by Google

Akash-Sriram:
its greate patch why it cant be a universal patch the issue mentioned here is for all root mounted app issue

Issue #200, opened 2026-01-13
feat(YouTube Music): playlist/album/artist better background color transition


Altarentys:
### Feature description

By default, when on a playlist/artist/album page, the image is blurred in the background with a gradient going to true black at the bottom, which leads to an ugly transition when app theme is anything other than true black.

First picture below is with true black theme (YTM default), the transition is smooth.
Second picture is with Material You theme, transition isn't smooth.

I don't really know what can be done about this but since it's not truly a bug I made it a feature request.

Maybe putting a "reverse gradient" going from true black to whatever the theme color is, under the transition.

1st
<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/cae850b8-a0f4-426a-b269-4aa95b1ed93a" />

2nd
<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/350c9f4d-26ba-4d91-9200-46e250b6f29d" />

### Motivation

Would be better looking

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] This is a feature request for YouTube or YT Music

PR #2524, opened 2026-08-20
feat(Theme): Change the app background color from the app settings

### LINKED ISSUE #784: feat: Ability to change theme in the patched YouTube app
### Feature description

An ability to change the theme on the patched YouTube app, which would be great. 

### Motivation

I think this feature should be implemented because people need to be able to change themes without having to repatch YouTube every time we want to change the theme. It would be very convenient. 

### Acknowledgements

MarcaDian:
### Description

Both the dark and the light background are now settings inside the app and can be changed at any time.

The app does not paint its background itself, it is resolved by the resource system from XML, so the patch writes one color variant per background and the extension asks for one of them by overriding the configuration of every context the app attaches. The qualifiers it uses are ignored by the app and affect nothing else.

- New settings `Dark theme background` and `Light theme background` under General, with the same choices the patch options had, apart from the one that was identical to the app default. Changing one asks to restart the app. YT Music has the dark background only.
- Every background shows its color next to the name, the same way the app icons are shown. The color is read from the variant that entry would apply, so a preview can never disagree with the result, and `App default` shows the color of the app version being patched.
- `Custom` lets you pick any color. On Android 14 and later the color is exact, because the app registers an overlay for itself, and the patch declares the background colors in `res/values/overlayable.xml` since the system rejects an overlay the target does not allow. On Android 8 to 13 nothing can be added to the resources at runtime, so the closest of 512 colors that were compiled in is used instead.
- Material You backgrounds fall back to the app default on Android 11 and older.
- The new content indicator, the dot of a tab and the count next to it, follows a Material You background. It is now done with a hook that colors the view, instead of replacing drawables, styles and layouts while patching, and it works in both apps.

Depends on:
- https://github.com/MorpheApp/morphe-patches-library/pull/47
- https://github.com/MorpheApp/morphe-patcher/pull/180

___

### DIFF
--- a/extensions/music/src/main/java/app/morphe/extension/music/patches/lyrics/ui/LyricsPanelView.java
+++ b/extensions/music/src/main/java/app/morphe/extension/music/patches/lyrics/ui/LyricsPanelView.java
@@ -50,6 +50,7 @@ import app.morphe.extension.shared.Logger;
 import app.morphe.extension.shared.ResourceType;
 import app.morphe.extension.shared.ResourceUtils;
 import app.morphe.extension.shared.Utils;
+import app.morphe.extension.shared.theme.ThemeUtils;
 import app.morphe.extension.shared.ui.Dim;
 import app.morphe.extension.shared.ui.ViewAnimations;
 
@@ -641,9 +642,9 @@ public final class LyricsPanelView extends FrameLayout implements LyricsManager.
     private static int lineTextColor() {
         final int colorId = ResourceUtils.getIdentifier(ResourceType.COLOR, APP_PRIMARY_TEXT_COLOR);
         if (colorId == 0) {
-            return Utils.getAppForegroundColor();
+            return ThemeUtils.getAppForegroundColor();
         }
-        return ResourceUtils.getColor(APP_PRIMARY_TEXT_COLOR, Utils.getAppForegroundColor());
+        return ResourceUtils.getColor(APP_PRIMARY_TEXT_COLOR, ThemeUtils.getAppForegroundColor());
     }
 
     private static String sourceText(String providerName) {
--- a/extensions/music/src/main/java/app/morphe/extension/music/patches/theme/ThemePatch.java
+++ b/extensions/music/src/main/java/app/morphe/extension/music/patches/theme/ThemePatch.java
@@ -1,5 +1,7 @@
 package app.morphe.extension.music.patches.theme;
 
+import androidx.annotation.ColorInt;
+
 import app.morphe.extension.shared.theme.BaseThemePatch;
 
 @SuppressWarnings("unused")
@@ -21,7 +23,8 @@ public class ThemePatch extends BaseThemePatch {
      * @param originalValue The original color value.
      * @return The new or original color value.
      */
-    public static int getValue(int originalValue) {
+    @ColorInt
+    public static int getValue(@ColorInt int originalValue) {
         return processColorValue(originalValue, DARK_VALUES, null);
     }
 }
--- a/extensions/music/src/main/java/app/morphe/extension/music/patches/theme/ThemePatch.java
+++ b/extensions/music/src/main/java/app/morphe/extension/music/patches/theme/ThemePatch.java
@@ -1,3 +1,13 @@
+/*
+ * Copyright 2026 Morphe.
+ * https://github.com/MorpheApp/morphe-patches/pull/2524
+ *
+ * Original hard forked code:
+ * https://github.com/ReVanced/revanced-patches/commit/724e6d61b2ecd868c1a9a37d465a688e83a74799
+ *
+ * See the included NOTICE file for GPLv3 Section 7 terms that apply to Morphe contributions.
+ */
+
 package app.morphe.extension.music.patches.theme;
 
 import androidx.annotation.ColorInt;
--- a/extensions/music/src/main/java/app/morphe/extension/music/settings/MusicActivityHook.java
+++ b/extensions/music/src/main/java/app/morphe/extension/music/settings/MusicActivityHook.java
@@ -1,3 +1,13 @@
+/*
+ * Copyright 2026 Morphe.
+ * https://github.com/MorpheApp/morphe-patches
+ *
+ * Original hard forked code:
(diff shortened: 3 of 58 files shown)


MarcaDian:
The only regression at the moment is the inability to change the notification dot for Material variants; however, since modifying that subtle dot would have required over 100 lines of code, and risked breaking at any time, I don't consider it an issue. I won't be spending time restoring it.

sr-shishir:
Wow it's really an unique feature. Would it be possible to display a small color preview next to each color name? Similar to how the app icon options are shown, where we can visually see which icon we're selecting. I think having a color preview alongside each name would make it much easier to identify and choose the right color.

EpirusTr:
Since the theme setting is no longer available, it's no longer possible to set our desired color combination before patching.

Do you plan to add the setting back?

LisoUseInAIKyrios:
The custom setting is the app settings

sr-shishir:
Is it possible to keep this portion colorful even after playing a video? Currently, the color changes when I swipe down.

https://github.com/user-attachments/assets/0752146a-0ab1-4f2d-9d0b-ee32fbafa01c


EDIT: The dark green color isn't being applied below the player until I swipe down, while the light green color works as expected.

MarcaDian:
It looks like the new hook for the colored dot also works.

MarcaDian:
@sr-shishir Turn off Ambient mode.

sr-shishir:
Tried on YouTube Music but when going to select theme then this popup shows every time.
<details>
Screenshot_2026-08-20-21-11-26-953_app srs android apps youtube music

</details>

Another thing every time it shows light green color at launch, while it should follow the theme color like pink, red etc. 👇
<details>

https://github.com/user-attachments/assets/9cd48b87-e530-441a-96a0-eccb1b166301

ILoveOpenSourceApplications:
Isn't "App default" and "Modern YouTube" the same color?

MarcaDian:
Yeah, this duplicate can now be deleted.

LisoUseInAIKyrios:
@sr-shishir Please post the logs after the error toast is shown, I cannot reproduce.

sr-shishir:
> Please post the logs after the error toast is shown, I cannot reproduce.

Error toast fixed after repatching but opening theme still stuck at light green color.

MarcaDian:
The green screen issue should be fixed in the latest changes.

sr-shishir:
> The green screen issue should be fixed in the latest changes.

Yup it fixed the issue. Now it's following the theme which I choose. 👌

MarcaDian:
The colored notification counter hook now works for YT Music as well.

sr-shishir:
Is this ok? Or the color will blend like gradient. Can't understand. 
IMG_20260820_235444

MarcaDian:
It's old bug https://github.com/MorpheApp/morphe-patches/issues/200

sr-shishir:
> It's old bug

It's difficult to notice when the background is black, but the actual bug becomes clearly visible as soon as I choose any other background color. 😖

98xe:
Some UI element colors are broken when using `App default`.

<details>
<summary>Example 1</summary>
 


  
</details>

<details>
<summary>Example 2</summary>


  
</details>

MarcaDian:
> Some UI element colors are broken when using `App default`.

It should be fixed in the latest commit.

MarcaDian:
Still need fix Application context issue.

ILoveOpenSourceApplications:
<details><summary>It could be an A/b test, looks like a new color needs to be added to be modified by theme patch.</summary>
<p>

| With theme | Without theme |
|--------|--------|
| Screenshot_20260821_190736_YouTube | Screenshot_20260821_191451_YouTube | 



</p>
</details> 
<details><summary>Morphe and YouTube settings are of different color when theme patch is excluded. </summary>
<p>

| Morphe | YouTube |
|--------|--------|
| Screenshot_20260821_193201_YouTube | Screenshot_20260821_193154_YouTube |

</p>
</details> 

<details><summary>Menu that shows up when using "Sort by" button in playlist tab of a channel page doesn't follow the theme.</summary>
<p>

Screenshot_20260821_191528_YouTube

</p>
</details> 

None of these issues are associated with this PR.

sr-shishir:
What do you think about adding a Opacity Slider to the Custom Background Theme section?

Rat-Man-1015:
Why is it not possible to keep the background pure black while having everything else, such as text, icons, cards, buttons, and other UI elements, use the Material You dynamic theming system? Why is there no option for a pure-black Material You theme?

It would be especially useful on OLED/AMOLED displays where a true black background can turn off individual pixels and reduce power consumption while still retaining the Material You color system and dynamic theming for the rest of the interface.

MarcaDian:
Could you show what you think an OLED material should look like? I don't even understand how that would work on YouTube, which doesn't use standard material elements.

EpirusTr:
Android 10 Problem (!?)

Two days ago, I downloaded the first file and the patch was successful.

https://github.com/MorpheApp/morphe-patches/actions/runs/32362193439

Since last night, I've been trying both developer and PR files. The patch process stops halfway through at the "Writing resource APK" stage.


https://github.com/user-attachments/assets/8045fc59-6d6f-430a-a772-84d827198cc5



- The "Debug" file is empty because it doesn't give an error message.

[morphe_logcat-2026-08-22-142135.log.txt](https://github.com/user-attachments/files/31333198/morphe_logcat-2026-08-22-142135.log.txt)


- Memory settings cannot be configured in Morphe Manager because it's Android 10.

- I tried it on a phone with Android 8, the result was the same.

- I canceled the patch and tried again. The process was successful.

- 20.21.37 and 21.04.223

- I've been patching for months, this is the first time I've failed :)

Addition:

Morphe Manager: 1.26.1 and 1.27.0

TidalMist:
Very good update but i want use this colors -e 'Theme' -O'Dark theme background color'=@color/m3_ref_palette_dynamic_neutral_variant6 -O'Light theme background color'=@color/m3_ref_palette_dynamic_neutral_variant98"

MarcaDian:
Open the separate request. This PR is merged.

Mehedi5585:
> It looks like the new hook for the colored dot also works.
> 
> 

How ..

PR #2553, opened 2026-08-22
fix(YouTube - Add to queue): Match the flyout menu items to the app's own items

MarcaDian:
### Description

Makes the "Add to queue" flyout items match the rest of the menu.

 - "Add to Morphe queue" is built from the app's own menu item layout, so its spacing, font and icon match the native items, and the whole row highlights on touch.
 - The separator below it is now a thin subtle line across the full menu width.
 - "Play next in queue" uses the same style as the other menu items

### Test results

___


0xrxL:
@MarcaDian increase the lateral padding of divider. From the screenshot it takes all the the panel width.

MarcaDian:
It was designed that way.

Example

0xrxL:
> It was designed that way.
> 
> Example
> 
> 

Oh...ok. Forget of it. 👍

PR #2556, opened 2026-08-22
feat(YouTube Music): Add `Play song version in albums` patch

### LINKED ISSUE #514: feat(YouTube Music): Force album song instead of music video
### Feature description

In albums instead of the audio version the video will play instead sometimes for non premium users. Is there a way to fix this as a patch?

### Motivation

I think this was a feature in RVX Manager but it didn't work very well. It would be really nice because the music videos sound different usually.

### Acknowledgements

MarcaDian:
### Description

Some album tracks in YouTube Music play as music videos instead of the normal song. This adds a setting that plays the song version of those tracks.
The album queue stays untouched, so the next track still comes from the album.
The first track of an album takes about half a second longer to start while the album is looked up. Every track after that starts instantly.

Notes:
- Requires "Spoof video streams" to be turned on.
- The track title and artwork still come from the music video.
- While this is on, the built-in "Don't play music videos" setting is ignored, because it only mutes the video instead of playing the song.

### Test results

### DIFF
--- a/extensions/music/src/main/java/app/morphe/extension/music/patches/album/AlbumMusicVideoPatch.java
+++ b/extensions/music/src/main/java/app/morphe/extension/music/patches/album/AlbumMusicVideoPatch.java
@@ -1,0 +1,151 @@
+/*
+ * Copyright 2026 Morphe.
+ * https://github.com/MorpheApp/morphe-patches/pull/2556
+ *
+ * See the included NOTICE file for GPLv3 Section 7 terms that apply to this code.
+ */
+
+package app.morphe.extension.music.patches.album;
+
+import androidx.annotation.GuardedBy;
+import androidx.annotation.NonNull;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import app.morphe.extension.music.settings.Settings;
+import app.morphe.extension.shared.Logger;
+import app.morphe.extension.shared.spoof.requests.StreamingDataRequest;
+
+/**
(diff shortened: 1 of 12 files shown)


justcharlie520-cmd:
Please how do I talk to any of you? I really have something to say and I don't know who to talk to or make friends with, please if you could reply I'll be waiting please.

MarcaDian:
New artifact
https://github.com/MorpheApp/morphe-patches/actions/runs/32713602741/artifacts/9515127350

gototheskinny:
The patch works well, but it causes some side effects. Instead of using the album/single metadata (YouTube Music version), the app recognizes the video title. This messes up Last.fm scrobbles and gives the app an unpolished feel. The 'Now Playing' title should prioritize official track metadata over video titles.

LisoUseInAIKyrios:
Scrobbling already has some logic to try and fix incorrect data. More of the same might be done. 

Side effects are expected with this feature, and this already fixes numerous issues with the feature from it's original form in inotia00 RVX.

gototheskinny:
> Scrobbling already has some logic to try and fix incorrect data. More of the same might be done. 
> 
> Side effects are expected with this feature, and this already fixes numerous issues with the feature from it's original form in inotia00 RVX.

Honestly you guys are doing the gods work and I'm super appreciative of you guy's work. This feature puts YouTube Music mods on the map after the Spotify apocalypse. We really needed this feature ty so much again.

PR #2582, opened 2026-08-25
feat(YouTube): Add `Disable scrolling speed limit` patch

### LINKED ISSUE #94: feat(YouTube): Disable scroll speed throttling
### Feature description

Implement a patch that disables YouTube's scroll speed limitations.

### Motivation

In the more recent versions on YouTube, scrolling through feeds has been deliberately throttled.

This has been mentioned here in https://github.com/MorpheApp/morphe-patches/issues/50 and numerous times on Reddit.

https://www.reddit.com/r/youtube/comments/1ojqh1s/youtube_is_limiting_scroll_speed/
https://www.reddit.com/r/youtube/comments/1p1rxau/the_android_youtube_app_is_now_throttling_how/
https://www.reddit.com/r/GalaxyS25/comments/1nveu1a/youtube_app_slow_scrolling/
https://www.reddit.com/r/youtube/comments/1ohk33a/scroll_speed_on_mobile/

I've done a bit of quick testing, here's what I've found:

- In the Morphe recommended version of YouTube (`20.37.48`), both scrolling down and up is slow.
- In the current version of YouTube as of writing this (`20.51.39`), scrolling down is slow, but scrolling back up is normal.
- The latest Morphe supported version of YouTube that does not have any restrictions (in both scroll directions) is `20.31.42`.
- The latest version of YouTube that does not have any restrictions (in both scroll directions) is `20.34.39`.

I've been using `20.34.39` with patches `v1.1.0-dev.2` for a few hours, and it seems to be working fine so far. But while using older versions works as a temporary workaround, it would be preferable to remove this limitation for current and future YouTube versions, via a patch.

### Acknowledgements

GalaxyXYZ888:
### Description

YouTube's feed RecyclerView (SnappyRecyclerView) overrides fling(int, int) and clamps velocity to a server-configured value on affected versions. Since fling distance scales with input velocity, this cuts scroll/fling distance way down regardless of touch strength or frame timing.

The clamp is gated by a boolean field, set from an unrelated constructor that reads a server-delivered config proto field. It's not part of the client-side feature flag/experiment system and isn't tied to a specific app version, which is why disabling flags or spoofing the app version never fixed it for anyone on the linked issue.

This patch finds the gating field generically, by type, on SnappyRecyclerView. That class keeps its real, unobfuscated name (required for LayoutInflater to resolve it from XML), so it can be targeted directly instead of by per-build obfuscated names. Every write to that field gets replaced with a no-op, so it stays at its constructor default of false and the clamp branch in fling() never engages, no matter what the server sends.

Older supported versions predate this clamp and have no such field on SnappyRecyclerView. The field lookup is optional, so the patch just no-ops silently there instead of failing.

Closes #94

### Additional context

Found the root cause by diffing decompiled sources between 20.34.39 (last unaffected version) and 20.37.48 (first affected version), combined with live on-device logcat instrumentation from a temporary debug patch logging raw fling velocity. That's what surfaced SnappyRecyclerView in a stack trace.

Two other things got investigated and ruled out as the main cause:

A pair of double-valued feature flags that turned out to gate CSI/Streamz telemetry sampling, not scroll physics. Real, but only a minor contributing factor.
RecyclerView's own cached max-fling-velocity field (this.H), which is also wrong on affected versions, but downstream of and masked by the SnappyRecyclerView clamp fixed here.

### Test results

Verified on-device across 20.21.37 (minimum supported, no such field, no-op, no crash), 20.34.39 (unaffected baseline, unchanged behavior), 20.37.48 (first affected version, clamp confirmed and fixed), 21.04.223 (maximum non-experimental supported version, fix confirmed), and 21.34.243 (maximum experimental version, fix confirmed).

Had help from Claude tracking down the root cause and writing this patch.


LisoUseInAIKyrios:
Thanks for the help.

PR #2591, opened 2026-08-25
fix(YouTube - Theme): Some elements keeps the stock color after leaving picture in picture

MarcaDian:


0xrxL:
That force push be with you...

PR #2628, opened 2026-08-27
feat(YouTube): Add `Disable playlist autoplay` patch

### LINKED ISSUE #805: feat(YouTube): Disable autoplay in a playlist
### Feature description

Disable auto play in a playlist

### Motivation

It's inconvenient when you want to watch a single video in a playlist or want to read comments when the video is ending while auto play in a playlist is on.

### Acknowledgements

GalaxyXYZ888:
### Description

Adds an opt-in "Disable playlist autoplay" setting (Player screen) that stops a playlist from automatically advancing to the next video. This is separate from YouTube's own "Autoplay next video" setting, which only governs suggested-video autoplay outside of a playlist context.

Closes #805

### Additional context

N/A

### Test results

Had help from Claude writing this patch.

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/DisablePlaylistAutoplayPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/DisablePlaylistAutoplayPatch.java
@@ -7,9 +7,6 @@
 
 package app.morphe.extension.youtube.patches;
 
-import java.lang.reflect.Field;
-
-import app.morphe.extension.shared.Logger;
 import app.morphe.extension.youtube.settings.Settings;
 
 @SuppressWarnings("unused")
@@ -18,24 +15,7 @@ public class DisablePlaylistAutoplayPatch {
     /**
      * Injection point.
      */
-    public static boolean shouldSkipPlaylistAutoplay(Object navigationIntent) {
-        if (!Settings.DISABLE_PLAYLIST_AUTOPLAY.get()) {
-            return false;
-        }
-        try {
-            for (Field field : navigationIntent.getClass().getDeclaredFields()) {
-                if (!Enum.class.isAssignableFrom(field.getType())) {
-                    continue;
-                }
-                field.setAccessible(true);
-                Object value = field.get(navigationIntent);
-                if (value instanceof Enum<?> && "AUTOPLAY".equals(((Enum<?>) value).name())) {
-                    return true;
-                }
-            }
-        } catch (Exception ex) {
-            Logger.printException(() -> "shouldSkipPlaylistAutoplay failure", ex);
-        }
-        return false;
+    public static boolean shouldSkipPlaylistAutoplay(Enum<?> navigationIntent) {
+        return Settings.DISABLE_PLAYLIST_AUTOPLAY.get() && "AUTOPLAY".equals(navigationIntent.name());
     }
 }
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/DisablePlaylistAutoplayPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/DisablePlaylistAutoplayPatch.java
@@ -1,0 +1,34 @@
+package app.morphe.extension.youtube.patches;
+
+import java.lang.reflect.Field;
+
+import app.morphe.extension.shared.Logger;
+import app.morphe.extension.youtube.settings.Settings;
+
+@SuppressWarnings("unused")
+public class DisablePlaylistAutoplayPatch {
+
+    /**
+     * Injection point.
+     */
+    public static boolean shouldSkipPlaylistAutoplay(Object navigationIntent) {
+        if (!Settings.DISABLE_PLAYLIST_AUTOPLAY.get()) {
+            return false;
+        }
+        try {
+            for (Field field : navigationIntent.getClass().getDeclaredFields()) {
+                if (!Enum.class.isAssignableFrom(field.getType())) {
+                    continue;
+                }
+                field.setAccessible(true);
+                Object value = field.get(navigationIntent);
+                if (value instanceof Enum<?> && "AUTOPLAY".equals(((Enum<?>) value).name())) {
+                    return true;
+                }
+            }
+        } catch (Exception ex) {
+            Logger.printException(() -> "shouldSkipPlaylistAutoplay failure", ex);
+        }
+        return false;
+    }
+}
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/DisablePlaylistAutoplayPatch.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/DisablePlaylistAutoplayPatch.java
@@ -1,3 +1,10 @@
+/*
+ * Copyright 2026 Morphe.
+ * https://github.com/MorpheApp/morphe-patches/pull/2628
+ *
+ * See the included NOTICE file for GPLv3 Section 7 terms that apply to this code.
+ */
+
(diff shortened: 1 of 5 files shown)


GalaxyXYZ888:
Also, i was thinking, YouTube already shows an autoplay toggle overlay during playback for regular videos, but it never appears when watching a video from a playlist. Since that spot is always empty in a playlist context, could we add a toggle styled and positioned the same way that hooks into this patch's setting? It would only ever show where the native one is absent, so there shouldn't be any conflict between the two.

That makes sense to me and would be really useful, but let me know if you think it might be confusing.

I assume this would need to be a separate patch, right?

MondayNitro:
this was already present in rvx as an overlay button, though i never used the feature

patches/src/main/kotlin/app/morphe/patches/youtube/layout/playlistautoplay/DisablePlaylistAutoplayPatch.kt
```diff
@@ -0,0 +1,96 @@
...
+            if (wrapperType != null) return@classDefForEach
+            for (method in classDef.methods) {
+                val params = method.parameterTypes
+                if (method.name == "<init>" && params.size == 3 && params.firstOrNull() == enumType) {
+                    wrapperType = classDef.type
+                    return@classDefForEach
+                }
+            }
+        }
+        val resolvedWrapperType = wrapperType ?: return@execute
```
LisoUseInAIKyrios: This can be replaced with a Fingerprint

patches/src/main/kotlin/app/morphe/patches/youtube/layout/playlistautoplay/DisablePlaylistAutoplayPatch.kt
```diff
@@ -0,0 +1,96 @@
...
+                val params = method.parameterTypes
+                if (params.size == 1 && params.firstOrNull() == resolvedWrapperType) {
+                    if (method.returnType == "V") vMethod = method
+                    if (method.returnType == "I") iMethod = method
+                }
+            }
+            if (vMethod != null && iMethod != null) {
+                navigateMethods += vMethod
+            }
+        }
```
LisoUseInAIKyrios: This can probably be replaced with a fingerprint and use `matchAll().forEach {`

patches/src/main/kotlin/app/morphe/patches/youtube/layout/playlistautoplay/Fingerprints.kt
```diff
@@ -0,0 +1,30 @@
...
+        string("NEXT"),
+        string("PREVIOUS"),
+        string("AUTOPLAY"),
+        string("AUTONAV"),
+        string("JUMP"),
+        opcode(Opcode.RETURN_VOID),
+    ),
+    custom = { _, classDef ->
+        classDef.methods.any { it.name == "<init>" && it.parameterTypes.size > 2 }
+    },
```
LisoUseInAIKyrios, line 28: Is this custom check needed? The ordered strings in the filters may be enough to unique identify the method.

extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/DisablePlaylistAutoplayPatch.java
```diff
@@ -0,0 +1,34 @@
...
+            for (Field field : navigationIntent.getClass().getDeclaredFields()) {
+                if (!Enum.class.isAssignableFrom(field.getType())) {
+                    continue;
+                }
+                field.setAccessible(true);
+                Object value = field.get(navigationIntent);
+                if (value instanceof Enum<?> && "AUTOPLAY".equals(((Enum<?>) value).name())) {
+                    return true;
+                }
+            }
```
LisoUseInAIKyrios: Ideally the patch code finds the field with a fingerprint or manual searching, then passes in the field to this method so the extension code doesn't need to use reflections.

patches/src/main/kotlin/app/morphe/patches/youtube/layout/playlistautoplay/Fingerprints.kt
GalaxyXYZ888: Yeah i tested without it first, it's needed. There are two different enums with the exact same constant names (NEXT/PREVIOUS/AUTOPLAY/AUTONAV/JUMP), so the string filter alone matches both and grabs the wrong one.

LisoUseInAIKyrios (commented):
Much improved, thanks for the help.

LisoUseInAIKyrios:
> Also, i was thinking, YouTube already shows an autoplay toggle overlay during playback for regular videos, but it never appears when watching a video from a playlist. Since that spot is always empty in a playlist context, could we add a toggle styled and positioned the same way that hooks into this patch's setting? It would only ever show where the native one is absent, so there shouldn't be any conflict between the two.

If the button can be restored without complicated patching, then perhaps.

But where is this button at? Is it a player overlay button? We are always a little reluctant to add new player buttons because of UI space.

GalaxyXYZ888:
> > Also, i was thinking, YouTube already shows an autoplay toggle overlay during playback for regular videos, but it never appears when watching a video from a playlist. Since that spot is always empty in a playlist context, could we add a toggle styled and positioned the same way that hooks into this patch's setting? It would only ever show where the native one is absent, so there shouldn't be any conflict between the two.
> 
> If the button can be restored without complicated patching, then perhaps.
> 
> But where is this button at? Is it a player overlay button? We are always a little reluctant to add new player buttons because of UI space.

Screenshot_20260829_103640_YouTube Morphe

I meant this button. It doesn't appear when watching a video in a playlist. I was proposing to add it back in that same UI spot, but have it control this patch's new setting instead of native autoplay.

The thing is, the same-looking button would mean different things depending on context (toggling default autoplay outside a playlist vs toggling playlist autoplay inside one), which could confuse users.

Since it'd be opt-in anyway, maybe a different icon would help tell the two apart. 

This would be useful for me at least, but idk if anyone cares xd.

LisoUseInAIKyrios:
The upper player buttons are a bit messy and fragile as it needs to support both old and new versions of YT, and it has a lot of old legacy code mixed in.

Make a new issue for a feature request and if there's enough interest then we can consider adding it.

MarcaDian:
If we drop support 20.21, I think there won't be any more issues with older versions. Sooner or later, we have to do it.

LisoUseInAIKyrios:
20.21.37 is needed because it's the last useful version for Android 8.0

We should continue to support it until YouTube drops support for it (maybe sometime next year).

AmazingKo:
> > > Also, i was thinking, YouTube already shows an autoplay toggle overlay during playback for regular videos, but it never appears when watching a video from a playlist. Since that spot is always empty in a playlist context, could we add a toggle styled and positioned the same way that hooks into this patch's setting? It would only ever show where the native one is absent, so there shouldn't be any conflict between the two.
> > 
> > 
> > If the button can be restored without complicated patching, then perhaps.
> > But where is this button at? Is it a player overlay button? We are always a little reluctant to add new player buttons because of UI space.
> 
> Screenshot_20260829_103640_YouTube Morphe
> I meant this button. It doesn't appear when watching a video in a playlist. I was proposing to add it back in that same UI spot, but have it control this patch's new setting instead of native autoplay.
> 
> The thing is, the same-looking button would mean different things depending on context (toggling default autoplay outside a playlist vs toggling playlist autoplay inside one), which could confuse users.
> 
> Since it'd be opt-in anyway, maybe a different icon would help tell the two apart.
> 
> This would be useful for me at least, but idk if anyone cares xd.

YES! Please consider adding an overlay toggle for this. having to open settings every single time is super tedious

PR #2638, opened 2026-08-28
feat(Enable debugging - Feature flags manager): Redesign and add flag forcing, import and export, and binary search

MarcaDian:
### Description

Reworks the feature flags manager into a single scrollable list and adds the tooling needed to track down which flag causes a behavior.

- Redesigned layout: one search box, filter chips (`Active` / `Not active` / `Blocked` / `Forced`), compact rows with a state badge, and a bottom bar that switches between global actions and state actions once flags are selected
- Active flags can be disabled as before, but not active flags can be forced on. Forcing on certain flags can cause the app to crash or become unstable, so be sure you have a recent backup of your YouTube Morphe settings.
- `Export` and `import`: the current flags can be shared as text, copied to the clipboard, or written to and read from a file
- Binary search: blocks half of the candidates, then asks after each restart whether the behavior is still present, until the responsible flag is found
- Known flags are now remembered across restarts, so the list is complete right after launch and flags seen on another device can be imported

The file picker plumbing was made generic (`exportTextActivity` / `importTextActivity`), so the settings export, the debug log export and the new flags export now share one implementation instead of three.

<details>
 <summary>Screenshots</summary>

screenshot

Screenshot_2

Screenshot_3

</details>


0xrxL:
This will be very useful to fix issues caused instead by disabled flags.

ILoveOpenSourceApplications:
@MarcaDian, anything more to add to this PR?

MarcaDian:
It's done from my side.

LisoUseInAIKyrios:
It needs a tweak to allow forcing on flags that are off.

Changing that now.

Edit: done

PR #2663, opened 2026-08-29
fix(YouTube - PlayerOverlayButton): Correct spacing when fullscreen button is hidden

### LINKED ISSUE #2657: bug(YouTube): Player overlay button spacing remains unchanged when Fullscreen button is hidden
### Bug description

When the Fullscreen button is hidden from the YouTube player, the remaining five player buttons do not automatically redistribute their spacing. They are too close to each other.

Image

Interestingly hiding other buttons such as Mute or Video Scaling, correctly causes the remaining five buttons to automatically adjust their spacing. 

### Expected behaviour
The same auto-spacing behavior should apply when the Fullscreen button is hidden. The remaining player controls should automatically redistribute themselves across the available space, just as they do when other buttons are hidden.

### App version patched

21.28.208, 21.33.321

### Error logs

```shell
N/A
```

### Acknowledgements

ILoveOpenSourceApplications:
### Description

Previously, if a user hid the native Fullscreen button, the remaining custom buttons (like Quality, Speed, Copy, etc.) remained tightly packed together instead of redistributing their spacing to fill the newly available room.

This happened because the layout math calculated the spacing multiplier based strictly on the size of the buttonControllers array. The formula was designed to compress spacing based on how many extra custom slots were added. However, when the Fullscreen button is hidden, the first custom button simply takes its native slot and shouldn't count toward that spacing penalty.

Closes #2657.

### Additional context

I corrected the mathematical logic for the button layout:

- Replaced the raw array size with an effectiveCustomButtons calculation. If the Fullscreen button is hidden, we subtract 1 from the custom button count, since one custom button absorbs the vacant native slot without increasing the row's density.
- Passed this unified effectiveCustomButtons value into both the X offset calculation and MarginAdjustableContainer.updateMargin().

Now, if a user hides the Fullscreen button, the remaining controls accurately evaluate the true density of the row and redistribute their spacing evenly.

### Test results

### DIFF
--- a/extensions/youtube/src/main/java/app/morphe/extension/youtube/videoplayer/PlayerOverlayButton.java
+++ b/extensions/youtube/src/main/java/app/morphe/extension/youtube/videoplayer/PlayerOverlayButton.java
@@ -97,11 +97,11 @@ public class PlayerOverlayButton {
         }
 
         /**
-         * Adjusts the container's end margin to reserve space for {@code totalButtons}
-         * overlay buttons of the same width as {@code sourceButton}.
+         * Adjusts the container's end margin to reserve space for {@code extraButtonSlots}
+         * overlay buttons using the globally calculated {@code widthPercentage}.
          * Skips the layout pass when the computed value hasn't changed.
          */
-        void updateMargin(int buttonWidth, int totalButtons) {
+        void updateMargin(int buttonWidth, int extraButtonSlots, float widthPercentage) {
             View container = containerRef.get();
             if (container == null) return;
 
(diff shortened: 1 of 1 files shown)


0xrxL:
About player buttons: looks like there's a new cast button on 21.35, that enables companion features on TV. Maybe you'll have more luck to find (and hid) it, because I didn't found anything.

ILoveOpenSourceApplications:
Haven't been part of A/b tests in sometime, but I'll keep an eye out.

PR #2666, opened 2026-08-29
feat(YouTube Music - Scrobbling): Add "Fill in missing album" setting

Nai64:
When the music does not include an album, look it up on Last.fm so your scrobble is complete. Enable Fill in missing album in Music > Scrobbling. If Last.fm cannot find it, scrobble continues as usual without an album.


LisoUseInAIKyrios:
@Nai64 Please try with the latest changes. I don't use last fm so I cannot easily check.

LisoUseInAIKyrios:
@Nai64 Any update?

This could be merged as-is and fixed later if needed.

Nai64:
Sorry! I haven't been receiving mails/notifications whatsoever, thus I forgot about this lolz

But right now I've tested it, and the feature works perfectly.

PR #2689, opened 2026-08-30
refactor: Clean up UI code

ILoveOpenSourceApplications:
### Description

UI Cleanups:
- Moved the search bar and rounded background code into `CustomDialog` to reduce copy-pasted code across multiple files.
- Cleaned up some repeated code in `AbstractPreferenceFragment` and `FeatureFlagsManagerPreference`.
- Applied the same rounded background from the feature flags menu to the debug logs to make them visually consistent.

### Test results


extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/patches/EnableDebuggingPatch.java
```diff
@@ -87,9 +87,15 @@ public static boolean isBooleanFeatureFlagEnabled(boolean value, long flag) {
...
-                Logger.printDebug(() -> "boolean feature is enabled: " + flag);
+
+            if (value) {
+                // If true, force the map to update. Log only if it wasn't true before.
+                if (!Boolean.TRUE.equals(featureFlags.put(flagObj, Boolean.TRUE))) {
+                    Logger.printDebug(() -> "boolean feature is enabled: " + flag);
+                }
+            } else {
+                // Record as false only if we haven't already seen it as true.
+                featureFlags.putIfAbsent(flagObj, Boolean.FALSE);
```
LisoUseInAIKyrios: As far as we know, flags do not change while the app is running and they are the same the entire app session.

I don't think this particular change is needed.

extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/patches/EnableDebuggingPatch.java
```diff
@@ -100,15 +106,20 @@ public static boolean isBooleanFeatureFlagEnabled(boolean value, long flag) {
...
-                        + " value: " + value + (defaultValue == 0 ? "" : " default: " + defaultValue));
+
+            if (defaultValue != value) {
+                if (!Boolean.TRUE.equals(featureFlags.put(flagObj, Boolean.TRUE))) {
+                    // Align the log outputs to make post-processing easier.
+                    Logger.printDebug(() -> " double feature is enabled: " + flag
+                            + " value: " + value + (defaultValue == 0 ? "" : " default: " + defaultValue));
+                }
+            } else {
+                featureFlags.putIfAbsent(flagObj, Boolean.FALSE);
```
LisoUseInAIKyrios: There is no way to force on non boolean flags because there is no way for the user to enter a different non default value.

This change isn't needed. Same with the other non boolean methods below.

LisoUseInAIKyrios (commented):
Other changes look ok.

extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/patches/EnableDebuggingPatch.java
ILoveOpenSourceApplications: Ah, that makes perfect sense. I'll revert the changes to `EnableDebuggingPatch.java` and just push the UI/DRY refactoring for the dialog components. Thanks for catching that and explaining it to me!

PR #2714, opened 2026-08-31
fix(YouTube - Hide ads): Prevent button swap when Navigation bar patch is excluded

ILoveOpenSourceApplications:
### Description

The previous fix in PR #2659 introduced a flaw where `AdsFilter` checked the navigation swap setting to fix BROWSE endpoint bleed-over. Because the setting defaults to `true`, users who excluded the Navigation Bar patch were forced into the Automotive UI with no settings menu to disable it.

This fix uses the patcher's `setExtensionIsPatchIncluded` API to dynamically evaluate if the Navigation patch was applied. `AdsFilter` now uses this check to safely revert the OS to "Android" for the `/guide` endpoint only when the Navigation Bar patch is completely excluded.

### Test results


extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/spoof/SpoofOSNamePatch.java
```diff
@@ -14,6 +14,7 @@
 import app.morphe.extension.shared.Utils;
 
 public class SpoofOSNamePatch {
+    public static boolean isNavBarPatchApplied = false;
```
LisoUseInAIKyrios: Instead use setExtensionIsPatchIncluded(EXTENSION_CLASS) like Reddit patches do.

LisoUseInAIKyrios (commented):
Looks good.

PR #2753, opened 2026-09-03
feat(YouTube): Add `Remember livestream playback position` patch

### LINKED ISSUE #2592: feat(YouTube): Preserve livestream playback position
### Feature description

Add an option to remember the user's playback position for ongoing YouTube livestreams.

There are two different situations this feature could handle:

1. Watching a livestream with a delay

If the user is watching the playback of an ongoing livestream rather than watching it live, It should remember their playback position and resume from there when they reopen the livestream.

Eg. if the livestream is currently at 2:30:00 but the user is watching at 1:10:00 and leaves the Livestream, reopening it should resume at 1:10:00 instead of jumping to the current live position.

2. Watching the livestream live

If the user was watching a livestream live and then leaves or pauses it, reopening the livestream should give them 2 options:

- Resume from last position (continue from where they stopped watching the Livestream)
Or go to live (jump to the current live position)

### Motivation

It is very inconvenient to manually find the position where you stopped watching an ongoing livestream, especially with long livestreams, as there is no indication of where you left off.

This feature would make watching the playback of ongoing livestreams much easier by automatically remembering the last playback position.

### Acknowledgements

GuysLetsPlay:
### Description

This patch adds functionality to preserve the watch position when watching a livestream, either during playback or live.

The default behavior is that when someone watches a livestream behind the live position, e.g. 1 hour behind, then closes the livestream and comes back to it later, the player starts playing from the current live position.

This patch adds two options to the settings:

Remember livestream playback position: When enabled, the position where the user was watching a livestream in playback is preserved.

Resume even when watched live: When this option is enabled, the watch position is also saved when the user is watching a livestream at the live position. For example, if the user closes the livestream and comes back 40 minutes later, pressing the livestream with this option enabled will continue playback from where they left off, rather than jumping to the current live position.

### Additional context

Is related to #2592 

### Test results

- [X] Tested on version v1.41.0-dev.3 of the morphe patches


GuysLetsPlay:
Oh, I forgot to mention that this whole feature does not work with 24/7 livestreams because they don't have a growing duration, which is how the video is detected as a livestream.

LisoUseInAIKyrios:
If 24/7 livestreams don't work, can the second "resume even if watching the head" be removed and turned always on?

So if the setting is on, then non 24/7 livestreams always resume where the user was watching.

It's easy to fast forward to the live head and it makes the settings a lot simpler.

GuysLetsPlay:
I added the second option mainly to make the behavior customizable.

LisoUseInAIKyrios:
It comes at a cost of making the settings more complicated and increasing the chance of user error if they don't understand how the second setting works.

I think a single setting is best. If the user has this setting on and wants to resume a non 24/7 livestream at the play head they can easily fast forward to the live seek time.

LisoUseInAIKyrios:
Please double check everything works as expected. I checked and it all works great for me.

GuysLetsPlay:
Yes it works as expected I tested all features

PR #2786, opened 2026-09-04
feat: Declare a category for every patch

MarcaDian:
## Summary

Adds an optional, bundle-declared **category** to patches and uses it to section the patch lists in the manager. A bundle picks its own taxonomy; the manager groups by whatever it finds and leaves everything else exactly as it is today.

Closes MorpheApp/morphe-manager#929
Closes MorpheApp/morphe-manager#933

Large bundles are hard to navigate: 145 patches in Morphe Patches, and third party bundles reach several hundred. Both issues ask for the same thing: collapsible sections by patch type, with search reaching inside them.

## What changes

**`morphe-patcher`**
- `Patch.category: String?`, added as a trailing field with a default, exactly the way `availability` was added.
- `PatchBuilder.category(name)` for the DSL.
- `api/morphe-patcher.api` regenerated.

**`morphe-patches-library`**
- `PatchListGenerator` writes `category` into `patches-list.json`. Purely additive: existing consumers ignore the new field.

**`morphe-patches`**
- Every named patch declares a category. 129 declarations, which is all 145 patches the manager loads (the difference is shared factories in `shared/` that build one patch per app).

**`morphe-manager`**
- `PatchInfo.category`, read straight off the loaded patch.
- `PatchesListShared.kt` is generalized from the fixed pair "app specific + universal" into a list of blocks: `PatchGroup`, `buildPatchGroups()`, `patchGroupRows()`. The universal block is now simply the last group, still folded by default.
- Settings → Appearance → **Patch list → Group by category** (on by default).

## Backwards compatibility

- **Existing bundles keep working untouched.** Bundles call the `bytecodePatch { }` DSL,never the `Patch` constructor, so a bundle compiled before this change simply never calls `category(...)` and the field stays `null`.
- **A bundle that declares no categories renders exactly the list it renders today**: one headerless block plus the universal tail. This is not a special case in the code, it is what `buildPatchGroups` produces when every category is `null`.
- **No data migration.** Selections, options, `SeenPatch` and exported configurations are keyed by patch name via `uniqueNames()`. `category` is deliberately not part of that key, so every stored selection stays valid.
- **Users who do not want it** switch the setting off, which collapses the categories back into the plain list.

## Declaring a category

One line in the patch builder. Patches that do not call it stay ungrouped.

```kotlin
val hideAdsPatch = bytecodePatch(
    name = "Hide ads",
    description = "Adds options to hide general ads, Premium promotions and video ads."
) {
    category("Ads")

    dependsOn(
        hideAdsResourcePatch,
        settingsPatch
    )

    compatibleWith(COMPATIBILITY_YOUTUBE)

    execute {
        // ...
    }
}
```

The name is free-form, matched by the manager as an opaque key, so a third party bundle can ship any taxonomy it likes. For a patch built by a shared factory, the call goes in the factory and covers every app at once:

```kotlin
internal fun disableDRCAudioPatch(
    block: BytecodePatchBuilder.() -> Unit,
    preferenceScreen: BasePreferenceScreen.Screen,
    useLegacyNormalizationFlag: BytecodePatchBuilder.() -> Boolean,
    useNormalizationFlag: BytecodePatchBuilder.() -> Boolean
) = bytecodePatch(
    name = "Disable DRC audio",
    description = "Adds an option to disable DRC (Dynamic Range Compression) audio."
) {
    category("Video and audio")

    block()

    // ...
}
```

## How it looks in the manager

- Each category is a collapsible header carrying its patch count, and a badge with thenumber of enabled patches so a folded block never hides a selection silently.
- Categories start expanded; the universal block keeps its current folded default.
- Searching or filtering forces every block open and drops the empty ones, so a match isnever hidden behind a fold.
- Ordering: uncategorized patches first (no header), then categories alphabetically, thenuniversal patches last.
- All three patch lists share the same code: the expert mode selection dialog, the per app patch list, and the bundle browser in Patch sources.
<details>
 <summary>Screenshot</summary>

Screenshot_2026-09-05-00-14-55-060_app morphe manager debug-edit

</details>

## Merge order

The repositories depend on each other, so merge and release in this order:

1. https://github.com/MorpheApp/morphe-patcher/pull/198
2. https://github.com/MorpheApp/morphe-patches-library/pull/54
3. https://github.com/MorpheApp/morphe-manager/pull/941
4. https://github.com/MorpheApp/morphe-patches/pull/2786

### DIFF
--- a/patches/src/main/kotlin/app/morphe/patches/all/misc/clone/CloneAppPatch.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/all/misc/clone/CloneAppPatch.kt
@@ -125,6 +125,8 @@ val cloneAppPatch = resourcePatch(
             "may cause app crashes or other unexpected behavior.",
     default = false
 ) {
+    category("Miscellaneous")
+
     packageNameOption = stringOption(
         key = "packageName",
         default = "Default",
--- a/patches/src/main/kotlin/app/morphe/patches/all/misc/installer/ChangeInstallerSource.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/all/misc/installer/ChangeInstallerSource.kt
@@ -17,6 +17,8 @@ val changeInstallerSource = resourcePatch(
     description = "Spoofs the installer source so the app appears to be installed from an app store.",
     default = false
 ) {
+    category("Spoofing")
+
     val packageInstallerName = stringOption(
         key = "packageInstallerName",
         default = "com.android.vending",
--- a/patches/src/main/kotlin/app/morphe/patches/all/misc/network/OverrideCertificatePinningPatch.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/all/misc/network/OverrideCertificatePinningPatch.kt
@@ -41,6 +41,8 @@ val overrideCertificatePinningPatch = resourcePatch(
     description = "Overrides certificate pinning, allowing to inspect traffic via a proxy.",
     default = false
 ) {
+    category("Privacy and network")
+
     execute {
         val resXmlDirectory = get("res/xml")
         var networkSecurityFileName = "network_security_config.xml"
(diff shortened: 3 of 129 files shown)


ghost:
Thank you for making this feature optional and not forced ON to everyone.

MarcaDian:
Closed because it is unnecessary. This will only be for universal patches and third-party bundles, we only have 4 universal patches.

PR #2787, opened 2026-09-04
fix(GmsCore support): Prompt to update outdated MicroG

### LINKED ISSUE #2722: feat: Update "GmsCore support" patch
### Feature description

Over the past two years, the microg team has updated app permissions, intent actions, and content provider authorities.

The [constants](https://github.com/MorpheApp/morphe-patches/blob/3dfb76460a4337239a475f6d707dc7f47e5a9010/patches/src/main/kotlin/app/morphe/patches/shared/misc/gms/GmsCoreSupportPatch.kt#L241-L503) in the `GmsCore support` patch have not been updated for over two years.

The constants in the `GmsCore support` patch should be updated based on the [MicroG-RE 7.0.0](https://github.com/MorpheApp/MicroG-RE/releases/tag/7.0.0) (Check `AndroidManifest.xml`).

### Motivation

The service binder may not pass the service to microg due to missing constants. This could be the cause of other issues.

### Acknowledgements

rushiranpise:
Refs #2722

## 1. Sync rewrite constants with MicroG-RE 7.0.0
Verified against MicroG-RE's merged manifest and the shipped Morphe
YouTube/YT Music dex:
- **Add** `com.google.android.location.internal.GoogleLocationManagerService`
  and `com.google.android.gms.wallet.firstparty.ACTION_PURCHASE_MANAGER`
  to ACTIONS (both declared by RE under `app.revanced.*`; binds failed
  on RE-only installs).
- **Remove** firebase auth/dynamiclinks actions, GoogleSettingsLink and
  the legacy C2DM REGISTRATION action (RE declares these under their
  original `com.google.*` names, so rewriting them resolved to nothing).
- **Fix** the subscribedfeeds authority mapping
  (`app.revanced.subscribedfeeds` → actual
  `app.revanced.android.gsf.subscribedfeeds`, now via AUTHORITIES).

## 2. Prompt to update outdated MicroG installs
image

Old MicroG versions cause binder and feature issues, so patched apps
now compare the installed version against the **latest stable fetched
live from the MicroG-RE `main` branch**:
- Dialog shows both versions, e.g. "Your MicroG version (6.1.4) is
  outdated … latest version (7.0.0)".
- **Update** opens the MicroG download page; **Ignore** remembers which
  version was ignored and only prompts again once a newer stable
  release is available.
- Runs in the background on each app launch; fails silently when the
  version can't be fetched. Dev builds (e.g.
  `7.1.0-dev.3`) are newer than stable and are never asked to
  downgrade.


EpirusTr:
MicroG RE 7.0.0

Please fix the following problem in your free time.

Language switching is now possible but it doesn't work permanently.


https://github.com/user-attachments/assets/5ea6f2ec-b66a-4edc-ba89-fdeb0b620e91

BlackGold8282:
It would be great if there were options in the Morphe settings to turn feature on or off 
: Disabled, Stable only, Pre-release

LisoUseInAIKyrios:
MicroG has its own update check that can handle pre-release updates.

Pre-release downloads are only available from the GitHub release page which can be a little confusing if someone has never used it before: "where is the download link? which apk do I download? Why is arm64apk showing me 'ABI Unsupported'? Why is the launcher hidden after I installed `-noicon`??"

For now I think this is ok only prompting to update to stable MicroG because the primary purpose is to update users who are way outdated.

Bec-de-Xorbin:
Is 6.1.4 way outdated? Which issues it may cause? I really don't like to update apps that work. Previously major MicroG updates have caused more problems than solving them.

PR #2890, opened 2026-09-10
fix(YouTube - Loop video): Respect sleep timer end of video state

### LINKED ISSUE #842: bug(YouTube): Sleep timer does not worked when loop video button is active
### Bug description

Step 1: Click loop video action button
Step 2: Go to player flyout menu -> sleep timer
Step 3: Choose `end of the video` in sleep timer flyout menu
Step 4: Wait until the video ended

### Acknowledgements

ILoveOpenSourceApplications:
### Description

Closes #842.

### Additional context

The custom loop video patch previously intercepted the `ENDED` player status and immediately looped the video. This prevented YouTube's native Sleep Timer from receiving the event and stopping playback.

This fix hooks the native Sleep Timer controller to check if the "End of video" timer is active. If it is, the loop patch yields control back to the native player, allowing the sleep timer to correctly pause playback as the user intended.

### Test results

### DIFF
--- a/patches/src/main/kotlin/app/morphe/patches/youtube/interaction/loop/LoopVideoPatch.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/youtube/interaction/loop/LoopVideoPatch.kt
@@ -10,9 +10,12 @@
 
 package app.morphe.patches.youtube.interaction.loop
 
+import app.morphe.patcher.extensions.InstructionExtensions.addInstruction
+import app.morphe.patcher.extensions.InstructionExtensions.addInstructions
 import app.morphe.patcher.extensions.InstructionExtensions.addInstructionsWithLabels
 import app.morphe.patcher.extensions.InstructionExtensions.getInstruction
 import app.morphe.patcher.patch.bytecodePatch
+import app.morphe.patcher.util.proxy.mutableTypes.MutableMethod.Companion.toMutable
 import app.morphe.patches.shared.misc.settings.preference.SwitchPreference
 import app.morphe.patches.youtube.misc.extension.sharedExtensionPatch
 import app.morphe.patches.youtube.misc.settings.PreferenceScreen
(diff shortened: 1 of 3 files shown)


0xrxL:
@ILoveOpenSourceApplications I notice you're improving some of the issues that you found in the patches. At this point I'd like to ask you (if you can reproduce it ofc) if you can attempt to fix an issue, that often are annoying me: YouTube basically stops loading any data, even feeds, suddenly, as if a background task is running endlessly and blocking everything else.

ILoveOpenSourceApplications:
Hey, thanks, just trying to squash the bugs. 

From my understanding, the issue you're facing is that sometimes the UI takes longer to load? Correct me if I'm wrong though.

If a way to consistently replicate the issue is found, and if I come across said issue myself, yeah I can try to take a look at it.

0xrxL:
> Hey, thanks, just trying to squash the bugs.
> 
> From my understanding, the issue you're facing is that sometimes the UI takes longer to load? Correct me if I'm wrong though.
> 
> If a way to consistently replicate the issue is found, and if I come across said issue myself, yeah I can try to take a look at it.

Yes it is a problem that prevents the UI from continuing its operations, but without freezing it by sending the app in ANR state.

The only way I have to get it working again is to change the wifi connection (I guess this blocks the execution of the problematic code).

patches/src/main/kotlin/app/morphe/patches/youtube/interaction/loop/LoopVideoPatch.kt
```diff
@@ -70,5 +82,58 @@ val loopVideoPatch = bytecodePatch(
...
+
+                SleepTimerCancelMethodFingerprint.classDef.apply {
+                    interfaces.add(EXTENSION_SLEEP_TIMER_INTERFACE)
+
+                    methods.add(
+                        ImmutableMethod(
+                            type,
+                            "patch_isSetToEndOfVideo",
+                            listOf(),
+                            "Z",
```
LisoUseInAIKyrios, line 101: Instead of adding a new method, does either `VideoInformation.isAtEndOfVideo()` or `VideoState.ENDED` work for this?

LisoUseInAIKyrios: Or is this exposed interface method specific to the sleep timer?

ILoveOpenSourceApplications: Yes, this exposed interface method is pretty specific to the Sleep Timer.

Using `VideoInformation.isAtEndOfVideo()` or `VideoState.ENDED` unfortunately won’t work here, because the player is already in the `ENDED` state by the time our loop patch kicks in.

The problem is essentially a race between our loop patch and YouTube’s Sleep Timer. When the video ends, our patch catches the `ENDED` state and immediately seeks back to 0, so the event never reaches the Sleep Timer. As a result, the timer doesn’t get a chance to pause the app.

To handle this, we need to know whether the user actually has the Sleep Timer set to end with the video before we start looping. This interface checks the Sleep Timer controller’s internal state and specifically looks for `ACTIVE_END_OF_VIDEO`.

If that mode is active, `shouldLoopVideo()` returns false and lets YouTube handle the end-of-video event normally. That way, the Sleep Timer gets a chance to pause the player instead of our loop patch taking over.

PR #2917, opened 2026-09-12
chore: Resolve build warnings, deprecations, and static analysis issues

ILoveOpenSourceApplications:
### Description

- Safer reflection: Caught specific reflection exceptions instead of broad ones, and added a quick type check to dodge an unboxing NPE.
- Intentional deprecations: Added `@SuppressWarnings` to older APIs we still explicitly need for compatibility (`ACTION_UNINSTALL_PACKAGE`, `GET_SIGNATURES`, `versionCode`).
- Gradle fix: Swapped a config delegate to `configurations.create()` to clear a Gradle 9.6 warning.
- JSON tweak: Stopped passing null into a `@NotNull` JSON parameter to make the linter happy.


extensions/all/change-version-code/build.gradle.kts
ILoveOpenSourceApplications:
```suggestion
    compileSdk = 37
```

Will bumping this to 37 break the build for other contributors if they don't have the new SDK downloaded yet?

LisoUseInAIKyrios: Android Studio (and I think also intellij) should prompt to download SDK 37.

ILoveOpenSourceApplications: So it's fine if I change it to 37?

LisoUseInAIKyrios: Yes

PR #2938, opened 2026-09-13
feat(YouTube - Swipe controls): Add a volume steps option and set the swipe distance per step

### LINKED ISSUE #1665: feat(YouTube): Add more steps to volume adjustment using swipe gesture
### Feature description

Please add much more steps to the volume level adjustment using swipe gesture to make it smoother, like the brightness. Like it is in video player apps.

### Motivation

Currently, there are only 15 steps from mute to full volume which is way too low to get a desired volume level, especially when using headphones.

### Acknowledgements

### LINKED ISSUE #2365: bug(YouTube): Volume gesture sensitivity is too high even at the minimum scale value
### Bug description

The volume swipe-gesture patch uses a sensitivity scale from 1 to 10 (step 1), with a default value of 1. The issue is that even at this minimum value, the sensitivity is too high: swiping through roughly half the screen's height (with the phone in landscape orientation) is enough to go from 0 to 15 (max volume).

Because of this, there's little point in increasing the sensitivity above 1 - the range is already covered too quickly, and at higher values controlling the volume becomes even harder.

For comparison: NewPipe has a similar volume gesture feature where covering the same range (0-15) requires swiping through the entire height of the screen. To me, this is the reference point for a good default sensitivity - it feels comfortable and predictable.

Suggestion
- Reduce the current minimum sensitivity by at least a factor of 2, so that a full-height swipe results in a smoother, more controlled volume change (closer to NewPipe's default behavior)
- Accordingly, the default value (1) should probably correspond to roughly the middle of the current scale, so users still have room for fine adjustment in both directions

### App version patched

YouTube 21.04.223, Paches 1.38.0

### Error logs

```shell

```

### Acknowledgements

### LINKED ISSUE #2757: feat(YouTube): Volume swipe sensitivity, change scale to 0.1
### Feature description

Change scale sensitivity from 1 to 0.1 or atleast 0.5 if possible.

### Motivation

For example my samsung s26ultra has media volume 1 to 150, its 1 to 15 while swiping, but minimum swipe volume sensitivity is only 1, so for example 1 is not loud enought for night listening but 2 is too loud, perfect would be between 1.5.
Samsung s26ultra volume is 1 to 150, so when swipe is disabled I can change system volume for example to 13 which will be 1.3/15, etc.

Image

Image

### Acknowledgements

MarcaDian:
### Description

On devices whose volume stream exposes a fine grained index range (150 instead of 15), every key press and swipe tick moved the volume by a tenth of what the device volume UI does, and the volume sensitivity worked as an index multiplier, so unlike brightness and speed it could not be slowed down. This adds a "Volume steps" option that splits the stream into a chosen number of levels, listing only the counts the device can actually produce, and changes the volume sensitivity to set the swipe distance per step like the other gestures.

Closes #1665, #2365, #2757
___


CiupaSs:
Hi thanks for update, but it's still won't show 150 in volume steps with Samsung One UI 8.5, recommended version.
If I misunderstood it, let me know, thanks.

Screenshot_20260914_073940_YouTube Morphe
Screenshot_20260914_073936_YouTube Morphe
Screenshot_20260914_073813_Morphe
Screenshot_20260914_073803_YouTube Morphe
Screenshot_20260914_073800_YouTube Morphe

MarcaDian:
This PR doesn't promise that; it promises to reduce the value from 150 to 15.

This value is obtained from a system component. We cannot control the system steps, specifically, we cannot increase them; we can only reduce them by dividing them.

CiupaSs:
> This PR doesn't promise that; it promises to reduce the value from 150 to 15.
> 
> This value is obtained from a system component. We cannot control the system steps, specifically, we cannot increase them; we can only reduce them by dividing them.

Understand, thanks <3

PR #2942, opened 2026-09-13
refactor(Reddit - Custom app icon): Migrate app icon picker to `CustomDialog`

ILoveOpenSourceApplications:
### Description

- Replaced standard `AlertDialog` builders with `CustomDialog` utility. The icon picker list now uses standard checkmarks (removing the old text badge) and removes list dividers to match other preference dialogs.
- Updated `MiscellaneousPreferenceCategory` to properly check `CustomAppIconPatch.isPatchIncluded()` in `getSettingsStatus()`, ensuring the category remains visible if only this patch is applied.

### Test results


extensions/reddit/src/main/java/app/morphe/extension/reddit/patches/CustomAppIconPatch.java
```diff
@@ -305,21 +334,30 @@ public View getView(int position, View convertView, @NonNull ViewGroup parent) {
...
-                TextView badge = new TextView(context);
-                badge.setText(str("morphe_app_icon_active"));
-                badge.setTextSize(12f);
-                col.addView(badge);
-            }
             row.addView(col);
 
+            ImageView checkIcon = new ImageView(context);
+            checkIcon.setImageResource(Utils.appIsUsingBoldIcons()
+                    ? CustomDialogListPreference.DRAWABLE_CHECKMARK_BOLD
```
LisoUseInAIKyrios: The last time I checked Reddit doesn't use this bold icon hook

ILoveOpenSourceApplications:
All looking good?

extensions/reddit/src/main/java/app/morphe/extension/reddit/patches/CustomAppIconPatch.java
ILoveOpenSourceApplications: So, should I remove it here as well?
https://github.com/MorpheApp/morphe-patches/blob/c898919d011835f1aee71e8b038d2cdd855f095b/patches/src/main/kotlin/app/morphe/patches/reddit/misc/settings/SettingsPatch.kt#L84-L89

LisoUseInAIKyrios: Yeah might as well.

LisoUseInAIKyrios:
Yes looks good.

I initial had a setting to save/load the icon, but the user can select icons from the Reddit built-in premium menu which complicates things a little.

Additional the icons are not added by a patch and reddit can remove them anytime, which makes importing settings a little more complicated.

I think for now it's ok to just have a picker and not import/export the icon.

ILoveOpenSourceApplications:
The reason I removed the setting was due to the possibility of the user importing the setting, which in turn creating a ghost icon of Reddit.

Edit: I would have also preferred the app restart rather than just closing, but that'll also complicate things.

PR #2958, opened 2026-09-14
fix(YouTube - Settings): Fix crash opening settings after the app process is restarted

### LINKED ISSUE #2940: bug(YouTube): AbstractPreferenceFragment: onCreate() failure
### Bug description

I had Morphe settings open in YouTube Morphe and I switched over to another app and then later switched back. I saw a popup that said "AbstractPreferenceFragment: onCreate() failure...". I took a screenshot and tried to back out of it but the app reopened itself. Debug logging was not enabled.

Image

### App version patched

YouTube_Morphe-v21.36.45-patches-v1.43.0-dev.3.apk

### Error logs

```shell

```

### Acknowledgements

MarcaDian:
### Description

Fixes a crash when opening or returning to the settings after the app process was restarted in the background: the PoToken and OAuth2 preferences registered their activity lifecycle callbacks through the Application stored by the "Spoof video streams" main activity hook, which is null if the settings activity is restored without the main activity ever being created.

They now take the Application from their own preference context, and the unused main activity reference is removed from the spoof patch.


0xrxL:
I was about to push the same fix. :rofl:

MarcaDian:
Fast and furious 🏎️

PR #2982, opened 2026-09-15
refactor(YouTube - Navigation bar): Disable translucent system buttons

LisoUseInAIKyrios:
### Description

With the recent changes to remove translucent navigation/status bar, the system navigation buttons can still use translucent for some devices.

This PR restores one of the feature flags that may not be as problematic as the other translucent flags.

### Test results


LisoUseInAIKyrios:
@BlackGold8282

Please try the artifact of this PR and verify if system navigation buttons are fixed when "Disable translucent navigation bar" is enabled.

BlackGold8282:
thanks, working well

PR #2997, opened 2026-09-16
fix(YouTube - Channel search): Search is scoped to a channel after leaving its page

### LINKED ISSUE #2985: bug(YouTube): Search keeps applying previous "Channel search" filter
### Bug description

After using the new Channel Search feature, returning to the Home tab and using the global search bar keeps searching within the last visited channel instead of searching all of YouTube.

*Steps to reproduce:*
-Go to any channel and search for a video using Channel Search.
-Go back to the Home tab.
-Use the normal search bar for a global search.
-The search results will still be filtered to the previous channel.

*Expected behavior*
Leaving the channel should reset the search scope back to global.

*Workaround*
Switching tabs (e.g., going to Subscriptions and then back to Home) resets the search bar and fixes the issue. Though it shouldn't be this way. 

### App version patched

21.13.164

### Error logs

```shell

```

### Acknowledgements

MarcaDian:
### Description

The channel search scope was taken from the last browse request, so a page served from cache made no request and left the previous channel id in place: pressing back from a channel to the home feed kept the next global search scoped to that channel.

The browse id is now read from the endpoint that `BrowseFragment` is handed for the page it shows, which also happens when returning to a cached page, so the scope always follows the page on screen.


thedognamedbella:
Tested the artifact, and it works if you click on Home tab, but it doesn't work when you exit the channel to the search feed. Search feed still shows channel search instead of global search.

This is especially troublesome when user has navigation bar hidden, and had searched a lot. To be able to perform a new search user is forced to hit previous button multiple times to go to the Home tab, because this feature is still showing searches from a channel when you're on the search feed.

Tested with v21.13.164.

MarcaDian:
> Tested the artifact, and it works if you click on Home tab, but it doesn't work when you exit the channel to the search feed. Search feed still shows channel search instead of global search.

Should be fixed.

thedognamedbella:
> > Tested the artifact, and it works if you click on Home tab, but it doesn't work when you exit the channel to the search feed. Search feed still shows channel search instead of global search.
> 
> Should be fixed.

Works correctly.

PR #3033, opened 2026-09-17
fix(YouTube Music - Third-party lyrics): Lyrics are not shown in a panel that has no heading

MarcaDian:
### Description

On some accounts YouTube Music shows the lyrics engagement panel without a heading, and in the view tree it is then indistinguishable from the other panels that share the same container.

Since the third-party lyrics panel was only installed after matching the heading against `lyrics_tab_title`, those users never saw it at all.

This hooks the field where the app itself keeps the panel currently in the container, so the lyrics panel is recognized by identity, and the heading is kept only as a fallback for versions where the hook does not apply.

### Test results

---


DonDae001:
Tried the artifact, and it doesn't seem to work with this UI:

<details>
Image

Image
</details>

Only the recommended patches from the artifact were applied and 'Third-party lyrics' is enabled in settings.

MarcaDian:
> Tried the artifact, and it doesn't seem to work with this UI:
> 
> <details>
> Image
> 
> Image
> </details>
> 
> Only the recommended patches from the artifact were applied and 'Third-party lyrics' is enabled in settings.

Turn off
`Morphe > General > Spoof app version`

DonDae001:
Already turned off...

<details>
Screenshot_20260918_073248_YT Music Morphe
</details>

Import/Export is also empty as I haven't changed any settings yet

PR #3060, opened 2026-09-19
fix(YouTube - Remove background playback restrictions): Auto pause when video is loading

### LINKED ISSUE #2765: bug(YouTube): Auto pause when device is locked doesn't work
### Bug description

In some reason this new patch called "Auto pause when device is locked" doesn't work, maybe it's my device or my settings that is the culprit.

In my test video keeps playing when I lock my device, where it should auto pause it.

I will keep testing and will inform if I find what is causing it not to work.

### App version patched

v21.07.247 - v1.4.2 dev 1

### Error logs

```shell
No error popups appeared.
```

### Acknowledgements

gajera1080:
### Description

Fixes an issue where "Auto pause when device is locked" fails if the device screen is turned off while a video is still buffering or loading.

Previously, `handleScreenOff` immediately returned if `VideoState.getCurrent() != VideoState.PLAYING`. As a result, when a video loaded after the device was already locked, playback would still begin in the background and play audio.

This change:
- Defers the auto-pause trigger by setting a `pendingAutoPause` flag when the screen turns off during non-playing states.
- Listens to `VideoState.onChange` so that as soon as the video transitions to `PLAYING`, the media pause event is dispatched.
- Listens to `ACTION_SCREEN_ON` to safely discard the pending pause if the user turns the screen back on before the video finishes loading.

Resolves #2765

### Additional context

All logic is contained within `BackgroundPlaybackPatch.java` by subscribing to the existing `VideoState.onChange` observer, avoiding any external modifications or architecture leaks.

### Test results

- [x] Tested on `21.13.164`.


extensions/youtube/src/main/java/app/morphe/extension/youtube/patches/BackgroundPlaybackPatch.java
```diff
@@ -44,6 +44,8 @@ public enum AutoPauseOnLockMode {
 
     private static boolean receiverRegistered;
 
+    private static volatile boolean pendingAutoPause = false;
```
LisoUseInAIKyrios:
```suggestion
    private static boolean pendingAutoPause;
```

I'm somewhat sure everything is on main thread, so volatile isn't needed.

Issue #834, opened 2026-03-12
feat(YouTube): Hide Shorts flyout menu items


AkshayKanathur:
### Feature description

Please give us an option to hide the Shorts flyout menu items.

![Image](https://github.com/user-attachments/assets/a642a1e8-749f-4ffa-bf6c-bd941b03e091)

### Motivation

Adding this feature will help to keep settings which are only needed for the user.

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] This is a feature request for YouTube, YT Music, or Reddit

l205306:
You can hide it via `Morphe Settings > Feed > Enable feed flyout menu filter (Toggle On) > Feed flyout menu filter`.

All flyout items can be hidden except `Audio Track`. I believe this is caused by the language text next to the item (`English (US) original`).

Hence, I don’t think there’s a need to add another option, since we already have a lot of options now.

<details>

<img width="1229" height="2560" alt="Image" src="https://github.com/user-attachments/assets/0c567fc2-5be6-4eaf-a3b2-af95f0cbf9aa" />

</details>

PR #3082, opened 2026-09-20
fix(YouTube): Hide Shorts captions flyout item

### LINKED ISSUE #3067: bug(YouTube): Shorts captions option does not hide in flyout menu filter
### Bug description

<details>
Image

Image
</details>
The captions option in Shorts isn’t hiding from the three‑dot menu, even though the hide option is available in the flyout menu filter.

### App version patched

21.38.123

### Error logs

```shell

```

### Acknowledgements

Moltrax8:
### Description

Apply the existing "Hide Captions" flyout setting to the Shorts player flyout.

Shorts currently return early from the player flyout filter, so the captions item is never checked against the existing captions setting.

Keep the early return for Shorts, but filter the captions item when the setting is enabled.

Closes #3067

### Additional context

This change only affects the Shorts captions flyout item and does not enable the other player flyout filters for Shorts.

`git diff --check` passes.

### Test results


l205306:
Should this issue be closed as well? 

https://github.com/MorpheApp/morphe-patches/issues/834

Moltrax8:
> Should this issue be closed as well?
> 
> #834

I don't think #834 should be closed by this PR. This change only applies the existing Hide Captions setting to the Shorts captions item, while #834 covers Shorts flyout menu items more generally.

Issue #3091, opened 2026-09-21
bug(YouTube): Dislikes may display incorrectly depending on display DPI


98xe:
### Bug description

~~In the new update to the `Return YouTube Dislike` patch (https://github.com/MorpheApp/morphe-patches/pull/3075), the dislikes are misaligned.~~

~~Also, the spacing between the dislike value and the dislike icon doesn't match that of the like icon.~~

<img width="652" height="398" alt="Image" src="https://github.com/user-attachments/assets/f5e27acd-55cf-4de6-9602-74bdc6813dd7" />

---

After the recent update to the `Return YouTube Dislike` patch (https://github.com/MorpheApp/morphe-patches/pull/3075), dislikes may display incorrectly or may fail to display at all, depending on the device's DPI.

https://github.com/user-attachments/assets/53cf5945-6bc8-4a30-b972-a9d9b9b93a4d

### App version patched

21.37.42, 1.44.0-dev.13

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

crabmaster89-crypto:
> ### Bug description
> In the new update to the `Return YouTube Dislike` patch ([#3075](https://github.com/MorpheApp/morphe-patches/pull/3075)), the dislikes are misaligned.
> 
> Also, the spacing between the dislike value and the dislike icon doesn't match that of the like icon.
> 
[...]

PR #3098, opened 2026-09-21
fix(YouTube - Return YouTube Dislike): Avoid duplicate old bar count overlay

Moltrax8:
## Summary

- Only draw the dislike count beside the icon when the restored old video action bar is actually enabled.
- Avoid treating other wide button layouts as the restored old action bar.
- Prevents the extra count overlay from being applied to DPI/layout variants that already handle the dislike text themselves.

Refs #3091

## Testing

- `git diff --check`
- Runtime testing not performed.


MarcaDian:
It looks like this PR doesn't fix anything (there is no duplication) and doesn't resolve any issues. Thanks.

PR #3099, opened 2026-09-21
fix(YouTube - Voice over translation): Add button accessibility label

### LINKED ISSUE #3096: bug(YouTube): Floating button for voice over translation is missing an accessibility label
### Bug description

Hello, I would like to report an accessibility issue regarding the following feature in YouTube Morphe:

“Voice over translation — Read video captions aloud using on-device text-to-speech, synchronized with playback”

According to the feature description, “Shows an overlay button in video player.” This feature adds several floating buttons to the video player. However, when I use TalkBack to navigate the video player, I found that the buttons added by this feature do not have accessibility labels. This makes it very difficult for visually impaired users to find and identify these buttons.

I would also like to mention that this feature is very helpful for visually impaired users who want to understand videos in foreign languages. I hope you can add appropriate accessibility labels and translations for these buttons.

Thank you!

### App version patched

YouTube 21.13.164

### Error logs

```shell
No need
```

### Acknowledgements

Moltrax8:
## Summary

- Add a content description to the Voice over translation overlay button.
- Add the same accessibility label to the legacy player button.
- Reuse the existing localized `morphe_vot_enabled_title` string for TalkBack.

Fixes #3096

## Testing

- `git diff --check`
- Runtime TalkBack testing not performed.


bzkmsjy:
Hello, I updated to the latest patch 1.44.0 and YouTube 21.16.256. TalkBack can now detect the accessibility label of the button, but I don't think the implementation is sufficient, because TalkBack does not announce whether this switch is turned on or off. This makes it difficult for visually impaired users to determine whether the feature is currently enabled. I hope the switch state can also be announced.

I would also like to report another unrelated accessibility issue involving a missing accessibility label. In the YouTube or YouTube Music settings, when entering the Morphe secondary settings menu, there is a search button in the upper-right corner of the screen that does not have an accessibility label.

Additionally, the back button in the upper-left corner of the Morphe settings screen also does not have an accessibility label on some of my devices. This is somewhat strange: on my Samsung S25 running Android 16, TalkBack can read the label of this back button, but on my Xiaomi 10T running Android 14, it cannot.

These two issues should hopefully be relatively easy to fix. If I need to open a separate issue for them, please let me know.

Issue #3093, opened 2026-09-21
bug(YouTube): 'Show estimated likes' is not working


BlackGold8282:
### Bug description

'Show estimated likes' is not working

<img width="800" height="114" alt="Image" src="https://github.com/user-attachments/assets/00d3788d-0a51-433e-a619-6b19d8608a8d" />

### App version patched

21.16.256 & 21.38.123 - 1.44.0-dev. 13

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] All requested information has been provided properly.

MarcaDian:
Share an example of such a video.

BlackGold8282:
https://youtu.be/DBMmFzz0DGY

LisoUseInAIKyrios:
Something broke on the last minute refactoring. Fixing.

PR #3100, opened 2026-09-21
fix(YouTube - Return YouTube Dislike): Show estimated likes

Moltrax8:
## Summary

- Distinguish visible count text from non-numeric labels in the icon-only action bar.
- Allow estimated likes to be drawn when YouTube hides the original like count.
- Preserve the existing behavior when the bar already contains a numeric count.

Fixes #3093

## Testing

- `git diff --check`
- Runtime testing not performed.


LisoUseInAIKyrios:
This has a layout issues with old segmented layout mode tablet mode where the likes overlay doesn't cover the "Like" label.

Fixed with https://github.com/MorpheApp/morphe-patches/commit/b6f7617d59bae523adb6b76b94ce0b2421de3968

Thanks for all the help.

LisoUseInAIKyrios:
Actually, this change may still be needed for the new bold layout

LisoUseInAIKyrios:
Fixed with https://github.com/MorpheApp/morphe-patches/commit/a6efacfd0cdb0c8aa11312c33900a4ab70689ae1

PR #3107, opened 2026-09-21
fix(YouTube Music - Return YouTube Dislike): Show dislikes beside the dislike icon like YouTube does

MarcaDian:


laayaa98:
please fix this

Issue #2945, opened 2026-09-14
bug(YouTube): Java heap leak and OOM during playback


OpenNerdz:
### Bug description

Patched YouTube gradually fills its Java heap during normal playback/UI use and then crashes with `OutOfMemoryError`. This happened at least twice on 2026-09-14 (04:59 and 05:07). After reopening the app following the first crash, it reproduced in about eight minutes.

Environment:

- Device: Samsung Galaxy S25 Ultra (SM-S938B)
- Android: 16 (API 36)
- Morphe Manager: 1.30.0
- Morphe Patches: 1.42.0
- Patched YouTube: 21.13.164 (`versionCode=1561063732`)
- Package: `app.morphe.android.youtube`
- Debug spannable logging was not enabled

Android's process-exit history records both failures as `APP CRASH(EXCEPTION)`. The first failed process had about 737 MB RSS; the second reached about 0.90 GB RSS. ART reported a completely occupied 512 MiB Java heap, with full collections reclaiming only roughly 100–150 KiB:

```text
Clamp target GC heap from 559MB to 512MB
Alloc concurrent mark compact GC freed 152KB AllocSpace bytes, 0(0B) LOS objects, 0% free, 511MB/512MB
```

The latest crash enters Morphe's span integration at `InclusiveSpanPatch.setConversionContext(InclusiveSpanPatch.java:178)` while converting a YouTube layout context (`uts.toString()`). Inspection of the installed 1.42.0 bytecode shows that this method converts the full context object to a string and saves it in a `ThreadLocal`. The context's `toString()` includes many fields such as materialization state, an atomic reference, path builder, services, and byte arrays. This looks like a possible source of large retained layout-context strings.

That is a lead rather than a definitive attribution: the earlier OOM surfaced in `android.text.StaticLayout.generate()` while Litho was measuring text, after the heap was already full. The ExoPlayer thread also failed after the same process-wide exhaustion.

Steps observed:

1. Patch and install YouTube 21.13.164 using Morphe Patches 1.42.0.
2. Open the patched app and use the feed/play videos normally.
3. After several minutes, the Java heap grows to its 512 MiB limit.
4. The app crashes. Reopening it resets memory temporarily, but the crash repeats.

### App version patched

YouTube 21.13.164 (`versionCode=1561063732`)

### Error logs

```shell
FATAL EXCEPTION: ComponentLayoutThread1-1
Process: app.morphe.android.youtube
[...]

PR #3120, opened 2026-09-21
fix: Reduce heap use from ConversionContext dumps and JS player cache

PaN1C27:
## Description

Compact `ConversionContext.toString()` so Litho callers that stringify the context no longer dump AtomicReferences, services, and byte arrays. Span filters still see identifier, path, `horizontalCollectionSwipeProtector`, and `heightConstraint`.

Bound `yt_cache_service%KEY%challenge-solver` to one `player:<hash>` (lib/core/wrapper keys unchanged).

Fixes #2945

## Test results

Patched YouTube **21.13.164** (`app.morphe.android.youtube`):

- Playback reached `PlaybackState state=3`.
- Comment `#RickAstley` and subtitle span filters still match.
- Dummy `player:` keys in challenge-solver prefs pruned to a single hash via `CacheService.get`/`save`.

YouTube’s Platypus 2160p AV1 buffer leak is out of scope.


extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/spoof/js/nsigsolver/common/CacheService.java
```diff
@@ -41,12 +49,64 @@ public CachedData get(String section, String key) throws CacheError {
...
+                removePlayerBase(pref, base);
+            }
+        }
+    }
+
+    private void removePlayerBase(SharedPrefCategory pref, String base) {
+        pref.removeKey(getCodeKey(base));
+        pref.removeKey(getVersionKey(base));
+        pref.removeKey(getVariantKey(base));
+    }
```
LisoUseInAIKyrios, line 120: Is keeping only one value the primary fix?

Is trimming strings and the patch changes necessary?

PaN1C27: Yeah, The reporter’s dump had 104 MB in 20 historical player: hashes. One current hash is enough.

The toString() change is still needed. InclusiveSpanPatch.setConversionContext stores conversionContext.toString() in a ThreadLocal, which is the original crash stack. Span filters only match identifier, path, and the two null-check fields.

0xrxL:
Merged dev here to try the artifact on 21.38, and seems promising. Let's see...

PaN1C27:
> Merged dev here to try the artifact on 21.38, and seems promising. Let's see...

Any updates? Have you gotten around to testing it?

0xrxL:
> > Merged dev here to try the artifact on 21.38, and seems promising. Let's see...
> 
> Any updates? Have you gotten around to testing it?

I'm actively testing it, bscause also if I wasn't able to reproduce the crash, thumbnails was slow to loading (also the original ones) due to the same problem I guess. For the moment I cannot reproduce this issue anymore.

But I want to test it furtherly to be 100% sure.

0xrxL:
Looks like this PR effectively changed the situation: it's the second day that I'm not able reproduce any issue related to thumbnails cache (also the one of media notifification).

extensions/shared-youtube/library/src/main/java/app/morphe/extension/shared/spoof/js/nsigsolver/common/CacheService.java
error-reporting: I think you should keep 3 to 5 player: hashes pairs instead of only 1

0xrxL:
@PaN1C27 Check if crashes is fixed after last refactors. 👍

LisoUseInAIKyrios:
Changes work. Also has modernization of some old code.

PR #3121, opened 2026-09-21
fix(YouTube - Legacy player controls): Preserve end screen on 21.36+

### LINKED ISSUE #2947: bug(YouTube): Weird miniplayer UI at the end screen of a video
### Bug description

Now frequently noticing this is their any fix or its a bug

Image

### App version patched

21.36.45

### Error logs

```shell

```

### Acknowledgements

Moltrax8:
## Summary

- Narrow the `45752335` player overlays feature flag fingerprint to boolean methods with no parameters.
- Leave unrelated inlined occurrences of the same flag untouched.
- Preserve the `45763727` RecyclerView scrolling override and fast scrolling behavior.

Fixes #2947

## Testing

- `git diff --check`
- `gradlew.bat :patches:buildAndroid --no-daemon`
- Tested on YouTube 21.36.45:
  - Autoplay transitions to the next video without the broken end-screen UI.
  - SponsorBlock "Skip sponsor" remains functional.
- Diagnostic matching for `45752335` found:
  - `Liaj;->gd()Ljava/util/Set;`
  - `Lslm;->p()Z`
- The fingerprint now targets the boolean no-argument method instead of overriding both occurrences.


0xrxL:
So the fast scrolling will not be available anymore on last 21.38? Mhh...it's a bit annoyjng.

ILoveOpenSourceApplications:
Wrong flag, `45752335` should be the one responsible.

Moltrax8:
> Wrong flag, `45752335` should be the one responsible.

Thanks, that makes sense. I’ll keep 45763727 unchanged so fast scrolling is preserved, and update the fix to target 45752335 instead.

Moltrax8:
> Wrong flag, `45752335` should be the one responsible.

I traced this back to 45752335 being forced off by disableNewPlayerControlsFeatureFlag(). That helper is also used on 21.36+ by SponsorBlock because its skip buttons do not work with the new player controls.

Would you prefer leaving 45752335 at its stock value on 21.36+, or should the override be narrowed to specific call sites? I don't want to trade the end-screen regression for broken SponsorBlock controls.

0xrxL:
> > Wrong flag, `45752335` should be the one responsible.
> 
> I traced this back to 45752335 being forced off by disableNewPlayerControlsFeatureFlag(). That helper is also used on 21.36+ by SponsorBlock because its skip buttons do not work with the new player controls.
> 
> Would you prefer leaving 45752335 at its stock value on 21.36+, or should the override be narrowed to specific call sites? I don't want to trade the end-screen regression for broken SponsorBlock controls.

Sometimes a flag is inlined in more smali files and this enables additional unwanted behaviors. Check if this is the case for this flag.

Moltrax8:
> > > Wrong flag, `45752335` should be the one responsible.
> > 
> > 
> > I traced this back to 45752335 being forced off by disableNewPlayerControlsFeatureFlag(). That helper is also used on 21.36+ by SponsorBlock because its skip buttons do not work with the new player controls.
> > Would you prefer leaving 45752335 at its stock value on 21.36+, or should the override be narrowed to specific call sites? I don't want to trade the end-screen regression for broken SponsorBlock controls.
> 
> Sometimes a flag is inlined in more smali files and this enables additional unwanted behaviors. Check if this is the case for this flag.

I checked the history. Before 21.31, 45752335 was matched only in a public final boolean method with no parameters. In the 21.31 support commit, those method constraints were removed and the patch was changed to matchAll(), so the flag is indeed being overridden at every inlined call site now.

That lines up with the 21.36 end-screen regression. I’ll narrow the override rather than disabling 45763727 or making it state-dependent.

0xrxL:
> > > > Wrong flag, `45752335` should be the one responsible.
> > > 
> > > 
> > > I traced this back to 45752335 being forced off by disableNewPlayerControlsFeatureFlag(). That helper is also used on 21.36+ by SponsorBlock because its skip buttons do not work with the new player controls.
> > > Would you prefer leaving 45752335 at its stock value on 21.36+, or should the override be narrowed to specific call sites? I don't want to trade the end-screen regression for broken SponsorBlock controls.
> > 
> > 
> > Sometimes a flag is inlined in more smali files and this enables additional unwanted behaviors. Check if this is the case for this flag.
> 
> I checked the history. Before 21.31, 45752335 was matched only in a public final boolean method with no parameters. In the 21.31 support commit, those method constraints were removed and the patch was changed to matchAll(), so the flag is indeed being overridden at every inlined call site now.
> 
> That lines up with the 21.36 end-screen regression. I’ll narrow the override rather than disabling 45763727 or making it state-dependent.

In this case you need to check which inline do what you want, and leave the others untouched. 👍

darkphantom819:
> 
> That lines up with the 21.36 end-screen regression. I’ll narrow the override rather than disabling 45763727 or making it state-dependent.
> 

The fix is not working still seeing the weird view in the end screen

